PR #5, opened 2025-01-10
Naming packages is hard

butonic:
changes only package names


butonic:
yes, generators and makefiles need to be fixed in subsequent prs

PR #9, opened 2025-01-13
use plain pkg module

butonic:
moves /ocis-pkg to /pkg ... also in the vendors dir which we need to refactor anyway


fschade:
> Nitpick: typo in commit message: s/pgk/pkg

✅ fixed

PR #16, opened 2025-01-14
small Code changes that mostly affect user visible strings ...

butonic:
or the binary name


pkg/shared/errors.go
```diff
@@ -9,87 +9,87 @@ import (
 func MissingMachineAuthApiKeyError(service string) error {
 	return fmt.Errorf("The Machineauth API key has not been configured for %s. "+
 		"Make sure your %s config contains the proper values "+
-		"(e.g. by using 'ocis init --diff' and applying the patch or setting a value manually in "+
+		"(e.g. by using 'opencloud init --diff' and applying the patch or setting a value manually in "+
```
dragonchaser, line 12: We need to move all those strings into a const... this is horrible. (not a blocker, more of a self-blame thing :facepalm: )

PR #17, opened 2025-01-14
Debrand base data and config dir path

fschade:
## Description
update opencloud $home dir

## Changes
- `$HOME/.ocis` <-> `$HOME/.opencloud`
- `$HOME/.ocis/config/ocis.yaml` <-> `$HOME/.opencloud/config/opencloud.yaml`

## Breaking
`..., ~/.ocis` is obsolete... re run init


protogen/gen/ocis/messages/settings/v0/settings.pb.go
```diff
@@ -627,13 +627,13 @@ type Bundle struct {
...
+	Id          string      `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`                                                 // @gotags: yaml:"id"
+	Name        string      `protobuf:"bytes,2,opt,name=name,proto3" json:"name,omitempty"`                                             // @gotags: yaml:"name"
+	Type        Bundle_Type `protobuf:"varint,3,opt,name=type,proto3,enum=ocis.messages.settings.v0.Bundle_Type" json:"type,omitempty"` // @gotags: yaml:"type"
+	Extension   string      `protobuf:"bytes,4,opt,name=extension,proto3" json:"extension,omitempty"`                                   // @gotags: yaml:"extension"
+	DisplayName string      `protobuf:"bytes,5,opt,name=display_name,json=displayName,proto3" json:"display_name,omitempty"`            // @gotags: yaml:"display_name"
+	Settings    []*Setting  `protobuf:"bytes,6,rep,name=settings,proto3" json:"settings,omitempty"`                                     // @gotags: yaml:"settings"
+	Resource    *Resource   `protobuf:"bytes,7,opt,name=resource,proto3" json:"resource,omitempty"`                                     // @gotags: yaml:"resource"
```
butonic: this does not look like it belongs into this PR

opencloud/pkg/init/init.go
```diff
@@ -6,12 +6,13 @@ import (
 	"path"
 
 	"github.com/gofrs/uuid"
-	"github.com/opencloud-eu/opencloud/pkg/generators"
 	"gopkg.in/yaml.v2"
+
+	"github.com/opencloud-eu/opencloud/pkg/generators"
```
butonic, line 11: While I personally like three import groups, we use the default go fmt style for imports which only uses two groups. Please double check why your tooling introduces three groups.

pkg/shared/errors.go
```diff
@@ -9,87 +9,87 @@ import (
 func MissingMachineAuthApiKeyError(service string) error {
 	return fmt.Errorf("The Machineauth API key has not been configured for %s. "+
 		"Make sure your %s config contains the proper values "+
-		"(e.g. by using 'ocis init --diff' and applying the patch or setting a value manually in "+
+		"(e.g. by using 'opencloud init --diff' and applying the patch or setting a value manually in "+
```
butonic: hm these changes were already merged with https://github.com/opencloud-eu/opencloud/pull/16/commits/97250d5b06e59e1e5b1719c4c00bec66cd4e110a .... can you rebase?

opencloud/pkg/init/init.go
aduffeck: Hm, at least in reva we also use three groups in some cases. Maybe - if we all prefer that style - we could settle on three groups moving forward?

butonic: I tried that in the past and had to find out how to configure it in vscode. Since it requires extra effort we should stick to two groups in all repos. Otherwhise, fly by PRs might change imports unintended and that increases our efforts in reviewing them.

aduffeck: Hm, if tooling really makes three groups hard I guess it's better to avoid them. I don't know if that's still a problem however. I just tried it in my pretty vanilla vscode on my new machine and it handled both 2 and 3 groups just fine.

fschade: after some... discussion, weve decided to stick to 3 impoprt groups, sdtlib, ecternam and internal.

PR #22, opened 2025-01-14
WEB theming

fschade:
contains temporary solution to get a branded web version, once the web theme is finalized we need to bring `assets/themes/opencloud/theme.json` back!

@aduffeck i hope my changes make sense to you, @kulmann feel free testing it

```bash
$ make generate
```


services/web/assets/themes/owncloud/theme.json
```diff
@@ -1,198 +0,0 @@
-{
```
butonic, line 1: why do we delete this file? AFAICT it is used as a starting point for admins so they can customize their theme

butonic:
COOL! I can log in:
```
install pnpm
make generate
ADMIN_PASSWORD=admin go run ./opencloud/cmd/opencloud/main.go server
OC_INSECURE=true OC_LOG_PRETTY=true OC_LOG_COLOR=true OC_LOG_LEVEL=debug go run ./opencloud/cmd/opencloud/main.go server
```

services/web/assets/themes/owncloud/theme.json
fschade: We decided to have the theme inside ocis, loading the theme from the web artifacts is just a fallback….

this should help bene to develop the theme in web, once that is done, we will bring it back!

the asset loading mechanism we use for web can be a bit complicated to understand. It uses a fallback mechanism (path overloading) and in the case of themes it also merges the individual theme layers together! I tried my best to describe it here: https://github.com/opencloud-eu/opencloud/tree/main/services/web#loading-themes

a more technical description of what’s going on can be found as comments in the source code.

services/web/Makefile
```diff
@@ -33,7 +33,11 @@ ci-node-generate: pull-assets
 .PHONY: pull-assets
 pull-assets:
 	git clean -xfd assets
-	curl --fail -slL -o- https://github.com/owncloud/web/releases/download/$(WEB_ASSETS_VERSION)/web.tar.gz | tar xzf - -C assets/core/
+	rm -rf _web && git clone -b theming --single-branch git@github.com:opencloud-eu/web.git _web
```
kulmann, line 36: Please add the `_web` folder to `.gitignore`. We must make sure that the logo and font are never checked in to this repo. Otherwise, because of Apache 2.0, anyone could do anything they want with those assets. We only want to bundle all the assets into the binary.

aduffeck: Good point! Done.

aduffeck:
@fschade Your changes actually do make sense to me :D Thanks for your help.

PR #28, opened 2025-01-15
Contributing md

butonic:
This just gets rid of some unwanted words ... but actually this needs a rewrite because it is hopelessly outdated and bloated, I'd say.


butonic:
@micbar who can give input for things like matrix channel, faq, etc.?

PR #35, opened 2025-01-16
idp

butonic:
favicon still downloads the owncloud asset


services/idp/pkg/config/config.go
```diff
@@ -49,7 +49,7 @@ type Ldap struct {
 	UUIDAttributeType string `yaml:"uuid_attribute_type" env:"IDP_LDAP_UUID_ATTRIBUTE_TYPE" desc:"LDAP User uuid attribute type like 'text'." introductionVersion:"pre5.0"`
 
 	UserEnabledAttribute string `yaml:"user_enabled_attribute" env:"OC_LDAP_USER_ENABLED_ATTRIBUTE;IDP_USER_ENABLED_ATTRIBUTE" desc:"LDAP Attribute to use as a flag telling if the user is enabled or disabled." introductionVersion:"pre5.0"`
-	Filter               string `yaml:"filter" env:"OC_LDAP_USER_FILTER;IDP_LDAP_FILTER" desc:"LDAP filter to add to the default filters for user search like '(objectclass=ownCloud)'." introductionVersion:"pre5.0"`
+	Filter               string `yaml:"filter" env:"OC_LDAP_USER_FILTER;IDP_LDAP_FILTER" desc:"LDAP filter to add to the default filters for user search like '(objectclass=OpenCloud)'." introductionVersion:"pre5.0"`
```
rhafer: Hm, we haven't touched the Schema yet. I think we should keep the default to `ownCloud` as least until we defined our own Schema. (https://github.com/opencloud-eu/internal/issues/39)

services/idp/pkg/config/defaults/defaultconfig.go
```diff
@@ -98,7 +98,7 @@ func DefaultConfig() *config.Config {
 			{
 				ID:              "e4rAsNUSIUs0lF4nbv9FmCeUkTlV9GdgTLDH1b5uie7syb90SzEVrbN7HIpmWJeD",
 				Secret:          "dInFYGV33xKzhbRmpqQltYNdfLdJIfJ9L5ISoKhNoT9qZftpdWSP71VrpGR9pmoD",
-				Name:            "ownCloud Android app",
+				Name:            "OpenCloud Android app",
```
rhafer: There is no "OpenCloud Android app". Yet I think we should keep that string as is for now.
Same for iOS

services/idp/tsconfig.tsbuildinfo
rhafer, line 1: The internet seems to think that this should not be under version control at all. Shouldn't we just remove it and put it into `.gitignore`

PR #41, opened 2025-01-16
frontend

butonic:


services/frontend/pkg/command/root.go
```diff
@@ -22,11 +22,11 @@ func GetCommands(cfg *config.Config) cli.Commands {
 	}
 }
 
-// Execute is the entry point for the ocis-frontend command.
+// Execute is the entry point for the opencloud-frontend command.
 func Execute(cfg *config.Config) error {
 	app := clihelper.DefaultApp(&cli.App{
 		Name:     "frontend",
-		Usage:    "Provide various ownCloud apis for oCIS",
+		Usage:    "Provide various ownCloud apis for OpenCloud",
```
rhafer: I don't think the endpoints provided are really specifc to ownCloud, are they?

How about:
"Provide various HTTP APIs for OpenCloud"?

PR #45, opened 2025-01-16
Rebrand auth-{basic,bearer,machine,service}

aduffeck:


aduffeck:
@butonic @rhafer I suppose nothing changes for the [`owncloudsql` auth provider](https://github.com/opencloud-eu/opencloud/blob/main/services/auth-basic/pkg/config/config.go#L55).

What about the [ldap schemas](https://github.com/opencloud-eu/opencloud/blob/main/services/auth-basic/pkg/config/defaults/defaultconfig.go#L57-L71)?

butonic:
owncloudsql stays the same, correct. ldap schema names like `ownCloudUserName` and `ownCloudUUID` should remain.

PR #30, opened 2025-01-15
Fix ocis init

micbar:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #52, opened 2025-01-16
Rebranding init command

ScharfViktor:
While running the opencloud binary, I found that the init command is not rebranded. Fixed it

fixed output while generating config
```
Do you want to configure OpenCloud with certificate checking disabled?
 This is not recommended for public instances! [yes | no = default] yes

=========================================
 generated OpenCloud Config
=========================================
 configpath : /Users/v.scharf/.opencloud/config/opencloud.yaml
 user       : admin
 password   : admin
```


fschade:
replaced by https://github.com/opencloud-eu/opencloud/pull/30

PR #89, opened 2025-01-20
enhancement: include web theme

fschade:
## Description
bring back the web theme, its now part of opencloud again


services/web/Makefile
```diff
@@ -33,11 +33,10 @@ ci-node-generate: pull-assets
 .PHONY: pull-assets
 pull-assets:
 	git clean -xfd assets
-	rm -rf _web && git clone -b main --depth 1 git@github.com:opencloud-eu/web.git _web
-	make -C _web release
-	tar xfv _web/release/web.tar.gz -C assets/core/
-	cp -r _web/dev/docker/ocis-themes/opencloud assets/core/themes/ && rm -rf assets/core/themes/owncloud
-	rm -rf _web
+	git clone -b main --depth 1 git@github.com:opencloud-eu/web.git assets/core/origin
```
aduffeck, line 36: I guess with this change we can remove [`services/web/.gitignore`](https://github.com/opencloud-eu/opencloud/blob/main/services/web/.gitignore) again.

fschade: good catch, done

PR #90, opened 2025-01-20
Remove the docs folder for now

aduffeck:
We'll bring back documentation once we decided on how to do it.


micbar:
Do you sneak in the change of the reva repo here?

butonic:
> Do you sneak in the change of the reva repo here?

nope

PR #97, opened 2025-01-20
More rebranding related adjustments to the Dockerfiles

rhafer:


Dockerfile
rhafer, line 1: Note that this files currently fails to build. Somehow I can't get the `make ci-node-generate` step accept my ssh key. Should be fixed as soon as web is public again.

Issue #87, opened 2025-01-20
Setup api-tests


ScharfViktor:
## setup api tests localy (until we set up CI ) and run all tests after rebranding

- [x] need php@7.3 otherwise TUS tests with `ankitpokhrel/tus-php` won't work. How setup see here: https://github.com/opencloud-eu/opencloud/pull/81#issue-2795834431 
- [x] need script which run all test suites https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/run_all_tests.sh

- [x] local api tests:
```
"suites": [
                "apiArchiver",
                "apiContract",
                "apiCors",
                "apiAsyncUpload",
                "apiDownloads",
                "apiDepthInfinity",
                "apiLocks",
                "apiActivities",
                "apiSettings",
                "apiGraph",
                "apiServiceAvailability",
                "apiGraphUserGroup",
                "apiSpaces",
                "apiSpacesShares",
                "apiSpacesDavOperation",
                "apiSearch1",
                "apiSearch2",
                "apiReshare",
                "apiSharingNg1",
                "apiSharingNg2",
                "apiSharingNgShareInvitation",
                "apiSharingNgLinkSharePermission",
                "apiSharingNgLinkShareRoot",
                "apiAccountsHashDifficulty",
                ]
```

- [x] local api test with extra server envs:
        
 ```
    "apiNotification",
	~"apiAntivirus",~
	"apiSearchContent",
	~"apiOcm",~
	~"apiCollaboration",~
	~"apiAuthApp",~
	~"cliCommands",~
```

- [x] core tests
all that have prefix `core`

- [x] run tests in docker #106 
- [x] run e2e tests #119  

related PRs: 
- #81 
- #92 
- #106

v-scharf:
closed as completed

here you can see how to run all tests https://github.com/opencloud-eu/opencloud/pull/119#issue-2809629259

PR #106, opened 2025-01-22
run tests in docker fix

ScharfViktor:
part of #87 

`make -C tests/acceptance/docker Core-API-Tests-ocis-storage-3` led to an error when running on mac - fixed it

### DIFF
--- a/tests/ocwrapper/Makefile
+++ b/tests/ocwrapper/Makefile
@@ -1,6 +1,8 @@
-.PHONY: build
-
+GOOS := linux ## only linux is supported
 GOARCH := $(shell go env GOARCH)
 
+.PHONY: build
 build:
-	env GOOS=linux GOARCH=$(GOARCH) go build -o bin/ocwrapper
+	GOOS=$(GOOS) \
+	GOARCH=$(GOARCH) \
+	go build -o bin/ocwrapper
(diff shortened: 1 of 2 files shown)


tests/ocwrapper/Makefile
```diff
@@ -1,3 +1,3 @@
 .PHONY: build
 build:
-	go build -o bin/ocwrapper
\ No newline at end of file
+	env GOOS=linux GOARCH=arm64 go build -o bin/ocwrapper
```
fschade: cant we build that in the container?

v-scharf:
```
GOARCH := $(shell go env GOARCH)

build:
	GOOS=linux GOARCH=$(GOARCH) go build -o bin/ocwrapper
```
maybe just do it that way?

dragonchaser: > ```
> GOARCH := $(shell go env GOARCH)
> 
> build:
> 	GOOS=linux GOARCH=$(GOARCH) go build -o bin/ocwrapper
> ```
> 
> maybe just do it that way?

this works for me on arch

fschade: please check https://github.com/opencloud-eu/opencloud/blob/main/.make/release.mk for reference...

PR #121, opened 2025-01-24
restart at new transifex orga

butonic:
restart translations at the opencloud-eu orga


butonic:
unit tests passed locally

PR #122, opened 2025-01-27
chore: remove or change branded test files in thumbnail service

JammingBen:
* Replaces mp3 test files because they had branded meta data
* Replaces test images that had a logo before
* Removes unused `test.ggp` test file


butonic:
- [x] can you double check the test files? images should not show a logo ...

PR #124, opened 2025-01-27
chore: adjust doc links in graph service readme svg file

JammingBen:
As per title.


services/graph/images/mermaid-graph.svg
JammingBen, line 1: Fyi I just adjusted the doc links at the end of the svg file. They are hidden though, so no visual change.

PR #131, opened 2025-01-28
Rebrand default IDP clients

rhafer:
Change default client ids and remove secrets. Desktop is a public client now. Let's assume we'll turn Android and IOS into public clients as well for now.

### DIFF
--- a/services/idp/pkg/config/defaults/defaultconfig.go
+++ b/services/idp/pkg/config/defaults/defaultconfig.go
@@ -107,7 +107,7 @@ func DefaultConfig() *config.Config {
 				Name:            "OpenCloud iOS app",
 				ApplicationType: "native",
 				RedirectURIs: []string{
-					"oc://ios.opencloud.com",
+					"oc://ios.opencloud.eu",
 				},
 			},
 		},


services/idp/pkg/config/defaults/defaultconfig.go
```diff
@@ -86,31 +86,28 @@ func DefaultConfig() *config.Config {
...
 			{
-				ID:              "mxd5OQDk6es5LzOzRvidJNfXLUZS2oN3oUFeXPP8LpPrhx3UroJFduGEYIBOxkY1",
-				Secret:          "KFeFWWEZO9TkisIQzR3fo7hfiMXlOpaqP8CFuTbSHzV1TUuGECglPxpiVKJfOXIx",
-				Name:            "ownCloud iOS app",
+				ID:              "OpenCloudIOS",
+				Name:            "OpenCloud iOS app",
 				ApplicationType: "native",
 				RedirectURIs: []string{
-					"oc://ios.owncloud.com",
+					"oc://ios.opencloud.com",
```
aduffeck: I suppose this should be `.eu` as well

```suggestion
					"oc://ios.opencloud.eu",
```

rhafer: True. Fixed.

PR #142, opened 2025-01-29
reset version to 0.1.0

butonic:


pkg/version/version.go
```diff
@@ -24,10 +24,9 @@ var (
 	Date = "dev"
 
 	// Legacy defines the old long 4 number OpenCloud version needed for some clients
-	Legacy = "10.11.0.0"
-
+	Legacy = "0.1.0.0"
```
rhafer, line 27: Hm, shouldn't we just remove `Legacy` and `LegacyString`.  ICBW but aren't they supposed to be owncloud 10 release numbers IIRC.  Resetting them seems weird at least.

butonic: Yeah, I'd like to get rid of them, too. @TheOneRing does the client need them? They are rendered in the fake status.php

but ... not something for today

TheOneRing: if it still does, its a client bug, so 🔥

butonic: @kulmann does web read the status.php?

butonic: Can we drop the whole status.php????

butonic: Well, none of that is part of this PR ... lets go to v0.1.0 first

rhafer: Fine with me. Please open an issue so that we do not forget to remove that stuff.

TheOneRing: > Can we drop the whole status.php????

The client still uses the endpoint to detect whether a server is installed at the given server url and whether it is reachable or moved. So if you remove status.php we need to update the clients with a new detection mechanism.

kulmann: Web doesn't use it. But I don't know (yet) about iOS and Android. Desktop was already mentioned.

TheOneRing: As both still support oc10 they will use status.php

kulmann: Then we need to keep it. We want to keep the both clients compatible until we have our own.

butonic: dropping status.php tracked in https://github.com/opencloud-eu/opencloud/issues/151

PR #178, opened 2025-02-10
clean up tests

ScharfViktor:
- blocksDownload is not implemented (issue and failed tests have been open for more than 1 year) so I deleted this tests and scenarios from expected failures file
- `A User can get information of another user with Graph API` was redundant. - cleaned it up a bit
- deleted tests `file versions do not report the version author` - isn't bug (enhancement) was opened more 3 years 
- fix preview tests


v-scharf:
> I'm ok wkith deleting the version author tests, unless @tbsbdr sees this on the roadmap.

we had failed tests more 3 years. I think - better to delete
but when we decide to add it I'll restore/write new tests

Issue #228, opened 2025-02-21
500 error while restoring file version


ScharfViktor:
## Describe the bug
user cannot restore file version

## Steps to reproduce
1. Create `.txt` file
2. Try to restore old file version

## Expected behavior
version is restored

## Actual behavior
500 error

![Image](https://github.com/user-attachments/assets/15a52140-c43e-40de-88d0-007f78e8d29f)

## Setup
start opencloud with docker compose using full example https://github.com/opencloud-eu/opencloud/tree/main/deployments/examples/opencloud_full

logs:

```console
2025-02-21 15:50:16 {"level":"error","service":"ocdav","name":"eu.opencloud.web.ocdav","traceid":"2ab19bc722be353e68e613528f3a4a4c","request-id":"b84257de-1bc9-43c8-b429-b10b6a8d0eaf","resourceid":{"storage_id":"b12bb9a1-3c55-4bb8-8d19-1bac4f10001b","opaque_id":"4ad6b6ea-f2c2-4293-a41d-c848dbddf421","space_id":"4ad6b6ea-f2c2-4293-a41d-c848dbddf421"},"key":"ee6cb948-180d-4dd9-95b4-8869662b9be5.REV.2025-02-21T14:48:08.190713512Z","status":{"code":15,"message":"restore file version:internal error: failed to copy blob xattrs to old revision to node: stat /var/lib/opencloud/storage/users/spaces/4a/d6b6ea-f2c2-4293-a41d-c848dbddf421/nodes/ee/6c/b9/48/-180d-4dd9-95b4-8869662b9be5.REV.2025-02-21T14:48:08.190713512Z: no such file or directory","trace":"2ab19bc722be353e68e613528f3a4a4c"},"code":500,"time":"2025-02-21T14:50:16Z","line":"github.com/opencloud-eu/reva/v2@v2.27.3-0.20250220094822-4ffb9dbabef5/internal/http/services/owncloud/ocdav/errors/error.go:178","message":"Internal Server Error"}
[...]

PR #236, opened 2025-02-24
Bump reva

aduffeck:
https://github.com/opencloud-eu/opencloud/issues/228  fix


ScharfViktor (approved):
checked. version restoring works fine. thank you 👍

Issue #220, opened 2025-02-20
PosixFS: Make PosixFS non-collaborative mode the new install default for User Storage


dragotin:
# Description

This story describes the tasks that need to be done to make the **non-collaborative PosixFS** the standard for new installations.

## User Stories

* > As an OpenCloud User I want a flawless new installation with non-collaborative PosixFS

## Value

With that, OpenCloud manages the data in a human readable file tree. The grows the confidence.

## Acceptance Criteria

- [x] OpenCloud just starts with PosixFS non collaborative https://github.com/opencloud-eu/opencloud/pull/237
- [x] CI runs on PosixFS non collaborative
- [x] System storage remains decomposed
- [x] Issue #192 needs to be decided & fixed
- [x] Documentation is updated 
- [x] A newly installed OpenCloud uses the non-collaborative mode of PosixFS. 
  - [ ] this would require checking the on disk layout ... I'd recommend documenting a breaking change. Users have to configure the `ocis` filesystem driver if they want to use opencloud with their old data. 
- [ ] Upgrades stay with their setting, which is probably ~~Decomposed FS~~ the `ocis` driver.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the OpenCloud charts is compatible

kulmann:
@dragotin please close if it is ok to do so. We've made the switch. We didn't tick all the boxes from the todo list.

PR #237, opened 2025-02-24
[posix] change storage users default to posixfs

butonic:
posixfs requires shared memory between the storageproevider and dataprovider, so we have to use by default.

related: https://github.com/opencloud-eu/opencloud/issues/220


rhafer (approved):

rhafer:
rebased to solve merge conflict

PR #245, opened 2025-02-25
fix: docker image names

kulmann:
## Description
Docker images live in the org namespace `opencloudeu` since `-` is not an allowed char on docker hub. Hope I found all the occurrences.

## Types of changes

- [ ] Technical debt

### DIFF
--- a/deployments/examples/opencloud_full/.env
+++ b/deployments/examples/opencloud_full/.env
@@ -36,13 +36,13 @@ TRAEFIK_ACME_CASERVER=
 # Note: the leading colon is required to enable the service.
 OPENCLOUD=:opencloud.yml
 # The opencloud container image.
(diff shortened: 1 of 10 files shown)


TheOneRing (approved):
LGTM 🚢

deployments/examples/opencloud_full/.env
Svanvith, line 1: `OC_DOCKER_IMAGE=opencloudeu/opencloud` Should also be  `OC_DOCKER_IMAGE=opencloudeu/opencloud-rolling`

kulmann: let's discuss that separately... since we don't have a production release, yet, compose up will fail at the moment. but in the long run it's arguable if we want the production release or the rolling release to be the default.

PR #278, opened 2025-02-27
fix continues "make generate" calls in services/web:pull-assets

jochumdev:
This commit fixes running "make generate" after a previous run by don't failing when the assets in services/web have been already pulled.

When the assets have been already pulled it tries a `git pull` this changes the previous behaviour.

Kind regards,
René


fschade:
Thanks for your pr ❤️can you please checkout main before pulling… just in case if for whatever reason the wrong branch is active

jochumdev:
@fschade I'm not sure thats a good idea, was thinking about that as well.

When a developer is working on web he mights want to keep his branch active?

fschade:
the assets are removed by default at the end, do you see any use case where that would be helpful?

But…. pulling the assets was just a workaround in the beginning because we had no web release at that point.

since the web rolling is out now we can change it back and just download the transpiled artifacts instead of using git.

I have no computer here but I will have a look tomorrow morning.

if you wanna have a speedy solution for local development, you can use the docker setup from web or use that env: https://github.com/opencloud-eu/opencloud/blob/fa498a031b56abe9c6f5a621f047da0fe57569ec/services/web/README.md?plain=1#L10

Thanks for your contribution, really appreciated

jochumdev:
Should have looked at the rest of the Makefile :)

I force pushed the checkout.

Thanks for OpenCloud :)

fschade:
@jochumdev SRY for invalidating your work, as said above, pulling web and building the artifacts in place was just a workaround because we had no web release back then. We would have missed that for some more weeks maybe without your contribution.

jochumdev:
@fschade np, makes stuff much simpler, thats the goal. :+1:

fschade:
@individual-it found another side effect here https://github.com/opencloud-eu/opencloud/pull/295

by default the make steps are using the official released artifacts, as @jochumdev pointed out in some cases it could be helpful to not use those and instead use a branch from the web repo.

to support pulling and building the assets instead of downloading the pre build artifacts the MAKEFILE got a new task:

```bash
make pull-assets # uses git clone or pull
make download-assets # uses the pre build release artifacts
```

this takes care of cloning (or updating), building and copying the individual licenses.

@kulmann can you check it and see if i missed something? @jochumdev thanks for your food for thought.

services/web/Makefile
```diff
@@ -28,15 +29,33 @@ include ../../.make/generate.mk
 ci-go-generate: # CI runs ci-node-generate automatically before this target
 
 .PHONY: ci-node-generate
-ci-node-generate: pull-assets
+ci-node-generate: download-assets
```
kulmann: I'd very much like it if our default would be `pull-assets`. We can then use `download-assets` when crafting a release.

fschade:
@kulmann fine for me, one downside would be that pull-assets also includes `assets/core/origin/*` (`assets/core/origin/node_modules`) in the binary the way it is now... we should be really careful to not use it for releases...

plan b, keep the pulled assets out of assets/core.

what do you think?

kulmann:
> @kulmann fine for me, one downside would be that pull-assets also includes `assets/core/origin/*` (`assets/core/origin/node_modules`) in the binary the way it is now... we should be really careful to not use it for releases...
> 
> plan b, keep the pulled assets out of assets/core.
> 
> what do you think?

I don't fully understand what the issue is... a release should anyway always include a `make clean` as pre-requisite, shouldn't it?!

fschade:
@kulmann I've made some updates based on your feedback. Specifically, I've split the ci-go-generate and ci-node-generate Makefile steps into separate prod and dev targets. Additionally, web assets are now built outside of the bin filesystem and are no longer included in it.

Could you please re-check the changes and let me know if there's anything else that needs adjustment?

Thanks!

kulmann:
With the changed scope of this PR I think it would be best if @micbar and/or @dragonchaser could review it.

Thank you for the changes @fschade ! ❤️

.make/default.mk
```diff
@@ -0,0 +1,31 @@
+.PHONY: node-generate-dev-default
+node-generate-dev-default: node-generate-prod
+
+.PHONY: node-generate-prod-default
```
dragonchaser, line 4: can we remove this, it does not do anything

.make/deprecate.mk
dragonchaser, line 1: what is the use of this file? please add a comment, otherwise it gets confusing, if these targets are deprecated => rename file accordingly

dragonchaser (changes requested):

.make/default.mk
fschade: it does, this allows the overwrite of steps, check the % at the end, this is a catch all which then triggers the defaults.

that way we can overwrite and don’t have to define empty steps in each Makefile

.make/deprecate.mk
fschade: this keeps the previous make tasks intact, we already have some users out there, with that file it’s an drop in replacement

rhafer (approved):
lgtm

dragonchaser (approved):

Issue #282, opened 2025-02-28
posix. user cannot create GDPR export


ScharfViktor:
## Steps to reproduce
1. run opencloud with posix
2. go to settings and request GDPR export
3. wait

## Expected behavior
GDPR export is created

## Actual behavior
GDPR export is not created

<img width="1027" alt="Image" src="https://github.com/user-attachments/assets/c246172d-e93e-4141-a788-ed449c591055" />

How to test:

```
make test-acceptance-api \
   TEST_SERVER_URL=https://opencloud-server:9200 \
   BEHAT_FEATURE=tests/acceptance/features/apiGraph/userGDPRExport.feature:11
```

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OC_XXX=somevalue
OC_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

v-scharf:
```
apiGraph/userGDPRExport.feature:11
  apiGraph/userGDPRExport.feature:75
  apiGraph/userGDPRExport.feature:183
  apiGraph/userGDPRExport.feature:308
  apiGraph/userGDPRExport.feature:413
  apiGraph/userGDPRExport.feature:523
  apiGraph/userGDPRExport.feature:587
  apiGraph/userGDPRExport.feature:716
  apiGraph/userGDPRExport.feature:819
  apiGraph/userGDPRExport.feature:862
  apiGraph/userGDPRExport.feature:946
```

butonic:
we do not seem to create the personal space in time. when adding a 
```
And user "Alice" has stored id of folder "/"
```

the test succeeds 🤔

v-scharf:
I also tested it manualy.
- create user
- user logs in (personal space exists)
- reuests Gdpr export and wait

Actual: nothing. The GDPR export is not created

PR #297, opened 2025-03-03
look up personal space for GDPR export

butonic:
We cannot assume the personal space of a user always matches his user id ... that would be nice, but is not a given. POSIX for example has to assign a random uuid, which is why the GPDR tests failed.

fixes https://github.com/opencloud-eu/opencloud/issues/282

@ScharfViktor  With this PR I still have four failing tests:
```
apiGraph/userGDPRExport.feature:75
apiGraph/userGDPRExport.feature:308
apiGraph/userGDPRExport.feature:413
apiGraph/userGDPRExport.feature:587
```

AFAICT because locally my admin user has an id `some-admin-user-id-0000-000000000000`
```
      JSON Schema validation failed:
      1. properties.event.properties.Executant.properties.opaque_id:
         - "some-admin-user-id-0000-000000000000" does not match to ^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$
```

so the tests expect a uuid ... which strictly speaking is not necessary

same goes for the provider id:
```
      JSON Schema validation failed:
      1. properties.event.properties.ShareID.properties.opaque_id:
         - "storage-users-1:4328e170-c41a-4f4e-a490-e965125752fa:4c2ad83b-d3cf-4c12-842f-6224a4fcc27d" does not match to ^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}:[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-4[0-9A-Fa-f]{3}-[89ABab][0-9A-Fa-f]{3}-[0-9A-Fa-f]{12}$
```

Strictly speaking, none of the ids must be a uuid. We could have an implementation that uses ints. 

If the testsuite rally must verify an id property, I would just check that the id parts are non empty.

@ScharfViktor lut us get this in first, then you can update theid checks in the tests in a followup PR.

### DIFF
--- a/services/graph/pkg/service/v0/personaldata.go
+++ b/services/graph/pkg/service/v0/personaldata.go
@@ -16,6 +16,7 @@ import (
 	user "github.com/cs3org/go-cs3apis/cs3/identity/user/v1beta1"
 	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	storageprovider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	"github.com/go-chi/chi/v5"
 	ehmsg "github.com/opencloud-eu/opencloud/protogen/gen/opencloud/messages/eventhistory/v0"
 	ehsvc "github.com/opencloud-eu/opencloud/protogen/gen/opencloud/services/eventhistory/v0"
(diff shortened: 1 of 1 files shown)


ScharfViktor (approved):
image

GDPR tests are green

PR #298, opened 2025-03-03
First step towards a HA docker-compose setup (2 storage-user services)

aduffeck:
/cc @micbar @butonic


rhafer:
I am closing this. IIRC we decided to not implement this with docker-compose. If that is wrong, please re-open and put it into the `dev-tools` directory.

PR #306, opened 2025-03-04
[docs-only] Add the bare metal simple install script as example

dragotin:
This adds my simple install script to the main code repo. It is referenced there from the opencloud-eu website.

### DIFF
--- a/deployments/examples/bare-metal-simple/README.md
+++ b/deployments/examples/bare-metal-simple/README.md
@@ -1,0 +1,44 @@
+# Simple Bare Metal Install
+
+This install script downloads and installs the OpenCloud binary and
+configures it in a sandbox directory in the folder where you called
+the install script. It also adds a start script called `runopencloud.sh`
+to start OpenCloud later.
+
+The installation only consists of the bare minimum functionality 
+without web office and other optional components. Also, it is bound
+to localhost and has no valid certificates. ** It is only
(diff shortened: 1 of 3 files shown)


deployments/examples/bare-metal-simple/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+OC_VERSION="1.0.0" ./install.sh
+``` 
+to install the OpenCloud version 1.0.0
+
+There is also a hosted version of this script that makes it even
+easier:
+
+```
+curl -L https://opencloud.eu/oc-install.sh | bash -x
```
tbsbdr: Do you see a benefit in the `oc-` prefix?
I'd suggest to keep it simple and memorable. It would be useful if one could remember it from memory and eg. should over in the suse bus like "hey dirk, enter `körl -L opencloud.eu/install.sh`"

tbsbdr:
> I couldn't find this script, does it exist?

`https://opencloud.eu/oc-install.sh` will be a 301 redirect to the script on github; the redirect is not yet fully estsablished.

deployments/examples/bare-metal-simple/README.md
dragotin: Ok, correct.

ScharfViktor (approved):

tbsbdr (approved):

PR #344, opened 2025-03-11
Reintroduce check for go before including bingo Makefile

rhafer:
This re-adds the check for go being installed before including the bingo variables make file to avoid repeating errors about missing a missing go binary when running 'make node-generate' in the ci (the node container doesn't have go installed)

### DIFF
--- a/opencloud/Makefile
+++ b/opencloud/Makefile
@@ -7,7 +7,9 @@ ifdef ENABLE_VIPS
 	TAGS := ${TAGS},enable_vips
 endif
 
+ifneq (, $(shell command -v go 2> /dev/null)) # suppress `command not found warnings` for non go targets in CI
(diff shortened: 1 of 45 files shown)


rhafer:
@individual-it This should us get rid of the `bash: go: command not found` errors (like in https://ci.opencloud.eu/repos/3/pipeline/152/26)

fschade:
@rhafer cant we just add it to the top of `.make/go.mk` or `.make/default.mk`?

rhafer:
> @rhafer cant we just add it to the top of `.make/go.mk` or `.make/default.mk`?

I tried to add it to `.make/default.mk` but that make the recursive `make generate` run unbearable slow. I didn't investigate why.

fschade (approved):

PR #385, opened 2025-03-14
fix: make re-build node artifacts

fschade:
## Description
$OC_HOME/opencloud/Makefile contains `node-generate-prod` which is called by `node-generate-dev` (`*-dev` calls `*-prod` if undefined) which then calls all `*-prod` in each service... which overwrites the assets in web.

sounds complicated which it isnt, `porp ™` (a plain old recursion problem) 🤣 , we have multiple ways to fix it:

1. remove the fallback `*-prod` if no `*-dev` step is defined
1. remove the `node-generate-prod` loop in $OC_HOME/opencloud/Makefile

i decided for NR. 2 because I still think it is practical to have that fallback.


kulmann (approved):
works now, thank you so much! 😍

Issue #192, opened 2025-02-12
Default posixfs setup breaks when user is renamed


rhafer:
## Describe the bug

The default configuration for the posix driver uses the `./users/<username>` as the foldername of the personal space on disk. This means that when the user is renamed it will loose access to the home space.

## Steps to reproduce
1. Login in as user A  to have the personal space created
2. Upload some file
3. Logout again
4. Login as admin and change the username of A to B
5. Login is as user B -> a new Personal Space is created (you might need to use an incognito window here) the old files are "lost"

## Additional context

To fix this we should consider using the user id has the foldernames. Even if the `ls -l storage/users` is not as beautiful when it show uuids
instead of usernames.

Trying to rename the folder when the user is rename is not really an option as we are not always aware when a username changes (think of external LDAP setups).

PR #401, opened 2025-03-17
posixfs: Use userid as the foldername for personal space

rhafer:
This avoids loosing the user's personal space after renaming the user.

Closes: #192


rhafer:
@aduffeck, @butonic  I am not sure what do do with the space Alias. Could we just update it after the (changed) user logs in the next time? Or do client rely on the alias to be stable?

rhafer:
> @aduffeck, @butonic I am not sure what do do with the space Alias. Could we just update it after the (changed) user logs in the next time? Or do client rely on the alias to be stable?

Actually the settings for the space alias is the same as for decomposed fs. I guess we can just keep it as it is for now.

micbar (approved):

PR #405, opened 2025-03-17
run test with posix

ScharfViktor:
While `posix` is not running by default I added the ability to run tests with it if you add a `[posix]` tag to the title
if no `[posix]` in the title -> `decomposed` fs

when we get green CI - I make posix by default

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -521,11 +521,15 @@ def testPipelines(ctx):
     if config["litmus"]:
         pipelines += litmus(ctx, "decomposed")
 
+    storage = "decomposed"
+    if "[posix]" in ctx.build.title.lower():
+        storage = "posix"
+
     if "skip" not in config["cs3ApiTests"] or not config["cs3ApiTests"]["skip"]:
-        pipelines.append(cs3ApiTests(ctx, "ocis", "default"))
+        pipelines.append(cs3ApiTests(ctx, storage, "default"))
     if "skip" not in config["wopiValidatorTests"] or not config["wopiValidatorTests"]["skip"]:
-        pipelines.append(wopiValidatorTests(ctx, "ocis", "builtin", "default"))
-        pipelines.append(wopiValidatorTests(ctx, "ocis", "cs3", "default"))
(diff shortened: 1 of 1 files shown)


v-scharf:
CI run with `posix` @aduffeck please see failed tests here https://ci.opencloud.eu/repos/3/pipeline/440/

the last commit I ran with decomposed to get green CI

v-scharf:
hmm, my last commit was without [posix] but I see posix in CI. @individual-it is that cache issue?

individual-it:
@ScharfViktor I don't think its a cache issue, your PR title and commit title is `run test with posix` so `if "posix" in ctx.build.title.lower():` will be `True`

aduffeck (approved):

v-scharf:
> @ScharfViktor I don't think its a cache issue, your PR title and commit title is `run test with posix` so `if "posix" in ctx.build.title.lower():` will be `True`

I thought posix had to be as tag, like `[posix] run test with posix` ? Isn't it enough just to mention `posix` in the title?

v-scharf:
>  Isn't it enough just to mention `posix` in the title?
looks like it enough

then make as `if "[posix]" in ctx.build.title.lower()` - to expect `[posix]` in the title

.woodpecker.star
```diff
@@ -947,7 +955,7 @@ def localApiTestPipeline(ctx):
 
 def localApiTests(ctx, name, suites, storage = "decomposed", extra_environment = {}, with_remote_php = False):
     test_dir = "%s/tests/acceptance" % dirs["base"]
-    expected_failures_file = "%s/expected-failures-localAPI-on-%s-storage.md" % (test_dir, storage)
+    expected_failures_file = "%s/expected-failures-localAPI-on-decomposed-storage.md" % (test_dir)
```
individual-it, line 958: I know I am too late to review, but I think we will be in trouble here if we hard-code the file-name.
If that is the correct name for all storages, maybe better to remove the storage from the file name

v-scharf: np, I'll do it in the next PR

Issue #392, opened 2025-03-14
Edition: Don't identify as "Community" or "Enterprise"


tbsbdr:
## Steps to reproduce
1. look at the bottom left of the web ui
2. it says "Community" oder "Enterprise"

## Expected behavior
- Don't show "Community" or "Enterprise" (just leave it blank and show only the technical version)

(leave the option to show an edition - we could eg. reflect here if someone has a commerical suppot contact, but community/enterprise is just a wrong open core legacy from o*ncloud)

## Actual behavior
- "Community" or "Enterprise" is shown

![Image](https://github.com/user-attachments/assets/d4411620-8444-42ef-a12f-8fa73c447815)

tbsbdr:
Assesing as P2 as this misleading and is harmful for communication - there are no 2 different software versions of opencloud.

PR #421, opened 2025-03-18
feat: clear edition fallback

JammingBen:
Clears the `edition` fallback on "Community" in the capabilities. We want to have this as empty string without any fallback.

closes https://github.com/opencloud-eu/opencloud/issues/392


rhafer:
@TheOneRing IIRC one of the client, I believe the IOS one depended on a specific value of the `Edition`in the capabilities for certain features. Do you know if it breaks stuff if we change it to ""?

TheOneRing:
> @TheOneRing IIRC one of the client, I believe the IOS one depended on a specific value of the `Edition`in the capabilities for certain features. Do you know if it breaks stuff if we change it to ""?

Our clients won't rely on that string.

rhafer (approved):

PR #433, opened 2025-03-19
Switch to jsoncs3 backend for app tokens and enable service by default

rhafer:
Switch the default config of the app-auth service to use the jsoncs3 backend. We generate password base on the diceware algorithm by default using the default length of 6 words.

This also enables the app token service by default.  :exclamation:


rhafer:
For making delete app token request work in the UI we need this reva change: https://github.com/opencloud-eu/reva/pull/119

rhafer:
@kulmann As discussed, this now also allows setting the label on the app password.

```
 curl -k -u alan --request POST 'https://localhost:9200/auth-app/tokens?expiry=30m&label=blah%20blub'
```

aduffeck (approved):

PR #434, opened 2025-03-19
Completely remove "edition" from capabilities

rhafer:
This removes the "edition" value for the capabities. We don't need that anymoreo.

related reva change: https://github.com/opencloud-eu/reva/pull/115


.woodpecker.star
```diff
@@ -2260,7 +2260,6 @@ def skipIfUnchanged(ctx, type):
         "README.md",
     ]
     unit = [
-        "**/*_test.go",
```
micbar: What is the reasoning behind that change?

rhafer: It's just a test. We're experiencing issues with the `exlcude` feature since your latest change.

micbar (approved):

PR #438, opened 2025-03-19
DNM test exclude/include condition

rhafer:
This is just for show-casing some weird CI behavior.

Something is weird with the `exclude` patterns on the `when` `path` condition. This PR changes the some go code and the README. So it should run the full blown test suite. However most if the test a skipped.

As soon as I remove the README change all test are enabled. :dizzy_face: 

cc @micbar

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -2251,18 +2251,12 @@ def skipIfUnchanged(ctx, type):
         return []
 
     base = [
-        ".github/**",
-        ".vscode/**",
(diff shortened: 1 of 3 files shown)


rhafer:
If I understand this correctly:
https://github.com/woodpecker-ci/woodpecker/blob/main/pipeline/frontend/yaml/constraint/constraint.go#L370

The pipeline is skipped as soon as one of the excludes matches. I must admit I am not sure how useful that feature is.

PR #465, opened 2025-03-24
add keycloak example

ScharfViktor:
Steps: 
`cd deployments/examples/opencloud_full`
enable keycloak in the `.env` file https://github.com/opencloud-eu/opencloud/pull/465/files#diff-57b97469284982c33a6c68511b9b7524e5142e7bfce6867b1186b457f74a0c56R242
`OC_DOCKER_IMAGE=opencloudeu/opencloud OC_DOCKER_TAG=2.0.0 docker compose up -d`

add `127.0.0.1       keycloak.opencloud.test` to `/etc/hosts`
open https://keycloak.opencloud.test/ admin:admin

image

open https://cloud.opencloud.test/ admin:admin


micbar:
@rhafer what do you think? Maybe add this to the opencloud_full example?

rhafer:
> @rhafer what do you think? Maybe add this to the opencloud_full example?

Yes. Let's try that.

v-scharf:
> Yes. Let's try that.

then I'll add it later (after release is done 😀)

amrita-shrestha:
> @ScharfViktor, everything i try to login to opencloud. I came across this error

please try to delete this line https://github.com/opencloud-eu/opencloud/pull/465/files#diff-40bc490c0c8e661bd131829ba24af60db667dda9fe7c5e3566433a95a6e53666R83  and start again

v-scharf:
@micbar @rhafer I have added `keycloak` to `opencloud_full`
works fine for me. please have look

micbar (approved):

HiFallMaple:
Hi,

After testing, I found that the following line in the csp.yaml file of the opencloud_full example:

```yaml
- 'https://${KEYCLOAK_DOMAIN|keycloak.opencloud.test}/'
```
does not work as expected because the docker-compose file for OpenCloud does not pass the environment variable. As a result, even if KEYCLOAK_DOMAIN is defined in the .env file, the CSP configuration still falls back to keycloak.opencloud.test due to the missing environment variable.

To resolve this, I added the following lines to keycloak.yml:

```yaml
opencloud:
  environment: 
    KEYCLOAK_DOMAIN: ${KEYCLOAK_DOMAIN:-keycloak.opencloud.test}  
```
I'm not sure if this is the best approach. Should the original example also be updated accordingly?

v-scharf:
> I'm not sure if this is the best approach. Should the original example also be updated accordingly?

Thank you very much. Your approach is working perfectly 😍. I've created a fix

Issue #485, opened 2025-03-25
listing a space shared to many users takes too long


butonic:
When running the testsuite I created 600 users. this causet a /me/drives request to take 14sec:
 
![Image](https://github.com/user-attachments/assets/5f10ae15-6bf7-4cbd-b4e7-515f810414d0)

I assume because the request has to fetche all recipients ... they are cached afterwards, but still ... we should investigate that and only fetch the recipients if they were requested?

butonic:
The GetUser calls happen in `cs3SpacePermissionsToLibreGraph` when sequentially iterating over all the permissions and calling `userIdToIdentity`.

We could:
- cache the display names and user types longer ... invalidate them when a user logs in? using events?
- only render the root.drive.permissions if they are requested with a `$select` query parameter
- concurrently look up users

The displayname und usertype is not really required in the web ui until a space is selected, right?

Does the `https://opencloud-server:9200/graph/v1beta1/me/drives?$orderby=name asc&$filter=driveType eq project` call really need to return the `permissions` property right away? does it need the `root` or can we fetch that part of the response in a subsequent call?
```json
{
    "value": [
        {
            "driveAlias": "project/cdperf",
            "driveType": "project",
            "id": "storage-users-1$607687cf-11bc-4827-9d3d-4a224e23576d",
            "lastModifiedDateTime": "2025-03-25T17:28:40.244872049+01:00",
            "name": "cdperf",
            "owner": {
                "user": {
                    "displayName": "",
                    "id": "607687cf-11bc-4827-9d3d-4a224e23576d"
                }
            },
            "quota": {
                "remaining": 878984308,
                "state": "normal",
                "total": 1000000000,
                "used": 121015692
            },
            "root": {
                "eTag": "\"4cb89ccba8e0917d70d049f3f3efc57a\"",
                "id": "storage-users-1$607687cf-11bc-4827-9d3d-4a224e23576d",
                "permissions": [
                    {
                        "grantedToV2": {
                            "user": {
                                "@libre.graph.userType": "Member",
                                "displayName": "perf-test-user-210",
                                "id": "0a5e9a0e-39ce-4b97-8e74-672c916e6001"
                            }
                        },
                        "id": "u:0a5e9a0e-39ce-4b97-8e74-672c916e6001",
                        "roles": [
                            "58c63c02-1d89-4572-916a-870abc5a1b7d"
                        ]
                    },
                    {
                        "grantedToV2": {
                            "user": {
[...]

PR #495, opened 2025-03-26
do not automatically expand drive root permissions

butonic:
We no longer automatically expand the root permissions when listing drives. Part of an investigation into https://github.com/opencloud-eu/opencloud/issues/485

AFAICT the web ui already does a dedicated GET request ffor the permissions when viewing the participants of a space. The current effort of fetching the permissions is thus a waste of time and resources.

To expand the permissions send a request like this:
```
curl 'https://opencloud-server:9200/graph/v1.0/me/drives?$expand=root($expand=permissions)' -kv -u admin:admin
curl 'https://opencloud-server:9200/graph/v1.0/users/admin?$expand=drive($expand=root($expand=permissions))' -kv -u admin:admin
```

Be aware that expanding the permissions can currently still take seconds if hundreds of user identities have to be fetched because they have not been cached.

### DIFF
--- a/services/graph/pkg/odata/odata.go
+++ b/services/graph/pkg/odata/odata.go
@@ -34,10 +34,10 @@ func collectExpandPaths(item *godata.ExpandItem, prefix string) ([]string, error
 
 	// Build the current path
 	currentPath := prefix
-	if len(item.Path) > 0 {
-		if len(item.Path) > 1 {
-			return nil, godata.NotImplementedError("multiple segments in $expand not supported")
-		}
+	if len(item.Path) > 1 {
+	   return nil, godata.NotImplementedError("multiple segments in $expand not supported")
+	}
+	if len(item.Path) == 1 {
 		if currentPath == "" {
 			currentPath = item.Path[0].Value
 		} else {


butonic:
the failing e2e tests happen because the web ui no longer shows the input field to add members:
grafik


however the request (does not match the screenshot ... irrelevant to this problem)
```http
GET https://opencloud-server:9200/graph/v1beta1/drives/storage-users-1%249b9f9ab5-e5bd-407b-be20-681fccb63389/root/permissions
```
contains the exact same response as without this PR:
```json
{
    "@libre.graph.permissions.actions.allowedValues": [
        "libre.graph/driveItem/permissions/create",
        "libre.graph/driveItem/children/create",
        "libre.graph/driveItem/standard/delete",
        "libre.graph/driveItem/path/read",
        "libre.graph/driveItem/quota/read",
        "libre.graph/driveItem/content/read",
        "libre.graph/driveItem/upload/create",
        "libre.graph/driveItem/permissions/read",
        "libre.graph/driveItem/children/read",
        "libre.graph/driveItem/versions/read",
        "libre.graph/driveItem/deleted/read",
        "libre.graph/driveItem/path/update",
        "libre.graph/driveItem/permissions/delete",
        "libre.graph/driveItem/deleted/delete",
        "libre.graph/driveItem/versions/update",
        "libre.graph/driveItem/deleted/update",
        "libre.graph/driveItem/basic/read",
        "libre.graph/driveItem/permissions/update",
        "libre.graph/driveItem/permissions/deny"
    ],
    "@libre.graph.permissions.roles.allowedValues": [
        {
            "@libre.graph.weight": 1,
            "description": "View and download.",
            "displayName": "Can view",
            "id": "a8d5fe5e-96e3-418d-825b-534dbdf22b99"
        },
        {
            "@libre.graph.weight": 2,
            "description": "View, download, upload, edit, add, delete including the history.",
            "displayName": "Can edit",
            "id": "58c63c02-1d89-4572-916a-870abc5a1b7d"
        },
        {
            "@libre.graph.weight": 3,
            "description": "View, download, upload, edit, add, delete and manage members.",
            "displayName": "Can manage",
            "id": "312c0871-5ef7-4b3a-85b6-0e4074c64049"
        }
    ],
    "value": [
        {
            "grantedToV2": {
                "user": {
                    "@libre.graph.userType": "Member",
                    "displayName": "Admin",
                    "id": "some-admin-user-id-0000-000000000000"
                }
            },
            "id": "u:some-admin-user-id-0000-000000000000",
            "roles": [
                "312c0871-5ef7-4b3a-85b6-0e4074c64049"
            ]
        }
    ]
}
```

how does the web UI decide wether or not to show the member input element?

butonic:
- [ ] the admin ui counts the members ... we could support the a `count()` function to count the numbers of permissions
`https://graph.microsoft.com/v1.0/me/drive/root/permissions?$count=true&$top=0` returns
```json
{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users('48d31887-5fad-4d73-a9f5-3c356e68a038')/drive/root/permissions",
    "@odata.count": 1,
    "value": []
}
```

The odata [example](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SystemQueryOptioncount) has `GET http://host/service/Categories?$expand=Products($count=true)`


but I fail to build a proper query:
```
https://graph.microsoft.com/v1.0/me/drives?$expand=root($expand=permissions($count=true&$top=0))

{
    "error": {
        "code": "BadRequest",
        "message": "Parsing OData Select and Expand failed: Found an unbalanced bracket expression.",
        "innerError": {
            "date": "2025-03-28T08:55:42",
            "request-id": "558ad745-a520-4be1-abf1-5bdddb22aef4",
            "client-request-id": "558ad745-a520-4be1-abf1-5bdddb22aef4"
        }
    }
}
```

```
https://graph.microsoft.com/v1.0/me/drives?$expand=root($expand=permissions($count=true,$top=0))

{
    "error": {
        "code": "BadRequest",
        "message": "Parsing OData Select and Expand failed: Count option must be a boolean value, it is set to 'true,$top=0' instead.",
        "innerError": {
            "date": "2025-03-28T08:54:28",
            "request-id": "2d759b7f-03f1-46db-abac-569027b6419b",
            "client-request-id": "2d759b7f-03f1-46db-abac-569027b6419b"
        }
    }
}
```
 
```
https://graph.microsoft.com/v1.0/me/drives?$expand=root($expand=permissions($count=true))

{
    "error": {
        "code": "invalidRequest",
        "message": "Can only provide expand and select for expand options",
        "innerError": {
            "date": "2025-03-28T08:53:22",
            "request-id": "d0b45b26-d5e3-4ff5-965f-de2f0ec5e3f5",
            "client-request-id": "d0b45b26-d5e3-4ff5-965f-de2f0ec5e3f5"
        }
    }
}
```

JammingBen:
> how does the web UI decide wether or not to show the member input element?

This happens based on the space permissions. Since they're not expanded anymore with this PR, Web fails to do this check. I expect a few more places/things to be broken with this PR in Web... needs some deeper investigation where the permissions are being used.

services/graph/pkg/service/v0/drives.go
```diff
@@ -324,43 +362,57 @@ func (g Graph) canCreateSpace(ctx context.Context, ownPersonalHome bool) bool {
...
+	webDavBaseURL, err := g.getWebDavBaseURL()
+	if err != nil {
+		log.Error().Err(err).Msg("could not create drive: error parsing webdav base url")
+		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	log = log.With().Str("url", webDavBaseURL.String()).Logger()
+
+	_, expandPermissions, err := parseDriveRequest(r)
```
rhafer: Hm, `$expand` on POST. Do we really want to support this?

rhafer:
> > how does the web UI decide wether or not to show the member input element?
> 
> This happens based on the space permissions. Since they're not expanded anymore with this PR, Web fails to do this check. I expect a few more places/things to be broken with this PR in Web... needs some deeper investigation where the permissions are being used.

For that you'd "only" need to know the permission of the specific user that is doing the query, right? It would be a bit of a hack, but:

`GET /v1beta1/drives/:drive-id/items/:item-id/permissions/:perm-id`

with `:perm-id` set to `u:<userid>` could work for that. 

:thinking: unless the user has access via group membership

We could maybe add a special permission id `me` or `self` that would also return group permissions for groups that onself is a member of.

`GET /v1beta1/drives/:drive-id/items/:item-id/permissions/self`

Note: `GET /v1beta1/drives/:drive-id/items/:item-id/permissions/:perm-id` is currently not implemented in the graph service it seems.

services/graph/pkg/service/v0/drives.go
butonic: We need to support the syntax on the url to express that a client needs this information. Or the client needs to make a subsequent request.

We don't need to support expanding all relationships, we can limit it to permissions and document the behavior.

rhafer: POST creates a space. We don't support to add permission with the POST. Why should we be able to $expand permissions here?

butonic: If the web ui requires it we can give them the information in the response ... or they just assume that the current user has become the manager ...

JammingBen:
> For that you'd "only" need to know the permission of the specific user that is doing the query, right?

That should be sufficient, yes. Other memberships are only necessary for the display in the right sidebar, and for that we request the permissions anyways.

For group memberships we also need a solution as you said.

> It would be a bit of a hack, but:
> 
> `GET /v1beta1/drives/:drive-id/items/:item-id/permissions/:perm-id`
> 
> with `:perm-id` set to `u:<userid>` could work for that.

That would mean we need to call this for each space the current user is a member of in the spaces overview. Hmm. Or maybe we can only do it when selecting a specific space. Not sure about the feasibility and the implications though. It could mean quite a bit of refactoring.

In general, as proposed above, we should start in Web by identifying the places where we rely on permissions and then think about how to get this information. I'd be happy to help here since I implemented most of the sharing stuff in Web.

services/graph/pkg/service/v0/drives.go
JammingBen: > or they just assume that the current user has become the manager ...

While it would feel a bit hacky just assuming this, I guess we could do it. I don't see any other scenario where you create a space without being manager.

butonic: hm. do we need the actuae permission or only our `@libre.graph.permissions.actions.allowedValues` and  `@libre.graph.permissions.roles.allowedValues`. we expand the `"@libre.graph.permissions` properties but leave out the actual permissions objects?

butonic:
From what I can tell the UI representation for a space does not differ between a manager or a viewer.

The viewer has a different context menu though:
grafik
vs
grafik

we could wait until the user hovers over the three dots or the space before fetching the permissions...
and we could assume that if we are the owner of a space we can edit it

JammingBen:
> The viewer has a different context menu though:

Yep that's the key point here. The actions for the context menu are not being loaded on click but when loading the list. Hence we (currently) need to the permissions right from the start. Not sure if/how we can change that.

> we could wait until the user hovers over the three dots or the space before fetching the permissions...

I'd rather do this when selecting the space. This automatically happens when e.g. opening the context menu. Either way, a mechanism like that doesn't exist yet. It would most likely require quite some refactoring in Web.

services/graph/pkg/service/v0/drives.go
rhafer: I have the feeling we should leave the `POST /v1.0/drives` implementation alone for now. It seems to return some pretty broken response in the permissions anyway:

```
"permissions": [
      {
        "grantedToIdentities": [
          {
            "user": {
              "@libre.graph.userType": "Member",
              "displayName": "Admin",
              "id": "b75115e1-db1f-48f0-aeb0-87517bf3abf5"
            }
          }
        ],
        "grantedToV2": {
          "user": {
            "@libre.graph.userType": "Member",
            "displayName": "Admin",
            "id": "b75115e1-db1f-48f0-aeb0-87517bf3abf5"
          }
        },
        "roles": [
          "manager"
        ]
      }
    ],
```

@JammingBen  Do you really use anything from that currently? I somehow think we should properly implement `POST /v1beta1/drives`  and move forward.

BTW, the odata spec if not really clear on whether $expand is allowed on POST. From https://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part2-url-conventions/odata-v4.0-errata03-os-part2-url-conventions-complete.html#_Toc453752357: 

"The semantics of system query options applied to POST requests to entity sets as well as all PATCH, PUT and DELETE requests are not defined by this specification and are reserved for future versions."

JammingBen: > @JammingBen Do you really use anything from that currently? I somehow think we should properly implement `POST /v1beta1/drives` and move forward.

We don't need any of that when creating a drive. The same goes for the `$expand` on a POST, no need for that.

services/graph/pkg/odata/odata.go
```diff
@@ -0,0 +1,104 @@
...
+	if len(item.Path) > 0 {
+		if len(item.Path) > 1 {
+			return nil, godata.NotImplementedError("multiple segments in $expand not supported")
+		}
+		if currentPath == "" {
+			currentPath = item.Path[0].Value
+		} else {
+			currentPath += "." + item.Path[0].Value
+		}
+	}
```
aduffeck: Nitpick to avoid nested `if`s
```suggestion
	if len(item.Path) > 1 {
	   return nil, godata.NotImplementedError("multiple segments in $expand not supported")
	}
	if len(item.Path) == 1 {
		if currentPath == "" {
			currentPath = item.Path[0].Value
		} else {
			currentPath += "." + item.Path[0].Value
		}
	}
```

aduffeck (approved):
lgtm

PR #506, opened 2025-03-27
update generated code

butonic:
I ran `make generate` ... this is the diff.

Don't we have a CI step that checks if generated code was updated?


fschade (approved):
`make generate`:

- new folder: docs/apis/
- changes in: protogen/gen/opencloud/messages/settings/v0/settings.pb.go (removed yaml tag)

should i commit those?

PR #565, opened 2025-04-03
Simplify item-trashed SSEs. Also fixes it for coll. posix fs.

aduffeck:


micbar:
@aduffeck please backport to `stable-2.0`

aduffeck:
@micbar is that really something we need backported? it really only changes the behavior for "collaborative" deletes and collaborative mode wasn't part of 2.0 yet.

PR #569, opened 2025-04-04
example Ldap setup

ScharfViktor:
## Description

- Adds an LDAP option to use bitnami OpenLDAP. This is recommended for **production** use
- Adds an LDAP manager UI

ScreenShot Tool -20250408004046

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


deployments/examples/opencloud_full/ldap.yml
```diff
@@ -0,0 +1,80 @@
...
+      LDAP_TLS_CERT_FILE: /opt/bitnami/openldap/share/openldap.crt
+      LDAP_TLS_KEY_FILE: /opt/bitnami/openldap/share/openldap.key
+      LDAP_ROOT: "dc=opencloud,dc=com"
+      LDAP_ADMIN_PASSWORD: ${LDAP_ADMIN_PASSWORD:-admin}
+    ports:
+      - "127.0.0.1:389:1389"
+      - "127.0.0.1:636:1636"
+    volumes:
+      - ./config/ldap/ldif:/ldifs
+      - ./config/ldap/schemas:/schemas
```
rhafer: Could you mount the schema from `../shared/config/ldap/schemas` here? Instead of duplicating the schema files?

deployments/examples/opencloud_full/ldap.yml
```diff
@@ -0,0 +1,80 @@
...
+      - ./config/ldap/docker-entrypoint-override.sh:/opt/bitnami/scripts/openldap/docker-entrypoint-override.sh
+      - ldap-certs:/opt/bitnami/openldap/share
+      - ldap-data:/bitnami/openldap
+    logging:
+      driver: ${LOG_DRIVER:-local}
+    restart: always
+
+  ldap-manager:
+    image: osixia/phpldapadmin:latest
```
rhafer: I think we cannot deploy this anymore, it's unmaintained since 2019. (Even the base docker image is hugely outdate by now)

But it seems phpldapadmin has been revived and is maintained again. See https://phpldapadmin.org/
They even release a docker image. It might not be a drop in replacement for osixia/phpldapadmin though.

rhafer (changes requested):

deployments/examples/opencloud_full/ldap.yml
micbar: done

deployments/examples/opencloud_full/ldap.yml
micbar: Working fine! Thanks for the hint.

v-scharf:
@micbar I checked new `phpldapadmin/phpldapadmin:latest` and noticed that user update via ldap admin (https://ldap.opencloud.test) doesn't work. Could you check it pls if it works for you?

micbar:
@rhafer I am a bit stuck. The login with DN doesn't work.

micbar:
I pushed my latest code.

rhafer:
> @rhafer I am a bit stuck. The login with DN doesn't work.

Yeah something is weird there. I've opened a bug for phpldapadmin. It seems to have some weird expectations on the LDAP server that I don't quite understand. I guess I can bend our setup so that it work with phpldapadmin, but I am not really sure it is worth it currently.

What is the goal for this PR? Do we really need an example setup with the built-in IDP but and external LDAP server. And if yes, what do we need phpldapadmin for?

I could understand we we turned this into external LDAP + Keycloak, but even that I don't think we'd need phpldapadmin.

micbar:
> I could understand we we turned this into external LDAP + Keycloak, but even that I don't think we'd need phpldapadmin.

I agree. Let us leave the phpldapadmin out for now.

micbar:
@rhafer are we sure, that we can create users in the Keycloak UI? I remember this has been a problem because of the missing but required property openCloudUUID, which Keycloak cannot set during creation.

micbar:
@rhafer I moved the Ldap manager to the folder with the schemas.

This way, one can use that for demo and debugging.

rhafer:
> @rhafer are we sure, that we can create users in the Keycloak UI? I remember this has been a problem because of the missing but required property openCloudUUID, which Keycloak cannot set during creation.

In the keycloak + ldap example I would suggest to use the `entryUUID` attribute as the userid. That attribute is maintained by the LDAP server and should always be there. Then we should be able to manage user within keycloak.

PR #559, opened 2025-04-02
feat(antivirus): add partial scanning mode

fschade:
## Description
This pull request introduces a partial mode for antivirus scanning to handle large files efficiently.

## Related Issue
- Inspired by [opencloud-eu/opencloud#529](https://github.com/opencloud-eu/opencloud/issues/529#issuecomment-2768923084)

## Dependencies
This pull request depends on the following PR being merged first:
- [opencloud-eu/reva#153](https://github.com/opencloud-eu/reva/pull/153)

## Motivation and Context
This change is required to enhance the antivirus scanning efficiency for large files. The partial mode allows handling large files by scanning parts rather than loading the entire file.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Documentation added

Issue #529, opened 2025-03-30
Uploading larger video file: Hangs on "this item is in processing"


DecaTec:
## Describe the bug
When uploading a larger video file (MOV), the upload is finished. But the file cannot be accessed, because it hangs on the status "this item is in processing".

## Steps to reproduce
1. Upload lager file (I think it doesn't have to be a video). File size: ~190 MB
2. Wait until upload finishes.

## Expected behavior
The item should be processed and the file should be accessible trough the UI.

## Actual behavior
The status hangs on "this item is in processing". File cannot be accessed (even after a longer time waiting).

Confusing for the user as no error message is shown and the file cannot be removed/deleted either.

## Setup
 "opencloud_full" (with ClamAV)

OpenCloud 2.0.0 undefined
OpenCloud Web UI 2.1.0

<details>

Logs OpenCloud
<p>

```
{"level":"error","service":"antivirus","error":"write unix @->/tmp/clamd.sock: write: broken pipe","uploadid":"48a5f497-940b-429a-809c-e49318f64452","time":"2025-03-30T16:34:25Z","line":"github.com/opencloud-eu/opencloud/services/antivirus/pkg/service/service.go:246","message":"error scanning file"}

```

Logs ClamAV
<p>

```
WARNING: Sun Mar 30 16:31:47 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 202)
WARNING: Sun Mar 30 16:31:53 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 977)
WARNING: Sun Mar 30 16:32:03 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 202)
WARNING: Sun Mar 30 16:32:24 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 202)
WARNING: Sun Mar 30 16:33:04 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 298)
WARNING: Sun Mar 30 16:34:25 2025 -> INSTREAM: Size limit reached, (requested: 1024, max: 796)

```

</p>
</details>

## Additional context
Seems to be a problem with ClamAV refusing to scan such a big file.

However, there are no config options in `.env` or `clamav.yml` raising limits.

Generally, the OpenCloud should not deny access to files which cannot be scanned by ClamAV. There should be an option to "unlock" or delete these files.

criegerde:
Increased e.g. to 512 MB in clamav.yaml


<pre>clamav:
[...]

PR #620, opened 2025-04-08
fix(antivirus | backport): introduce a default max scan size for the full example deployment

fschade:
## Description
The current full example Docker deployment lacks a maximum scan size when the virus scanner is enabled. As a result, all files are scanned regardless of size. However, since Clamav has limits set in the default setup, this combination results in orphaned files that get stuck in post-processing.

The problem has already been fixed in the current main branch. Since an additional scan mode has been integrated here, a full backport is not possible. Therefore, I have only fixed the bare minimum to resolve the bug in these installations.

## Related Issue

- Backports parts of https://github.com/opencloud-eu/opencloud/pull/559
- Issue  https://github.com/opencloud-eu/opencloud/issues/529

## Motivation and Context
Quite simply, the basic installation from our example deployment should run error-free out of the box

## How Has This Been Tested?
- local test deployment

## Types of changes
- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/antivirus/pkg/config/config.go
+++ b/services/antivirus/pkg/config/config.go
@@ -21,7 +21,7 @@ type Config struct {
 	Workers              int `yaml:"workers" env:"ANTIVIRUS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"1.0.0"`
 
 	Scanner     Scanner
-	MaxScanSize string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum scan size the virus scanner can handle. Only this many bytes of a file will be scanned. 0 means unlimited and is the default. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
+	MaxScanSize string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum scan size the virus scanner can handle. 0 means unlimited. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
 
 	Context context.Context `json:"-" yaml:"-"`
 
--- a/deployments/examples/opencloud_full/.env
+++ b/deployments/examples/opencloud_full/.env
@@ -99,7 +99,7 @@ MINIO_DOMAIN=
 #DECOMPOSED=:decomposed.yml
 
 # Define SMPT settings if you would like to send OpenCloud email notifications.
-# 
+#
 # NOTE: when configuring Inbucket, these settings have no effect, see inbucket.yml for details.
 # SMTP host to connect to.
 SMTP_HOST=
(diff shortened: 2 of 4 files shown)


fschade:
@micbar I'm totally undecided about whether we should also set a default in the code... the advantage: even non-Docker example deployments would run out of the box with the default clamav config... on the other hand, we would be able to outsmart installations that deliberately have no limit (0) and a correspondingly configured scanner, skipping files larger than the "new" default...

e.g. https://github.com/opencloud-eu/opencloud/blob/9b1a79f7d7fc816e13ceab49eb47d8fc5aea4dc2/services/antivirus/pkg/config/defaults/defaultconfig.go#L35

hellllp :)

services/antivirus/pkg/config/config.go
```diff
@@ -21,7 +21,7 @@ type Config struct {
 	Workers              int `yaml:"workers" env:"ANTIVIRUS_WORKERS" desc:"The number of concurrent go routines that fetch events from the event queue." introductionVersion:"1.0.0"`
 
 	Scanner     Scanner
-	MaxScanSize string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum scan size the virus scanner can handle. Only this many bytes of a file will be scanned. 0 means unlimited and is the default. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
+	MaxScanSize string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum scan size the virus scanner can handle. 0 means unlimited. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
```
dragotin, line 24:
```suggestion
	MaxScanSize string `yaml:"max-scan-size" env:"ANTIVIRUS_MAX_SCAN_SIZE" desc:"The maximum size the virus scanner is allowed to scan per file. Bigger files are scanned partially. 0 means unlimited. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
```

dragotin: ok, ok, just fix it in main ;-)

v-scharf:
I tested it on my local setup
stable-2.0 
opencloud binary. also tried with docker setup (clamav image:` image: mkodockx/docker-clamav:alpine`)

```
PROXY_ENABLE_BASIC_AUTH=true \
ANTIVIRUS_SCANNER_TYPE="clamav" \
ANTIVIRUS_CLAMAV_SOCKET="tcp://host.docker.internal:3310" \
POSTPROCESSING_STEPS="virusscan" \
ANTIVIRUS_MAX_SCAN_SIZE=500MB OC_ASYNC_UPLOADS=true \
LOG_LEVEL=debug OC_ADD_RUN_SERVICES="antivirus" \
```

uploading small image, pdf, txt, move 10mb files works ✅️
finds and deletes virus files ✅️

But: I cannot download: 
- move file 80Mb (screen recording on my mac)
- zip file ~ 70Mb
- other files 100Mb, 200Mb 1Gb (`mkfile 100m testfile100mb.bin`)

all files hang in postprocessing state with log: 
`{"level":"error","service":"antivirus","error":"write tcp 127.0.0.1:55547->127.0.0.1:3310: write: broken pipe","uploadid":"213952e6-819c-42f5-8345-541bcc62c6fd","time":"2025-04-11T11:51:56+02:00","message":"error scanning file"}`

Screenshot 2025-04-11 at 11 36 25

fschade:
> ANTIVIRUS_MAX_SCAN_SIZE=500MB

You set the size to 500mb, have you also adjusted the clamav configuration? If not it will fail! The changes in the back port only affect the docker compose setup and does not change any behavior in the binary

v-scharf:
> You set the size to 500mb, have you also adjusted the clamav configuration? If not it will fail! The changes in the back port only affect the docker compose setup and does not change any behavior in the binary

Thanks Flo) I figure out how to run `opencloud` with `clamav` localy on mac and can confirm that your fix. 
test docs is quite nice. I updated it

PR #659, opened 2025-04-13
fix: registry/nats-js-kv, fix ListService, add error, add testsuite.

jochumdev:
This patch changes the way nodes are registered:

Before: server.DefaultId
After: service.Nodes[0].Id

Fixes ListService:

ListService was only returning the last service it found, not all of them.

Also adds an error "registry.ErrNotFound" if there are no results in Get/List.

Everything now tested with a testsuite which I backported from go-orb.

**WARN**: I never tested Opencloud with this patch, please do so.


micbar:
@jochumdev Can you rebase please?

@butonic please review

PR #665, opened 2025-04-14
feat(web): remove old and unused color tokens

JammingBen:
These tokens are not being used anymore since we recently switched to material design.


services/web/assets/themes/opencloud/theme.json
```diff
@@ -167,46 +156,7 @@
               "icon-pdf": "#ec0d47",
               "icon-presentation": "#ee6b3b",
               "icon-spreadsheet": "#15c286",
-              "icon-video": "#045459",
```
AlexAndBear, line 170: Do we really want to remove this ?

AlexAndBear (changes requested):

services/web/assets/themes/opencloud/theme.json
JammingBen: It's not being removed, it's just being listed a new line because the comma at the end has been removed.

AlexAndBear (approved):

PR #671, opened 2025-04-15
fix deployment: do not create demo accounts when using keycloak

### LINKED ISSUE #670: moving from built idp in to keycloak causes settings service to fail
## Steps to reproduce
1. deploy docker compose full example without keycloak, login should work
2. configure external keycloak in .env
3. restart deployment with `docker compose up -d` to pick up new env

## Expected behavior
login should work

## Actual behavior
The container fails to start. `The admin user ID has not been configured for settings`... 

## Additional info
OC_ADMIN_USER_ID is set to "" when keycloak is enabled, but it was set in the initial deployment with the build in idp. Maybe the settings service stores that somewhere.

Maybe SETTINGS_SETUP_DEFAULT_ASSIGNMENTS needs to be set to false:
```go
	if cfg.SetupDefaultAssignments && cfg.AdminUserID == "" {
		return shared.MissingAdminUserID(cfg.Service.Name)
	}
```

Thet means that the env file must have DEMO_USERS=false. We could add that to the keycloak.yml

butonic:
fixes #670

### DIFF
--- a/deployments/examples/opencloud_full/keycloak.yml
+++ b/deployments/examples/opencloud_full/keycloak.yml
@@ -18,7 +18,11 @@ services:
       PROXY_USER_OIDC_CLAIM: "preferred_username"
       PROXY_USER_CS3_CLAIM: "username"
       OC_EXCLUDE_RUN_SERVICES: "idp"
+      
+      # admin and demo accounts must be created in Keycloak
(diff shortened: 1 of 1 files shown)


micbar (approved):

butonic:
or is this type DevOps?

PR #672, opened 2025-04-15
collaborative posix test

ScharfViktor:
first collaborative posix test. 
- The admin user creates a folder in the file system, and the Alice user can see that folder in their personal space

image

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -132,6 +132,7 @@ config = {
             "suites": [
                 "apiGraph",
                 "apiServiceAvailability",
+                "collaborativePosix",
             ],
             "skip": False,
             "withRemotePhp": [True],
@@ -945,6 +946,7 @@ def localApiTests(name, suites, storage = "decomposed", extra_environment = {}, 
         "OC_WRAPPER_URL": "http://%s:5200" % OC_SERVER_NAME,
         "WITH_REMOTE_PHP": with_remote_php,
         "COLLABORATION_SERVICE_URL": "http://wopi-fakeoffice:9300",
+        "OC_STORAGE_PATH": "$HOME/.opencloud/storage/users/users",
     }
 
     for item in extra_environment:
@@ -1920,6 +1922,7 @@ def opencloudServer(storage = "decomposed", accounts_hash_difficulty = 4, depend
         "WEB_DEBUG_ADDR": "0.0.0.0:9104",
         "WEBDAV_DEBUG_ADDR": "0.0.0.0:9119",
         "WEBFINGER_DEBUG_ADDR": "0.0.0.0:9279",
+        "STORAGE_USERS_POSIX_WATCH_FS": True,
     }
 
     if storage == "posix":
@@ -2001,6 +2004,8 @@ def opencloudServer(storage = "decomposed", accounts_hash_difficulty = 4, depend
             },
         },
         "commands": [
+            "apt-get update",
+            "apt-get install -y inotify-tools",
             "%s init --insecure true" % dirs["opencloudBin"],
             "cat $OC_CONFIG_DIR/opencloud.yaml",
             "cp tests/config/woodpecker/app-registry.yaml $OC_CONFIG_DIR/app-registry.yaml",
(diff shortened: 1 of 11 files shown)


tests/acceptance/bootstrap/CliContext.php
```diff
@@ -431,9 +440,45 @@ public function getJSONDecodedCliMessage(ResponseInterface $response): array {
...
 		$body = [
-			"command" => $command
+		  "command" => $command
 		];
 		$response = CliHelper::runCommand($body);
 		Assert::assertEquals("200", $response->getStatusCode(), "Failed to clean upload sessions");
 	}
+
+	/**
+	 * @When the administrator creates folder :folder for user :user
```
individual-it:
```suggestion
	 * @When the administrator creates folder :folder for user :user on the POSIX filesystem
```

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -431,9 +440,45 @@ public function getJSONDecodedCliMessage(ResponseInterface $response): array {
...
+	 */
+	public function theAdministratorCreatesFolder(string $folder, string $user): void {
+		$userUuid = $this->featureContext->getUserIdByUserName($user);
+		$storagePath = $this->getStoragePath();
+		$body = [
+		  "command" => "mkdir -p $storagePath/$userUuid/$folder",
+		  "raw" => true
+		];
+		$this->featureContext->setResponse(CliHelper::runCommand($body));
+		sleep(1);
```
individual-it, line 465: why do we need this sleep?

tests/acceptance/features/collaborativePosix/collaborativePosixFS.feature
```diff
@@ -0,0 +1,13 @@
+Feature: create a resources using collaborative posixfs
```
individual-it, line 2: should we add some tags here to make sure these tests only run when `STORAGE_USERS_POSIX_WATCH_FS` is set to `true`?

.woodpecker.star
```diff
@@ -1920,6 +1922,7 @@ def opencloudServer(storage = "decomposed", accounts_hash_difficulty = 4, depend
         "WEB_DEBUG_ADDR": "0.0.0.0:9104",
         "WEBDAV_DEBUG_ADDR": "0.0.0.0:9119",
         "WEBFINGER_DEBUG_ADDR": "0.0.0.0:9279",
+        "STORAGE_USERS_POSIX_WATCH_FS": True,
```
individual-it: could be moved into `if storage == "posix":` because decomposefs and other storages don't need it

tests/acceptance/bootstrap/CliContext.php
```diff
@@ -431,9 +440,45 @@ public function getJSONDecodedCliMessage(ResponseInterface $response): array {
...
+	}
+
+	/**
+	 * @When the administrator checks storage folder for user :user
```
individual-it:
```suggestion
	 * @When the administrator lists the content of the POSIX storage folder of user :user
```

tests/acceptance/config/behat.yml
```diff
@@ -458,6 +458,21 @@ default:
...
+      contexts:
+        - FeatureContext: *common_feature_context_params
+        - CliContext:
+        # - OcConfigContext:
+        # - FilesVersionsContext:
+        # - SharingNgContext:
+        # - SearchContext:
+        # - TagContext:
+        # - TrashbinContext:
+        # - SpacesTUSContext:
```
individual-it: should be enabled again before merging

individual-it (changes requested):

tests/acceptance/config/behat.yml
v-scharf: ne, I forgot to delete it

.woodpecker.star
v-scharf: deleted from .woodpecker.star 
now I set this env in the test using wrapper

tests/acceptance/bootstrap/CliContext.php
v-scharf: image

I think a minimum wait is needed here because the api doesn't return the created resource.

individual-it: hmm could loop checking if the API sees the file, I think that is how we do it for async uploads. Or we keep the sleep for now and wait for https://github.com/opencloud-eu/opencloud/issues/718

individual-it (approved):

Issue #206, opened 2025-02-19
Calendar and Contacts Backend Integration


tbsbdr:
# Description

## User Stories

* > As a user I want to use opencloud as calendar and contacts backend so that I'm independent from bigtech.

## Acceptance Criteria
- every user of opencloud has their own non-shareable calendar and addressbook
- every user of opencloud can copy and paste the endpointaddresses to configure their calendar/contacts clients
- every user of opencloud can copy/paste the access tokens for their calendar/contacts clients
- there is an option to reset their own credentials
- calendar and contacts are enabled per default in the full docker example
- the web client can retrieve the info if the calendar and contacts features are enabled on the server (capability)
- Calendar and Contacts work with these clients:
  - Thunderbird
  - Android (via DavX)
  - MacOS and iOS vanilla calendar 

implementation details:
- authentication is done via basic auth
- please dont forget to document

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the OpenCloud charts is compatible

butonic:
To create a baikal account we could use the graph api [applications](https://learn.microsoft.com/en-us/graph/api/resources/applications-api-overview?view=graph-rest-1.0)  which actually have a [passwordCredential](https://learn.microsoft.com/en-us/graph/api/resources/passwordcredential?view=graph-rest-1.0) ... hm that seems to be a different password ... AFAICT that would be the password the opencloud would use to talk to baikal ... althoug in our case we don't roll the password  .... well we could roll it, but it would never leave the ui ...

[...]

PR #756, opened 2025-04-29
proxy(router): Allow to set some outgoing headers

rhafer:
This introduces the "additional_headers", "remote_user_header" and
"skip_x_access_token" config keys to allow configuring routes to
external services that require addtional headers to be set.

"remote_user_header": defines the name of a Header that will carry the
userid of the authenticated user on the outgoing request.

"additional_headers": defines a list of header names and values that will
be added to outgoing requests on matching routes.

"skip_x_access_token": when set to true the reva access token will not
be added to the outgoing request.

This is the groundwork for #206 I'll submit required `opencloud_full` adjustments in a separate PR.

### DIFF
--- a/pkg/middleware/account.go
+++ b/pkg/middleware/account.go
@@ -42,7 +42,7 @@ func ExtractAccountUUID(opts ...account.Option) func(http.Handler) http.Handler 
 	}
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			token := r.Header.Get("x-access-token")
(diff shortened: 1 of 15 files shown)


rhafer:
Hm, I think we should have a setting that prevents adding the reva token to the x-access-token header when forwarding to some external service.

-> going back to draft for now.

rhafer:
> I'll submit required `opencloud_full` adjustments in a separate PR.

working on that in: https://github.com/rhafer/opencloud/tree/radicale_example

aduffeck (approved):

PR #773, opened 2025-05-05
deployment: Adapt opencloud_full to include radicale

rhafer:
This adds another piece for: #206 

To enable radicale:
* uncomment the `#RADICALE=:radicale.yml` line in `.env` and deploy
* Generate an App Token
* point your caldav and carddav clients to `https://your.cloud.name/caldav` (the proxy is also forwarding `https://your.cloud.name/carddav`, technically it shouldn't matter which one is used)

Radicale is currently configured to add a default Calendar and Addressbook for every user that logs in. So that most clients should work out of the box.

For enabling the radicale WebUI e.g. to import a calendar or address books, see the notes in the proxy.yaml file.

### DIFF
--- a/deployments/examples/opencloud_full/config/opencloud/proxy.yaml
+++ b/deployments/examples/opencloud_full/config/opencloud/proxy.yaml
@@ -1,0 +1,40 @@
+# This adds four additional routes to the proxy. Forwarding
+# request on '/carddav/', '/caldav/' and the respective '/.well-knwown'
+# endpoints to the radicale container and setting the required headers.
+additional_policies:
+  - name: default
+    routes:
+      - endpoint: /caldav/
+        backend: http://radicale:5232
+        remote_user_header: X-Remote-User
+        skip_x_access_token: true
+        additional_headers:
+          - X-Script-Name: /caldav
+      - endpoint: /.well-known/caldav
+        backend: http://radicale:5232
+        remote_user_header: X-Remote-User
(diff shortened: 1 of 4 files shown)


deployments/examples/opencloud_full/config/opencloud/proxy.yaml
```diff
@@ -0,0 +1,40 @@
+# This adds two addition routes to the proxy forwarding
```
kulmann:
```suggestion
# This adds four additional routes to the proxy forwarding
```

deployments/examples/opencloud_full/config/opencloud/proxy.yaml
```diff
@@ -0,0 +1,40 @@
...
+# request on /carddav/ and /caldav/ to the radicale container
```
kulmann:
```suggestion
# request on /carddav/ and /caldav/ and the respective .well-known endpoints to the radicale container
```

kulmann (approved):
Works flawlessly now on mac with the latest opencloud-rolling daily image. Great job, thank you! 🥳

rhafer:
ran some additional test with DAVx5, Thunderbird and the Korganizer/KAddressBook. Hit a bug with the discovery of the addressbook in Thunderbird (https://bugzilla.mozilla.org/show_bug.cgi?id=1925006 fixed in nightly builds). Everything else worked fine.

PR #782, opened 2025-05-06
run ci with decomposed

ScharfViktor:
- we need to set `[decomposed]` to the title to run tests with decomposed storage - checked that it works
Screenshot 2025-05-06 at 15 40 01

- by default we run posix. But CliCommand always uses decomposed. I changed title in CI

Screenshot 2025-05-06 at 15 46 58

- localApiTests used `expected-failures-localAPI-on-decomposed-storage.md`. Fixed it

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -899,6 +899,7 @@ def localApiTestPipeline(ctx):
                 for item in defaults:
                     params[item] = matrix[item] if item in matrix else defaults[item]
                 for storage in params["storages"]:
+                    print("Using storage:", storage)
                     for run_with_remote_php in params["withRemotePhp"]:
                         pipeline = {
                             "name": "%s-%s%s-%s" % ("CLI" if name.startswith("cli") else "API", name, "-withoutRemotePhp" if not run_with_remote_php else "", storage),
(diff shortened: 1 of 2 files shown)


.woodpecker.star
```diff
@@ -901,7 +901,7 @@ def localApiTestPipeline(ctx):
                 for storage in params["storages"]:
                     for run_with_remote_php in params["withRemotePhp"]:
                         pipeline = {
-                            "name": "%s-%s%s-%s" % ("CLI" if name.startswith("cli") else "API", name, "-withoutRemotePhp" if not run_with_remote_php else "", storage),
+                            "name": "%s-%s%s-%s" % ("CLI" if name.startswith("cli") else "API", name, "-withoutRemotePhp" if not run_with_remote_php else "", "decomposed" if name.startswith("cli") else storage),
```
individual-it, line 904: I'm confused, this will set the name to `decomposed` if the name starts with `cli`? :confused:

v-scharf: yes, since we only run CliCommand with decomposed -> I changed cli name

before: 
Screenshot 2025-05-12 at 08 54 44


now: 
Screenshot 2025-05-12 at 08 54 18

individual-it: ah now I understood

individual-it (approved):

PR #806, opened 2025-05-08
fix: show special roles at the end of the list

AlexAndBear:
## Description

image

## Related Issue

- Fixes https://github.com/opencloud-eu/web/issues/648

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/graph/pkg/unifiedrole/roles.go
+++ b/services/graph/pkg/unifiedrole/roles.go
@@ -226,7 +226,28 @@ var (
 					Condition:              proto.String(UnifiedRoleConditionFolderFederatedUser),
 				},
 			},
-			LibreGraphWeight: proto.Int32(0),
+			LibreGraphWeight: proto.Int32(10),
+		}
+	}()
+
+	// roleSecureViewer creates a secure viewer role
+	roleSecureViewer = func() *libregraph.UnifiedRoleDefinition {
+		r := conversions.NewSecureViewerRole()
+		return &libregraph.UnifiedRoleDefinition{
+			Id:          proto.String(UnifiedRoleSecureViewerID),
+			Description: proto.String(_secureViewerUnifiedRoleDescription),
+			DisplayName: proto.String(cs3RoleToDisplayName(r)),
+			RolePermissions: []libregraph.UnifiedRolePermission{
+				{
+					AllowedResourceActions: CS3ResourcePermissionsToLibregraphActions(r.CS3ResourcePermissions()),
+					Condition:              proto.String(UnifiedRoleConditionFile),
+				},
+				{
+					AllowedResourceActions: CS3ResourcePermissionsToLibregraphActions(r.CS3ResourcePermissions()),
+					Condition:              proto.String(UnifiedRoleConditionFolder),
+				},
+			},
+			LibreGraphWeight: proto.Int32(20),
 		}
 	}()
 
@@ -255,7 +276,7 @@ var (
 					Condition:              proto.String(UnifiedRoleConditionFolderFederatedUser),
 				},
 			},
-			LibreGraphWeight: proto.Int32(0),
+			LibreGraphWeight: proto.Int32(30),
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


AlexAndBear:
Before caring about the tests, I'd love to know if this is matching the code standards 😼

kulmann:
> Before caring about the tests, I'd love to know if this is matching the code standards 😼

I like the approach! Building two groups (normal and special) and then sorting those groups according to the weight prop sounds reasonable and gives us the expected result. 👍

services/graph/pkg/unifiedrole/roles.go
```diff
@@ -536,35 +536,58 @@ func GetLegacyRoleName(role libregraph.UnifiedRoleDefinition) string {
...
+		UnifiedRoleDeniedID,
+	}
+
+	specialSet := make(map[string]struct{}, len(specialIDs))
+	for _, id := range specialIDs {
+		specialSet[id] = struct{}{}
+	}
+
+	normalRoles := make([]*libregraph.UnifiedRoleDefinition, 0, len(roleSet))
+	specialRoles := make([]*libregraph.UnifiedRoleDefinition, 0, len(roleSet))
```
kulmann:
```suggestion
	specialRoles := make([]*libregraph.UnifiedRoleDefinition, 0, len(specialIDs))
```

... because you can never exceed the length of the specialIDs, so this is a good upper bound.

kulmann (commented):
Code-wise this looks good to me.

rhafer (changes requested):
Sorry,  but I am not really convinced that this approach is good. What exactly makes a role "special"? And even if we think they're somehow special, I don't think we should hardcode that somewhere deep in the code.

Roles already have weights for ordering them. We just set the weight to `0` everywhere and then try some clever sorting based on the number of permissions they grant. Maybe that is the wrong approach. Why not just initialize the weights correctly from the beginning and just sort by weights?

micbar:
> Sorry, but I am not really convinced that this approach is good. What exactly makes a role "special"? And even if we think they're somehow special, I don't think we should hardcode that somewhere deep in the code.
> 
> Roles already have weights for ordering them. We just set the weight to `0` everywhere and then try some clever sorting based on the number of permissions they grant. Maybe that is the wrong approach. Why not just initialize the weights correctly from the beginning and just sort by weights?

Same question here. Why do we override the backend weight? If something is wrong, we should  change the weight in the backend.

AlexAndBear:
@tbsbdr can the gatekeeper give us their opinion

I'd rather show `Can view (secure)` under `Can view` and `Cannot access` at last.

tbsbdr:
I'd prefer the order like you propsed it.

services/graph/pkg/unifiedrole/roles.go
```diff
@@ -531,39 +531,23 @@ func GetLegacyRoleName(role libregraph.UnifiedRoleDefinition) string {
...
 // descending - true - sorts the roles from most to least permissions
 func weightRoles(roleSet []*libregraph.UnifiedRoleDefinition, constraints string, descending bool) []*libregraph.UnifiedRoleDefinition {
-	slices.SortFunc(roleSet, func(i, j *libregraph.UnifiedRoleDefinition) int {
-		var ia []string
-		for _, rp := range i.GetRolePermissions() {
-			if rp.GetCondition() == constraints {
-				ia = append(ia, rp.GetAllowedResourceActions()...)
-			}
-		}
+	slices.SortFunc(roleSet, func(a, b *libregraph.UnifiedRoleDefinition) int {
```
rhafer, line 539: I think you could replace the whole block below with:

```
if descending {
    return cmp.Compare(b.GetLibreGraphWeight(), b.GetLibreGraphWeight())
}
return cmp.Compare(a.GetLibreGraphWeight(), b.GetLibreGraphWeight())
```

Or even use `a.GetLibreGraphWeight() - b.GetLibreGraphWeight()` instead of calling `cmp.Compare`.

rhafer (changes requested):

services/graph/pkg/unifiedrole/roles.go
AlexAndBear: you rather mean: 


if descending {
    return cmp.Compare(b.GetLibreGraphWeight(), **a**.GetLibreGraphWeight())
}
return cmp.Compare(a.GetLibreGraphWeight(), b.GetLibreGraphWeight())

right ?

rhafer (approved):

kulmann:
@rhafer @dragotin what do you think, can be backport this to `stable-2.0`? The original behaviour was quite annoying, but it's not more than that... don't know how strict you are with backports.

Issue #716, opened 2025-04-23
Activitylog service creates endless load on the backend after uploading


cscholz:
For the past few days, I've been seeing similar entries in journalctl roughly every two seconds, like the following one:

`2025-04-23T14:21:09+02:00 INF user idp:"none" opaque_id:"deaf1971-9364-4f3e-805a-2893fbc2cb3f" type:USER_TYPE_SERVICE authenticated line=[github.com/opencloud-eu/reva/v2@v2.31.0/internal/grpc/services/authprovider/authprovider.go:146](http://github.com/opencloud-eu/reva/v2@v2.31.0/internal/grpc/services/authprovider/authprovider.go:146) pkg=rgrpc service=auth-service traceid=c78fcdfee027f644912ea20fe5f493f7`

For now I've not found an issue somewhere but unfortunately I'm also not able to interprete the log line record.

I would appreciate if someone can help me to understand what the log entries wanted to say.

dragotin:
renders the log output more or less useless

rhafer:
> For now I've not found an issue somewhere but unfortunately I'm also not able to interprete the log line record.

As mentioned in the  matrix chat. I think we should not log stuff like this at the "INFO" level. It should at least be down to "DEBUG" or "TRACE".

Regarding interpretation: The line basically just says that the service `auth-service` just successfully authenticated a service user (`type:USER_TYPE_SERVICE`). I guess you're more interested in finding out why that authentication happened and what the service user is actually doing there. For the you'd need to look at the surrounding log lines, maybe searching for lines having the same `traceid`.

dragotin:
In my case, these logs appeared after I copied some GB of data into my non collaborative PosixFS and started OpenCloud up again. From that point on there is one core constantly on 100% with opencloud. I wasn't able yet to pin down what it exactly does... The logfile is so flooded ;-)

dragotin:
Ah, yes, I let it run for a few days to give enough time (I hope) to assimilate all the files. But I have no clue if that has finished yet or not.

cscholz:
> In my case, these logs appeared after I copied some GB of data into my non collaborative PosixFS and started OpenCloud up again. From that point on there is one core constantly on 100% with opencloud. I wasn't able yet to pin down what it exactly does... The logfile is so flooded ;-)

I've copied also a bunch of data on filesystem level. ~ 200 GB. My host system was with 8 cores busy for hours.
[...]

PR #825, opened 2025-05-12
nats: Don't enable debug and trace logging by default

rhafer:
Only enable trace and/or debug logging when the outer logger has the log level set accordingly. Even when the though the messages where not really logged the nats service wasted quite some cycles especially in the trace log related code path.

Related issue: #716

For details see: https://github.com/opencloud-eu/opencloud/issues/716#issuecomment-2866835467


aduffeck (approved):
From my local tests it doesn't seem to make a difference performance- or load-wise but it still doesn't make sense to have those enabled. lgtm

PR #837, opened 2025-05-13
PosixTest. Check that version, share and link still exist

ScharfViktor:
tests for: 
- rename file on POSIX filesystem
- checking that file versions are still up to date after renaming or changing the contents of a file in a POSIX file system
- checking that share and public link still up to date after renaming or changing the contents of a file in a POSIX file system

### DIFF
--- a/tests/acceptance/bootstrap/CliContext.php
+++ b/tests/acceptance/bootstrap/CliContext.php
@@ -579,6 +579,30 @@ class CliContext implements Context {
 	}
 
 	/**
+	 * @When the administrator renames the file :file to :newName for user :user on the POSIX filesystem
+	 *
+	 * @param string $user
+	 * @param string $file
+	 * @param string $newName
+	 *
+	 * @return void
+	 */
+	public function theAdministratorRenamesFile(string $user, string $file, string $newName): void {
+		$userUuid = $this->featureContext->getUserIdByUserName($user);
+		$storagePath = $this->getUsersStoragePath();
+
+		$source = "$storagePath/$userUuid/$file";
+		$destination = "$storagePath/$userUuid/$newName";
+
+		$body = [
+		  "command" => "mv $source $destination",
+		  "raw" => true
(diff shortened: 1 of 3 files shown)


individual-it (commented):
so changing the content on the posix file system is not creating a new version?

v-scharf:
> so changing the content on the posix file system is not creating a new version?

Yes, it doesn't create a new version (checked with @aduffeck), but it leaves old versions - the test checks this out

rhafer:
@ScharfViktor It seems we have some "funny" dependency issue in our `.woodpecker.star`. This PR did not run any of the API / core API test. (It should, because it makes adjustments to those tests).

Looking at the generate woodpecker configs (https://ci.opencloud.eu/repos/3/pipeline/1631/config) you can see that e.g.
the `API-*-bla` test have this dependency:

```
depends_on:
  - build_opencloud_binary_for_testing
```

But `build_opencloud_binary_for_testing` has this:

```
when:
...
  - event: pull_request
    path:
      exclude:
        - .github/**
        - .vscode/**
        - docs/**
        - deployments/**
        - CHANGELOG.md
        - CONTRIBUTING.md
        - LICENSE
        - README.md
        - tests/acceptance/**
```

As the PR only changes stuff in `test/acceptance/` the `build_opencloud_binary_for_testing` workflow is not enable. So the dependency of the `API-*` workflows can never be fulfilled. I guess something is wrong with our skipIfUnchange() logic in .woodpecker. star

v-scharf:
> @ScharfViktor It seems we have some "funny" dependency issue in our `.woodpecker.star`. This PR did not run any of the API / core API test. (It should, because it makes adjustments to those tests).

Thank you for hint. Here is fix #856 I'll try to rebase this PR after merge it

v-scharf:
> Thank you for hint. Here is fix #856 I'll try to rebase this PR after merge it

now it's work

PR #842, opened 2025-05-14
Reduce load caused by the activitylog service

### LINKED ISSUE #716: Activitylog service creates endless load on the backend after uploading

aduffeck:
This PR makes a few backward compatible changes to the activitylog service to reduce the load it imposes onto the system:

1. Writes are buffered (for 10s by default), which gets rid of a lot of Unmarshal/Marshal loops to add events to the lists, esp. on higher leves in the tree. Unmarshalling and marshalling theses lists was the main culprit for the high CPU load.
2. Cache parent ids for resources. There's now a TTL cache remembering the parents of resources, which gets rid of a lot of stat requests.
3. The list of events are now kept as messagepack lists instead of json which is a lot faster to maintain.
4. We now use the parent id from the event if available, removing the need to stat the resource just for that.

Fixes #716


services/activitylog/pkg/service/service.go
```diff
@@ -200,7 +305,13 @@ func (a *ActivitylogService) AddSpaceActivity(spaceID *provider.StorageSpaceId,
 		return fmt.Errorf("could not parse space id: %w", err)
 	}
 	rid.OpaqueId = rid.GetSpaceId()
-	return a.storeActivity(storagespace.FormatResourceID(&rid), eventID, 0, timestamp)
+	return a.storeActivity(storagespace.FormatResourceID(&rid), []RawActivity{
```
rhafer, line 306: I am not entirely sure, but might it make sense to debounce on Space related events as well? if you share a space with a lot of people IIRC there is also one event per sharee, isn't there?

services/activitylog/pkg/service/service.go
```diff
@@ -80,7 +163,11 @@ func New(opts ...Option) (*ActivitylogService, error) {
 		valService:       o.ValueClient,
 		lock:             sync.RWMutex{},
 		registeredEvents: make(map[string]events.Unmarshaller),
+		tp:               o.TraceProvider,
+		tracer:           o.TraceProvider.Tracer("github.com/opencloud-eu/opencloud/services/activitylog/pkg/service"),
+		parentIdCache:    cache,
```
rhafer, line 168: Is this cache really worth it? Aren't all of the Stat() calls answered from the NATs cache in the storageproviders anyway?

services/activitylog/pkg/service/service.go
```diff
@@ -265,70 +376,113 @@ func (a *ActivitylogService) activities(rid *provider.ResourceId) ([]RawActivity
 	}
 
 	var activities []RawActivity
-	if err := json.Unmarshal(records[0].Value, &activities); err != nil {
-		return nil, fmt.Errorf("could not unmarshal activities: %w", err)
+	if err := msgpack.Unmarshal(records[0].Value, &activities); err != nil {
```
rhafer, line 377: Nit: Hm, somehow I think we should have a debug log statement here when falling back json. (Feel free to ignore me here)

services/activitylog/pkg/service/service.go
```diff
@@ -265,70 +376,113 @@ func (a *ActivitylogService) activities(rid *provider.ResourceId) ([]RawActivity
...
 	}
+	subspan.End()
 
-	var activities []RawActivity
+	_, subspan = a.tracer.Start(ctx, "Unmarshal")
+	var existingActivities []RawActivity
 	if len(records) > 0 {
-		if err := json.Unmarshal(records[0].Value, &activities); err != nil {
-			return err
+		if err := msgpack.Unmarshal(records[0].Value, &existingActivities); err != nil {
```
rhafer, line 467: Nit: Hm, somehow I think we should have a debug log statement here when falling back json. (Feel free to ignore me here)

services/activitylog/pkg/service/service.go
aduffeck: Cached stats aren't super expensive, yes. But on the other hand it's one stat per uploaded file and parent (except for the direct parent) up to the space root. So uploading 500 files in "a/b/c/" would trigger 1000 superfluous gateway->storageprovider->nats roundtrips. And the in-memory ttl cache is really lightweight and straightforward, so... I would vote for keeping it.

services/activitylog/pkg/service/service.go
aduffeck: I'll add it :)

services/activitylog/pkg/service/service.go
aduffeck: Sharing the space with a group is just one event. If a user really shares with multiple users at once it's gonna be a couple of writes, yes, but it's gonna be much less than uploading a large directory and the writes also do not trigger a cascade up the tree.

rhafer: Hm, true I guess we can ignore this for now.

services/activitylog/pkg/service/service.go
rhafer: Fine with me. I just a bit cautious when it comes to more caches. They tend to cause funny side-effects. But I in this special case the risk it pretty low.

.woodpecker.star
```diff
@@ -1886,6 +1886,7 @@ def opencloudServer(storage = "decomposed", accounts_hash_difficulty = 4, depend
         "OC_JWT_SECRET": "some-opencloud-jwt-secret",
         "EVENTHISTORY_STORE": "memory",
         "OC_TRANSLATION_PATH": "%s/tests/config/translations" % dirs["base"],
+        "OC_WRITE_BUFFER_DURATION": "0", # Disable write buffer so that test expectations are met in time
```
```suggestion
        "OC_WRITE_BUFFER_DURATION": "0",  # Disable write buffer so that test expectations are met in time
```

For what ever reason the `check-starlark` job whats you to insert that space there.

rhafer (changes requested):

rhafer (approved):
:+1: provided the test are succeeding

aduffeck:
Yeah, working on that.

kulmann:
> 3. The list of events are now kept as messagepack lists instead of json which is a lot faster to maintain.

Does this break existing activity events? 🤔

rhafer:
> > 3. The list of events are now kept as messagepack lists instead of json which is a lot faster to maintain.
> 
> Does this break existing activity events? 🤔

No. It does a fallback to JSON when it can't unmarshal as msgpack

aduffeck:
@rhafer CI is green now. Still looking good with the commits I added after your review?

rhafer (approved):

Issue #810, opened 2025-05-09
a video file copied via the web is smaller than the original and therefore corrupt


cscholz:
## Describe the bug
I've recognized that when I copy a sepcific video file via web portal the new file is much smaller than the original one and by this the new file is corrupt. Unfortantely I can not reproduce the issue with guarentee but it occured just now again and I was able to greb the logs. Lines: 10-12 & 23-25

## Steps to reproduce
for now I don't know

## Setup
I'm running opencloud 2.2.0 on docker (podman) on an flat ext4 filesystem.
<details>
<p>

## Below my docker-compose config.
```
version: "3.8"
# Collabora ########################################
services:
  collabora:
    cap_add:
      - MKNOD
    command:
      - coolconfig generate-proof-key && /start-collabora-online.sh
    entrypoint:
      - /bin/bash
      - -c
    environment:
      DONT_GEN_SSL_CERT: YES
      aliasgroup1: https://wopi.domain.tld:443
      extra_params: |
        --o:ssl.enable=false \
        --o:ssl.ssl_verification=false \
        --o:ssl.termination=true \
        --o:welcome.enable=false \
        --o:net.frame_ancestors=oc.domain.tld
        --o:welcome.enable=false
      password: PASSWORD
      username: admin
    healthcheck:
      test:
        - CMD
        - curl
        - -f
        - http://localhost:9980/hosting/discovery
    image: collabora/code:24.04.13.2.1
    ports:
      - 9980:9980
    network_mode: host
    labels:
      traefik.enable: "true"
      traefik.http.routers.collabora.entrypoints: https
      traefik.http.routers.collabora.rule: Host(`api.domain.tld`)
      traefik.http.routers.collabora.service: collabora
      traefik.http.routers.collabora.tls.certresolver: http
      traefik.http.services.collabora.loadbalancer.server.port: "9980"
    restart: always
  # Collaboration ##################################################
  collaboration:
    command:
      - -c
      - opencloud collaboration server
    depends_on:
      collabora:
        condition: service_healthy
      opencloud:
        condition: service_started
    entrypoint:
      - /bin/sh
    environment:
      COLLABORATION_APP_ADDR: https://api.domain.tld
      COLLABORATION_APP_ICON: https://api.domain.tld/favicon.ico
      COLLABORATION_APP_INSECURE: "true"
      COLLABORATION_APP_NAME: CollaboraOnline
      COLLABORATION_APP_PRODUCT: Collabora
[...]

PR #845, opened 2025-05-14
avoid pushing rolling and production images at the same time

ScharfViktor:
related: https://github.com/opencloud-eu/opencloud/issues/810#issuecomment-2876829859

when we create a tag in stable branch we build and push containers for [production](https://hub.docker.com/r/opencloudeu/opencloud)  and  [rolling](https://hub.docker.com/r/opencloudeu/opencloud-rolling) 

I deleted  build and push containers for rolling in the stable branch

Screenshot 2025-05-14 at 10 55 03

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -337,13 +337,17 @@ config = {
         "architectures": ["arm64", "amd64"],
         "production": {
             # NOTE: need to be updated if new production releases are determined
-            "tags": ["2.0", "3.0"],
+            "tags": ["2.0"],
             "repo": docker_repo_slug,
             "build_type": "production",
         },
         "rolling": {
             "repo": docker_repo_slug + "-rolling",
             "build_type": "rolling",
+        },
+        "daily": {
+            "repo": docker_repo_slug + "-rolling",
+            "build_type": "daily",
         },
     },
     "litmus": True,
@@ -1519,7 +1523,7 @@ def logTracingResults():
 def dockerReleases(ctx):
     pipelines = []
     docker_repos = []
-    build_type = "daily"
+    build_type = ""
 
     if ctx.build.event == "tag":
         tag = ctx.build.ref.replace("refs/tags/v", "").lower()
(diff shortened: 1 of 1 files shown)


micbar (changes requested):
Please add this as a switch to the top config dict. We should be able to use the same woodpecker config for main and stable branches.

v-scharf:
> Please add this as a switch to the top config dict. We should be able to use the same woodpecker config for main and stable branches.

Thanks for the review. It should now work in both main and stable. I will move this change to main after the merge.

micbar:
I need to do thorough testing, these changes have always been tricky 😏

rhafer (dismissed):
Let's do this change in `main` first. And that backport to `stable`

v-scharf:
> Let's do this change in `main` first. And that backport to `stable`

https://github.com/opencloud-eu/opencloud/pull/859

v-scharf:
> Let's do this change in `main` first. And that backport to `stable`

https://github.com/opencloud-eu/opencloud/pull/859 merged into the main branch and tested on the last rolling release.
this PR is port to stable. we need to have it for the next patch release

.woodpecker.star
```diff
@@ -337,14 +337,18 @@ config = {
...
-            "tags": ["2.0", "3.0"],
+            "tags": ["2.0"],
             "repo": docker_repo_slug,
             "build_type": "production",
         },
         "rolling": {
             "repo": docker_repo_slug + "-rolling",
             "build_type": "rolling",
         },
+        "daily": {
```
micbar, line 348: I do not understand this change. We are not building daily on the stable branch.

micbar (changes requested):

.woodpecker.star
v-scharf: no, we don't have daily in main. and It doesn't work in stable because` build and push` works only when we have tag: https://github.com/opencloud-eu/opencloud/pull/845/files#diff-0d2cb487e7e8bfe77cea5a2347a1987e5c8864b49ffdc1a75543dba849928a74R1645-R1648

> We should be able to use the same woodpecker config for main and stable branches.

but you suggested to have similar woodpecker config for main and stable branches. That's why I left all how in main

I would also delete `rolling` and `daily` from config for stable to don't confuse

micbar: okay. Let us prefer similarity of the files.

micbar (approved):

PR #847, opened 2025-05-14
adjust app-provider config.go so it can be rendered in docusaurus

Svanvith:
Need some changes so the md file can be rendered in docusaurus


micbar (changes requested):
Doesn't work.

dragonchaser:
no longer relevant, fixed in templates

PR #848, opened 2025-05-14
adjust frontend defaultconfig.go so it can be rendered in docusaurus

Svanvith:


services/frontend/pkg/config/defaults/defaultconfig.go
```diff
@@ -108,7 +108,7 @@ func DefaultConfig() *config.Config {
 		OCS: config.OCS{
 			Prefix:                      "ocs",
 			SharePrefix:                 "/Shares",
-			HomeNamespace:               "/users/{{.Id.OpaqueId}}",
+			HomeNamespace:               "/users/`{{.Id.OpaqueId}}`",
```
micbar, line 111: Hehe, that is good for docusaurus but would break our template engine.

micbar (changes requested):

dragonchaser:

PR #850, opened 2025-05-14
adjust notification config.go so it can be rendered by docusaurus

Svanvith:


micbar (changes requested):
Doesn't work

dragonchaser:

PR #852, opened 2025-05-14
adjust storage-user defaultconfig.go so it can be rendered by docusaurus

Svanvith:


dragonchaser:

PR #853, opened 2025-05-14
adjust webdav defaultconfig.go so it can be rendered by docusaurus

Svanvith:


micbar (changes requested):
That would break the templating system in the code.

dragonchaser:

PR #856, opened 2025-05-14
do not skip test on the build binary step

ScharfViktor:
When I change only the files in the `tests` folder. e.g. add some tests -> tests in CI do not run

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -2107,7 +2107,7 @@ def skipIfUnchanged(ctx, type):
         skip = base + acceptance
     elif type == "build-binary" or type == "build-docker" or type == "litmus":
         skip = base + unit + acceptance
-    elif type == "cache": or "base"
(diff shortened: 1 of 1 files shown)


.woodpecker.star
```diff
@@ -2107,7 +2107,7 @@ def skipIfUnchanged(ctx, type):
         skip = base + acceptance
     elif type == "build-binary" or type == "build-docker" or type == "litmus":
         skip = base + unit + acceptance
-    elif type == "cache":
+    elif type == "cache": or "base"
```
```suggestion
    elif type == "cache" or type == "base":
```

rhafer (changes requested):

.woodpecker.star
v-scharf: Sorry, I rushed

rhafer (approved):

PR #859, opened 2025-05-15
avoid pushing rolling and production images at the same time

ScharfViktor:
related #845 

Screenshot 2025-05-14 at 10 55 03


.woodpecker.star
```diff
@@ -344,6 +338,16 @@ config = {
     },
     "dockerReleases": {
         "architectures": ["arm64", "amd64"],
+        "production": {
+            # NOTE: need to be updated if new production releases are determined
+            "tags": ["2.0", "3.0"],
```
rhafer: Why do we know that "3.0" will be a production release?

.woodpecker.star
```diff
@@ -1476,24 +1480,27 @@ def dockerReleases(ctx):
...
-                docker_repos.append(docker_repo_slug)
+                is_production = True
                 break
 
+        # Tags starting with 2.0 or 3.0 indicate a production release → push to opencloudeu/opencloud
+        if is_production:
+            docker_repos.append(config["dockerReleases"]["production"]["repo"])
+            build_type = config["dockerReleases"]["production"]["build_type"]
+
+            # Tags starting with 2.x or 3.x indicate a rolling release → push to opencloudeu/opencloud-rolling
```
rhafer: Where does this assumption come from? I might be misunderstanding our versioning scheme but to my understanding the next production release might as well be a 2.3.0 or a 2.2.5. And similar 3.0.0 might be a rolling release.

@micbar Could you clarify please?

v-scharf: this logic already existed: 
we defined that it was a production release if it matched this tag PRODUCTION_RELEASE_TAGS = [“2.0”, “3.0”].

nothing has changed here.

my logic, if I understand it correctly: 
```
if tag -> define rolling or production
     if production: build_type=production, repo=opencloudeu/opencloud
     else:  build_type=rolling, repo=opencloudeu/opencloud-rolling
else: build_type=daily, repo=opencloudeu/opencloud-rolling
```

.woodpecker.star
```diff
@@ -1474,26 +1482,33 @@ def logTracingResults():
...
-                docker_repos.append(docker_repo_slug)
+                is_production = True
                 break
 
+        # Tags starting with 2.0 indicate a production release → push to opencloudeu/opencloud
+        if is_production:
+            docker_repos.append(config["dockerReleases"]["production"]["repo"])
+            build_type = config["dockerReleases"]["production"]["build_type"]
+
+            # Tags starting with 2.x indicate a rolling release → push to opencloudeu/opencloud-rolling
```
```suggestion
```

I think the comment is misleading here as well.

.woodpecker.star
```diff
@@ -1474,26 +1482,33 @@ def logTracingResults():
...
-        for prod_tag in PRODUCTION_RELEASE_TAGS:
+
+        is_production = False
+        for prod_tag in config["dockerReleases"]["production"]["tags"]:
             if tag.startswith(prod_tag):
-                docker_repos.append(docker_repo_slug)
+                is_production = True
                 break
 
+        # Tags starting with 2.0 indicate a production release → push to opencloudeu/opencloud
```
```suggestion
```

IMO this is still misleading. Let's remove it completely. Whether this is a rolling or a production release is only defined through the tags listed in `config["dockerReleases"]["production"]["tags"]`

rhafer (changes requested):

.woodpecker.star
v-scharf: deleted

.woodpecker.star
v-scharf: for now we know for sure that 2.0.x refers to a production release. So let's keep this value and remove the others

rhafer (approved):

PR #896, opened 2025-05-20
Add ADR for simple multi-tenancy

### LINKED ISSUE #879: Multi Tenancy Technical Concept
# Description

Technical concept for Multi Tenancy Version 1

## User Stories

* > As a cloud provider, i want to separate accounts on my cloud offering to minimize operation and maintenance cost

## Value

Generate Cloud Provider Opportunities

## Acceptance Criteria

- [ ] Written concept
  - [ ] Authentication
  - [ ] Reva
  - [ ] Sharing
- [ ] ADR format

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the OpenCloud charts is compatible

rhafer:
Closes #879

### DIFF
--- a/docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
+++ b/docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
@@ -24,13 +24,13 @@ To further limit the scope there are a couple of constraints:
 
 - Tenants are rather small (sometimes just a single user, often less than 10)
 - There is just a single IDP with a single "realm".
-- The user-management is external to OpenCloud. There is no tenant-specific adminstrator
-  that can add/delete users.
+- The user-management is external to OpenCloud
 - The membership of a user to a tenant is represented by a tenant id
   that is provided via a claim in the users' Access Token/UserInfo or by a per User
   Attribute in the LDAP directory.
 - There is no need to support per tenant groups
 - There is no need to isolate the storage of the tenants from each other
+- Role Assignment happens at first login via OIDC claims
 
 ## Decision Drivers
 
@@ -39,7 +39,6 @@ To further limit the scope there are a couple of constraints:
 * Implementation effort: The solution should be easy to implement and maintain.
 * Security: The solution should prevent users from seeing or accessing anything
   from other tenants.
-
 
 ## Considered Options
 
@@ -53,7 +52,7 @@ is extended by a new property "tenantId".
 * Everywhere the UserID is used the tenant id is also available.
   * This might allow implementing more sophisticated checks e.g. on permission
     grants and to a certain extend during share creation.
-  * the tenant id is immediately available e.g. in Events/Auditlog without an addtional
+  * the tenant id is immediately available e.g. in Events/Auditlog without an additional
     user lookup
 
 #### Cons:
@@ -63,7 +62,7 @@ is extended by a new property "tenantId".
   opaque identifier without carrying and specific semantics other than being
   globally unique.
 * on the GraphAPI the ID of a User is just a opaque string without any additional
-  meaning. (Currently it is just using the `OpqueId` property of the CS3 UserId,
+  meaning. (Currently it is just using the `OpaqueId` property of the CS3 UserId,
   without considering the `idp` property.)
 
 ### Option 2: Tenant ID is stored as the `idp` value of the CS3 UserId
@@ -134,31 +133,19 @@ switch to only allow users to search for users that are part of the same group.
 
 ## Decision Outcome
 
+### Chosen option: Option 1 - Tenant ID as a new property of the CS3 UserId
+
 As part of the OIDC Connect Authentication OpenCloud will receive a tenant id via the
-Access Token or Userinfo endpoint. When no shared LDAP server is available
-OpenCloud will store the tenant id as part of the autoprovionsing process.
-Either as via a per-tenant subtree (e.g. 'ou=tenantid,dc=example,dc=com') or as
-an addtional user attribute. When a shared LDAP server is available the tenant
-id must be available there either as a separate attribute of the user or as
-part of the DN (i.e. separate per tenant subtrees).
+Access Token or Userinfo endpoint. For the initial implementation it is assume that OpenCloud
+has read access to a shared LDAP server, that contains all users of all tenants. Users in
+LDAP will have a the tenant id as an dedicated attribute.
 
-More details TBD. Depending on which option we choose from the above list.
+### Implementation Steps
 
-~~The CS3 API needs to be extended to support the tenant id. This will require adding a new
-(optional if possible) field to the User Object (or the UserId inside the User Object). When requesting
-users reva needs to assure that only user objects of the same tenant as the requesting user are
-returned.~~
-
-When creating a share there is a check in place that verifies that the sharer
-and all sharees are part of the same tenant.
-
-
-### Positive Consequences:
-
-* All opencloud service are shared by all tenants, no per-tenant service required
-* little changes to the internal APIs
-
-### Negative consequences:
-
-* no clear data separation
+* Extend the CS3 UserId with a new property "tenantId"
+* Adapt the `users` service` to only return users that are part of the same tenant
+* To cleanup technical debt and reduce code duplication the `cs3` users backend in the
+  graph service is being improved so that is can fully replace the `LDAP` users backend for
+  read operations.
+* The reva `share-provider` service only allow shares between users of the same tenant
 


docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,64 @@
...
+As part of the OIDC Connect Authentication OpenCloud will receive a tenant id via the
+Access Token or Userinfo endpoint. When no shared LDAP server is available
+OpenCloud will store the tenant id as part of the autoprovionsing process.
+Either as via a per-tenant subtree (e.g. 'ou=tenantid,dc=example,dc=com') or as
+an addtional user attribute. When a shared LDAP server is available the tenant
+id must be available there either as a separate attribute of the user or as
+part of the DN (i.e. separate per tenant subtrees).
+
+The CS3 API needs to be extended to support the tenant id. This will require adding a new
+(optional if possible) field to the User Object (or the UserId inside the User Object). When requesting
```
rhafer: @butonic During planing you mentioned that the tenant-id MUST be part of the CS3 user id. What is  the reason for that? The UserId itself is already unique across the whole OpenCloud instance. Can't we just make it part of the user attributes?

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Context and Problem Statement
+
+To reduce resource usage and cost service providers want a single OpenCloud
+instance to host multiple tenants. Members of the same tenant should be able to
+only see each other when trying to share resources. A user can only be a member
+of a single tenant. Moving a user from one tenant to another is not supported.
+For the scope of this ADR we assume the tenants are overall rather small
+(sometimes just a single user, often less than 10). Also there is no need to
+support per tenant groups of users for now.
```
butonic: I don't know if disabling group management is currently just a config flag, or if disabling this requires more work than adding tenants to groups.

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Decision Drivers
+
+* Low Resource Overhead: The solution should not require a lot of additional resources
+  (CPU, Memory, Storage) on the OpenCloud instance.
+* Implementation effort: The solution should be easy to implement and maintain.
```
butonic, line 39:
```suggestion
* Implementation effort: The solution should be easy to implement and maintain.
* Security: The solution should prevent users from seeing or accessing anything from other tenants.
```

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Considered Options
+
+* A tenant id is assigned to every user. The tenant id is used to filter the
+  resources that are shown to the user. The tenant id will be a new part of the
+  CS3 UserId. The user <-> tenant id relation in maintained by the OIDC IDP and
+  the tenant id will be available as a claim in the OIDC tokens.
```
butonic: Hm, do we want to rely on a claim issued by the IDP, or do we want to map the issuer to an internal tenant id, similar to how we should be doing it for the sub to internal user id mapping? I prefer the latter.

I'm scared that when we allow multiple idps someone will read the tenant from the claims and then someone can fake the tenant, or there might be collisions. IMO the tenant should be derived from the issuer. Does not every tenant also require its own IdP (well, realm in keycloak).

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
butonic: The permissions assembly in decomposedfs currently only uses the userid. When checking grants we should imo als check the tenant id to ensure that resources can never be accessed by a user from a different tenant.

When we want to support multiple idps we either MUST always roll our own uuid for the users or we MUST compare the combination of iss+sub which would be a tenant id and a userid.

So yeah, tenant might be ok to have in the user struct, but I really prefer to have it in the userid. Otherwise, we cannot really tell if a grant of a user belongs to the same tenant or not. 

...

🤔 

Ok. We can keep the tenant id out of the userid. The permissions assembly should also work in a world in which we have a federated sea of spaces. I'm happy to keep the tenant out of that. For that we need to disconnect the tenant/issuer and user id when authenticating the access token. But do we then even need the tenant id in the CS3 user message? Having it in the LDAP attributes should be sufficient, right?

rhafer: > The permissions assembly in decomposedfs currently only uses the userid. When checking grants we should imo als check the tenant id to ensure that resources can never be accessed by a user from a different tenant.

If we prevent grants from being added for users in a different tenant, resources shouldn't be accessible from another tenant. Also getting the tenant id of the owner is just a single user lookup away.

I am somewhat reluctant to add yet another attribute to the `userid`, because currently we're already not using the existing attributes `userid` + `idp` really consistently. Adding the tenantid would make things worse IMO. Ideally the `userid` would just be a single valued unique id anyway (I guess we agree on that), but that is a different story.

> When we want to support multiple idps we either MUST always roll our own uuid for the users or we MUST compare the combination of iss+sub which would be a tenant id and a userid.

Support for multiple IDPs is not in the scope of this ADR. The scope for now is really to provide "just enough multitenancy" so that users from different tenants can be hosted on the same opencloud instances without seeing each others' resources (spaces, files, shares, users). I guess I need to make the clearer in the Problem statement.

> So yeah, tenant might be ok to have in the user struct, but I really prefer to have it in the userid. Otherwise, we cannot really tell if a grant of a user belongs to the same tenant or not.
> 
> ...
> 
> 🤔
> 
> Ok. We can keep the tenant id out of the userid. The permissions assembly should also work in a world in which we have a federated sea of spaces. I'm happy to keep the tenant out of that. For that we need to disconnect the tenant/issuer and user id when authenticating the access token.

Side note: For the initial use case the `issuer`, if you are referring to the `iss` claim with that will always be the same for all users in all tenants. There is not going to be a realm per tenant. The tenant of a user will be a separate claim.

> But do we then even need the tenant id in the CS3 user message? Having it in the LDAP attributes should be sufficient, right?

I guess it could work like that, yes. I am still pondering whether that is a good idea. We'll have the tenant of the current user in the access token and we'll also need to have it in the reva token. The user's service could use that to construct the proper LDAP filters to only get users assigned to the current tenant.

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
rhafer: Since all users are centrally managed. There should not be any user that has the privileges to edit users and groups. Apart maybe from assigning roles.

Actually there are a few details we still need to sort out for that. How are roles being assigned? Since the IDP is centrally managed it might not be possible to configure the roles via OIDC claims. That means we would at least need to have one user per tenant that is able change the roles of other users in the tenant. E.g. to have SpaceAdmins.

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
rhafer: No. For this specific use case the is just a single IDP with a single realm AFAIK. So we need to rely on the claim.
@micbar please holler if I am wrong here.

We could still map the claim value to some internally rolled UUID of course.

micbar: I understood it the same way. Single IdP. TenantID as OIDC claim.

rhafer:
I've pushed a new iteration of the ADR. Addressing some of the feedback and adding a few options (mainly the one that shouldn't require changes to the CS3 Objects).

I don't think we're ready to pick an option yet. Let's discuss the options is some more detail first.

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,83 @@
...
+## Context and Problem Statement
+
+To reduce resource usage and cost service providers want a single OpenCloud
+instance to host multiple tenants. Members of the same tenant should be able to
+only see each other when trying to share resources. A user can only be a member
+of a single tenant. Moving a user from one tenant to another is not supported.
+For the scope of this ADR we assume the tenants are overall rather small
+(sometimes just a single user, often less than 10). There is just a single IDP
+with a single "realm". Also there is no need to support per tenant groups of
+users for now.
```
butonic: At least for keycloak ... how can that be managed? or are users provided via ldap and the tenant is a property of the user there? How do opencloud admins add users to their tenant? Through the opencloud web ui? and we provision them in LDAP with the same tenant the admin is a member of? That might work ...

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,83 @@
...
+
+* Low Resource Overhead: The solution should not require a lot of additional
+  resources (CPU, Memory, Storage) on the OpenCloud instance.
+* Implementation effort: The solution should be easy to implement and maintain.
+* Security: The solution should prevent users from seeing or accessing anything
+  from other tenants.
+
+## Considered Options
+
+The user <-> tenant id relation in maintained by the OIDC IDP and the tenant id
```
butonic:
```suggestion
The user <-> tenant id relation is maintained by the OIDC IDP and the tenant id
```

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
```diff
@@ -0,0 +1,83 @@
...
+(sometimes just a single user, often less than 10). There is just a single IDP
+with a single "realm". Also there is no need to support per tenant groups of
+users for now.
+
+## Decision Drivers
+
+* Low Resource Overhead: The solution should not require a lot of additional
+  resources (CPU, Memory, Storage) on the OpenCloud instance.
+* Implementation effort: The solution should be easy to implement and maintain.
+* Security: The solution should prevent users from seeing or accessing anything
```
dragotin, line 40: Add as decision driver that the solution should make easy horizontal scaling possible?

butonic:
> I anticipate admins want to have the tenants also separated on the storage

we might be able to use the users issuer property in the path template ... I don't know of the template magic allows us to ... use the first group of a user? `tenants/{{ index .User.Groups 0 }}/users/{{.User.Username}}` might work, but then every user requires to have a group.

Hm, misusing groups seems to be equally bad as misusing the user id idp property. When a hoster decides they want to have groups for their tenants we could provide that by misusing the user id idp property and then filter groups based on it.

Also especially for full blown multi tenancy the idp property will be different and we could even support multiple tenants on the same openldap. I think we should try to have only one way to do multi tenancy.

To be honest, after taking the discussions into consideration, I would prefer to have a TenantID on the CS3 UserID. That would clarify what it should be used for. We can always just populate it from a `tenant` claim if it was present. And we can then always add the tenant to the ldap filter when listing users (and groups, I see no reason why that should not be compatible) . No new configuration options necessary. Multi tenancy would just automatically be jailed into a tenant if the IdP sends that claim.

And since the seperation by tenant migh be critical in some services ... eg. search I think I having it only in the User object is not sufficient. If not now, then at some point in the future. Maybe when we have to pass a userid via events ...

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
micbar: There is currently no requirement for an "OpenCloud admin".

rhafer: The "how is that managed" question is not in the scope of this ADR. We assume:

- The user-management is external to OpenCloud. The user management UI of OpenCloud web, is not used at all. (Well maybe for role-assignments, still tbd)
- All the OpenCloud is supposed to be aware of (either by LDAP Attributes or OIDCs claims or both ), is that users belong to a tenant and users from different tenants are not able to see each other and each other's data.

rhafer:
> This ADR does not touch the storage layout in any way. Shouldn't we change that?

That was on purpose. For the scope of this ADR ("__simple__ multi-tenancy")  we don't need to have the storage separated per tenant. Unfortunately that aspect was up to now only communicated verbally, at least I can't find it in any ticket). I'll add to the "Context Problem Statement Section".

docs/adr/0001-simple-multi-tenancy-using-a-single-opencloud-instance.md
rhafer: I guess that is what I was referring to with "Low Resource overhead"

PR #901, opened 2025-05-20
graph: move to reva's auto-reconnect LDAP client

rhafer:
This reduces code duplication


micbar (approved):
Always happy to reduce duplication 😄

PR #941, opened 2025-05-27
Make activitylog service scalable

aduffeck:
This PR changes the data format of the data stored by the activitylog service in a way that removes the race condition when running multiple instances of the service.

In order to achieve this we switched from a go-micro store to a native nats implementation which stores chunks of activities per resource instead of maintaining a large single list.


services/activitylog/pkg/service/migrations.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// If version is something else, it might indicate a future version or an unexpected state.
+	// Add logic here if more complex version handling is needed.
+	return fmt.Errorf("unexpected activitylog version: %s, expected %s or older", version, currentMigrationVersion)
+}
+
+// migrateToV1 performs the data migration to version 1.
+// It iterates over all keys, expecting their values to be JSON arrays of strings.
+// For each such key, it creates a new key in the format "originalKey:count:timestamp"
+// and stores the original list of strings (re-marshalled to JSON) as its value.
```
rhafer: Nit: Isn't it re-marshalled to MsgPack?

services/activitylog/pkg/service/migrations.go
```diff
@@ -0,0 +1,128 @@
...
+		// Construct the new key
+		newKey := natsKey(val.Key, len(activities))
+		newValue, err := msgpack.Marshal(activities)
+		if err != nil {
+			a.log.Error().Err(err).Str("key", key).Msg("migrateToV1: Failed to marshal activities. Skipping.")
+			skippedCount++
+			continue
+		}
+
+		// Write the value (the list of strings, marshalled as JSON) under the new key
```
rhafer: here the comment seems to be off as well. `newValue` is msgpack not JSON

services/activitylog/pkg/service/service.go
```diff
@@ -166,8 +206,16 @@ func New(opts ...Option) (*ActivitylogService, error) {
 		tp:               o.TraceProvider,
 		tracer:           o.TraceProvider.Tracer("github.com/opencloud-eu/opencloud/services/activitylog/pkg/service"),
 		parentIdCache:    cache,
+		maxActivities:    o.Config.MaxActivities,
+		natskv:           kv,
+	}
+	s.debouncer = NewDebouncer(o.Config.WriteBufferDuration, s.storeActivity)
+
+	// run migrations
+	err = s.runMigrations(context.Background(), kv)
```
rhafer, line 215: Don't we need to assure that migrations are only run by a single instance? Otherwise they'll step on each other's toes, won't they?

services/activitylog/pkg/service/service.go
```diff
@@ -358,45 +404,52 @@ func (a *ActivitylogService) RemoveResource(rid *provider.ResourceId) error {
...
 
-	records, err := a.store.Read(resourceID)
-	if err != nil && err != microstore.ErrNotFound {
-		return nil, fmt.Errorf("could not read activities: %w", err)
-	}
+	glob := fmt.Sprintf("%s.>", base32.StdEncoding.EncodeToString([]byte(resourceID)))
 
-	if len(records) == 0 {
-		return []RawActivity{}, nil
+	watcher, err := a.natskv.Watch(glob, nats.IgnoreDeletes())
```
rhafer, line 415: Aren't you supposed to call `Stop()` on a watch at some point, when done?

rhafer (changes requested):
Just some minor nits.

services/activitylog/pkg/service/service.go
aduffeck: The documentations isn't really clear about this but it sounds like the right thing to do. Nice catch.

rhafer (approved):

PR #916, opened 2025-05-22
graph: Add more $select options to ListPermissions endpoint

rhafer:
This add the possibility to `$select` on `@libre.graph.permissions.actions.allowedValues`.

The request:
`GET /drives/<driveid>/root/permissions?$select=@libre.graph.permissions.actions.allowedValues`

can be used to figure out what permissions the requesting user has on a specific space (e.g. when deciding which action to display in the context menu)

Related Issue #485

PR #942, opened 2025-05-27
apiTest. select option for root/permissions endpoint

ScharfViktor:
related #916

select filter for: 
- `drives/$spaceId/root/permissions`
- `drives/$spaceId/items/$itemId/permissions`
endpoints

### DIFF
--- a/tests/acceptance/TestHelpers/GraphHelper.php
+++ b/tests/acceptance/TestHelpers/GraphHelper.php
@@ -2168,6 +2168,7 @@ class GraphHelper {
 	 * @param string $user
 	 * @param string $password
 	 * @param string $spaceId
+	 * @param string|null $query
 	 *
 	 * @return ResponseInterface
 	 * @throws GuzzleException
@@ -2177,9 +2178,14 @@ class GraphHelper {
 		string $xRequestId,
 		string $user,
 		string $password,
-		string $spaceId
+		string $spaceId,
+		?string $query = null
 	): ResponseInterface {
 		$url = self::getBetaFullUrl($baseUrl, "drives/$spaceId/root/permissions");
+		if ($query !== null) {
+			$url .= "?$query";
+		}
+
 		return HttpRequestHelper::get(
 			$url,
 			$xRequestId,
--- a/tests/acceptance/bootstrap/SharingNgContext.php
+++ b/tests/acceptance/bootstrap/SharingNgContext.php
@@ -182,6 +182,37 @@ class SharingNgContext implements Context {
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


v-scharf:
@individual-it I added json data response checking. Please have look it again

tests/acceptance/bootstrap/SharingNgContext.php
```diff
@@ -1878,16 +1886,7 @@ public function userTriesToListThePermissionsOfSpaceOwnedByUsingRootEndpointOfTh
...
-
-		$response = GraphHelper::getDrivePermissionsList(
-			$this->featureContext->getBaseUrl(),
-			$this->featureContext->getStepLineRef(),
-			$user,
-			$this->featureContext->getPasswordForUser($user),
-			$spaceId
-		);
-		$this->featureContext->setResponse($response);
+		$this->featureContext->setResponse($this->getDrivePermissionsList($user, $space));
```
individual-it: We also need to pass `$spaceOwner` to `getDrivePermissionsList()` because in this step the space owner is not the user.
That is also the reason why CI is failing

tests/acceptance/bootstrap/SharingNgContext.php
```diff
@@ -181,6 +181,31 @@ public function getPermissionsList(
...
 
+	/**
+	 * @param string $user
+	 * @param string $space
+	 * @param string|null $query
+	 *
+	 * @return ResponseInterface
+	 * @throws GuzzleException
+	 */
+	public function getDrivePermissionsList(
```
individual-it:
```suggestion
	private function getDrivePermissionsList(
```

tests/acceptance/TestHelpers/GraphHelper.php
```diff
@@ -2177,9 +2178,14 @@ public static function getDrivePermissionsList(
 		string $xRequestId,
 		string $user,
 		string $password,
-		string $spaceId
+		string $spaceId,
+		?string $query = null
 	): ResponseInterface {
 		$url = self::getBetaFullUrl($baseUrl, "drives/$spaceId/root/permissions");
+		if (!empty($query)) {
```
individual-it:
```suggestion
		if ($query !== null) {
```

`empty` returns `true` also for `0` `false`

individual-it (changes requested):

tests/acceptance/bootstrap/SharingNgContext.php
v-scharf: yeah, true. Fixed 👍

PR #946, opened 2025-05-27
fix(graph-metadata): lazy cs3 metadata storage initialization

fschade:
## Description
The recently introduced profile photo graph api (https://github.com/opencloud-eu/opencloud/pull/864) failed to initialize every now and then if the storage is not ready yet as pointed out in https://github.com/opencloud-eu/opencloud/issues/944#issuecomment-2911996712. 

Beside that i also made a wrong design decision which made testing unnecessary hard, both is fixed in this PR.

## Related Issue
- Extends: https://github.com/opencloud-eu/opencloud/pull/864
- Fixes: https://github.com/opencloud-eu/opencloud/issues/944#issuecomment-2911996712

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

PR #864, opened 2025-05-15
Add profile photo

fschade:
## Description
Implements the functionality described in [opencloud-eu/opencloud#732](https://github.com/opencloud-eu/opencloud/issues/732), which focuses on the ability to add custom profile pictures.

** MS-Graph API **
- https://learn.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-1.0&tabs=http
- https://learn.microsoft.com/en-us/graph/api/profilephoto-update?view=graph-rest-1.0&tabs=http
- https://learn.microsoft.com/en-us/graph/api/profilephoto-delete?view=graph-rest-1.0&tabs=http

## Related Issue
- Fixes #732

## Motivation and Context
allow users to upload and manage their profile pictures

## How Has This Been Tested?
- unit tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Todo:
- [X] Implementation
- [ ] Add Tests
- [ ] Add swagger spec (https://github.com/opencloud-eu/libre-graph-api)
- [ ] Onboard UI Team (@AlexAndBear FYI)

PR #956, opened 2025-05-30
enhancement: introduce nested metadata-storage and use it in the user…

fschade:
… profile photos graph api

## Description
the graph profile photo api introduced a dedicated space to store the users profile pictures, this is fine but weve discussed if it is efficient to have a dedicated space just for the images.

This pr introduces a deep metadata-storage which is allows the management of nested data structures, not sure if we want that or not, we should consider it as POC until we agree whether we want it that way or not!

If we decide thats the way to go i will add unit test!

@rhafer @butonic @dragonchaser @micbar what do you think... nonsense or useful?

## Related Issue

- https://github.com/opencloud-eu/opencloud/pull/946
- https://github.com/opencloud-eu/opencloud/pull/864

## Motivation and Context
have a bucket that can be used to store user related data.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


butonic:
AFAICT this wrapper just jails any file operations into a subfolder of the given space, right? Isn't that just syntactic sugar that ensures all operations prepend the same path prefix? Why is a simple
```go
func jail path(path string) {
  return path.Join("photos", path)
}
```
not sufficient? I am at a conference and haven't had the time to fully digest this PR. What am I missing?

fschade:
> AFAICT this wrapper just jails any file operations into a subfolder of the given space, right? Isn't that just syntactic sugar that ensures all operations prepend the same path prefix? Why is a simple
> 
> ```go
> func jail path(path string) {
>   return path.Join("photos", path)
> }
> ```
> 
> not sufficient? I am at a conference and haven't had the time to fully digest this PR. What am I missing?

the benefit in my opinion is that the deep storage can be stacked with other storages like https://github.com/opencloud-eu/opencloud/blob/e4bc971d02deeea08506f775a3f4be365f74f16b/pkg/storage/metadata/lazy.go.

But the main reason to have it is that it allows to create nested uploads without any hassle, e.g.

```go
root, err := revaMetadata.NewCS3Storage("", "", "", "", "")
if err != nil {
  log.Fatal(err)
}

if err := root.Init(context.Background(), "ROOT"); err != nil {
  log.Fatal(err)
}

// OK
_ = root.SimpleUpload(context.Background(), "test.txt", []byte("Hello, World!"))

// FAILS
_ = root.SimpleUpload(context.Background(), "child/test.txt", []byte("Hello, World!"))

nested, err := metadata.NewDeepStorage(root)
if err != nil {
  log.Fatal(err)
}

// OK
_ = nested.SimpleUpload(context.Background(), "child/test.txt", []byte("Hello, World!"))
```

butonic:
Ok, got it. Please correct me if I am wrong: you are implicity calling MkdirIfNotExists for every upload. So ... the DeepStorage ensures that files can always be written by creating the path if it does not exist, yet. At the cost of an additional MKCOL request per path segment in the path. 

1. It really depends on the use case if this implicit creation creates a lot of overhead. If files are written more than once a better approach might be to try uploading first and only create the path if it does not exist.
2. The web and desktop clients actually would like to implicitly create directories when uploading files as well. I would prefer solving this at the protocol level first. Unfortunately, the https://datatracker.ietf.org/doc/html/rfc4918#section-9.7.1 dictates that we MUST return a 409. The [original mailing list entry](https://lists.w3.org/Archives/Public/w3c-dist-auth/1999AprJun/0062.html) explains the problem well. IIRC there were ideas about an HTTP header, but I cannot find anything. We could also make this part of tus upload requests only?

I'm not against this. My only concern is that we may be creating a PROPFIND overhead if this is used for other use cases.

butonic (approved):

PR #976, opened 2025-06-02
adjust links in the readme

Svanvith:
fix links in the read me files


dragonchaser:
We discussed that this needs changes to the markdown generator. The output structure there has to mach the folder structure on github so the links work on both sides.

Issue #616, opened 2025-04-08
[Bug] Collaboration container is not reregistering when opencloud container restarts


suse-coder:
## Describe the bug
When only the opencloud container restarts (due to update), then the connection of the the collaboration container is lost, but it is not reregistering it (only after manual restart of the collaboration container). That leads that no collabora is available.

{"level":"error","service":"collaboration","error":"Failed to store data in bucket 'ONSXE5TJMNSS24TFM5UXG5DSPFPWK5JON5YGK3TDNRXXKZBOMFYGSLTDN5WGYYLCN5ZGC5DJN5XC4Q3PNRWGCYTPOJQU63TMNFXGKQDEGI2TMZLGHFTC2YLEMZTC2NBSGQZS2YRTGRTC2M3DMMZTEOBSGU2DMZLFIAZC4MJOGA======': nats: timeout"

```
Log of collaboration container:
{"level":"info","service":"collaboration","time":"2025-04-08T08:53:39Z","line":"github.com/opencloud-eu/opencloud/pkg/registry/register.go:19","message":"registering external service eu.opencloud.api.collaboration.CollaboraOnline-d256ef9f-adff-4243-b34f-3cc3282546ee@10.244.3.170:9301"}
{"level":"info","service":"collaboration","proto":"HTTP/1.1","request-id":"opencloud-collaboration-5b755959b4-rssc6/Z6wxIAWsp7-000002","traceid":"00000000000000000000000000000000","remote-addr":"10.244.3.170:37994","method":"GET","wopi-action":"","status":404,"path":"/","duration":0.048316,"bytes":19,"time":"2025-04-08T08:53:40Z","line":"github.com/opencloud-eu/opencloud/services/collaboration/pkg/middleware/accesslog.go:35","message":"access-log"}
{"level":"info","service":"collaboration","proto":"HTTP/1.1","request-id":"ce95e704-027e-46f0-a9ff-79894abcdb68","traceid":"00000000000000000000000000000000","remote-addr":"10.244.3.111:37763","method":"GET","wopi-action":"","status":200,"path":"/wopi/files/b012d6835a507b3c245d893f81ba6c65c954ac35e20d5201bc80b4c1cdd03db3","duration":18.790076,"bytes":686,"time":"2025-04-08T08:54:49Z","line":"github.com/opencloud-eu/opencloud/services/collaboration/pkg/middleware/accesslog.go:35","message":"access-log"}
[...]

PR #1035, opened 2025-06-10
fix(collaboration): re register app providers in a configurable interval

fschade:
## Description
App providers only register on startup of the collaboration service, this can lead to information loss if open-cloud (app-provider) goes down. The solution is fairly simple, i would LOVE to see an implementation which uses a registry instead... 

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/616

## How Has This Been Tested?
- docker deployment

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/collaboration/pkg/command/server.go
+++ b/services/collaboration/pkg/command/server.go
@@ -4,10 +4,14 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/oklog/run"
+	"github.com/urfave/cli/v2"
+	microstore "go-micro.dev/v4/store"
+
 	"github.com/opencloud-eu/opencloud/pkg/config/configlog"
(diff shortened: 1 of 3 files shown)


micbar:
@fschade thanks!

How does this work when you have more than one replica of the app provider service?

fschade:
> @fschade thanks!
> 
> How does this work when you have more than one replica of the app provider service?

Unfortunately, not at all. That was already the case before. We're just fixing the existing case here. I'd like to take some time and build a KV app manager... it's not really that complicated! Unfortunately, the existing micro manager doesn't work either, because it relies on metadata that's strictly tied to a single service.

micbar (approved):
Let us see if the 30 sec default is feasible. :+1:

PR #1086, opened 2025-06-22
fix: collabora use metrics instead of imperial metric system

AlexAndBear:
Coming from Matrix chat:

```
Hey everyone, I noticed that Collabora CODE gets it's UI language from the Opencloud user's language setting. However, there is only one English there which defaults to US. As a result, CODE shows imperial metrics (e.g. for page margin). Is there a way to switch to UK or Australian English so that I can use CODE with an English UI but with metric units?
```

## Description
Before **(Look to the right 👀 )**
image

After
image

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/collaboration/pkg/service/grpc/v0/service.go
+++ b/services/collaboration/pkg/service/grpc/v0/service.go
@@ -258,7 +258,7 @@ func (s *Service) addQueryToURL(baseURL string, req *appproviderv1beta1.OpenInAp
 	case "de":
 		lang = "de-DE"
 	case "en":
-		lang = "en-US"
+		lang = "en-GB"
 	case "es":
 		lang = "es-ES"
 	case "fr":
--- a/services/collaboration/pkg/service/grpc/v0/service_test.go
+++ b/services/collaboration/pkg/service/grpc/v0/service_test.go
@@ -244,7 +244,7 @@ var _ = Describe("Discovery", func() {
 			Expect(err).To(Succeed())
 			Expect(resp.GetStatus().GetCode()).To(Equal(rpcv1beta1.Code_CODE_OK))
 			Expect(resp.GetAppUrl().GetMethod()).To(Equal("POST"))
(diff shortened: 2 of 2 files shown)


micbar:
Hm. This feels like a hack.

AlexAndBear:
> Hm. This feels like a hack.

The whole code section is a hack. 
Please make a proposal then.

micbar:
We only need this for office online on prem. Maybe a good time to skip this for other products.

AlexAndBear:
What ?

AlexAndBear:
> We only need this for office online on prem. Maybe a good time to skip this for other products.

If you are referring to the lang code -> lang + region code conversion, this is not quite true. In order to show metric unit, Collabora is depening on the region code.
Otherwise it renders imperial metrics again. 

image

kulmann:
Also it feels quite natural for me, that we opt for `en-GB` instead of `en-US` if a region code is required somewhere in the platform. This is a European software stack...

AlexAndBear:
> Also it feels quite natural for me, that we opt for `en-GB` instead of `en-US` if a region code is required somewhere in the platform. This is a European software stack...

You are my saviour 🫶

micbar:
I think there might be a confusion here.

1.) Collabora uses language and region code.
2.) OpenCloud doesn't support region code, which has the issue that the unit systems for user facing numbers are not adapted to the region
3.) This conversion in the code is "just a hack" which should go away.

micbar:
Making decisions for users in our software stack which should be user decisions feels just wrong.

AlexAndBear:
Please add a proposal

micbar:
Proposal:

Add region suffixes to our codebase.

AlexAndBear:
@micbar this was discussed multiple times already and got denied.

I would also love to have this.

fschade:
thanks for ur PR, in my opinion merge it, but as @micbar proposed we should have a ticket to have support for regions, @tbsbdr FYI

Issue #1094, opened 2025-06-23
Mardown Image Link


DennisFaucher:
## Describe the bug
What is the correct format for inserting an image link into a markdown file in OpenCloud?

## Steps to reproduce
I have tried links to local images and Internet images and none of them display the image

I would rather use a link than upload an image as uploading inserts an extremely long string into my markdown file

## Actual behavior
A clear and concise description of what happened.

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OC_XXX=somevalue
OC_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

AlexAndBear:
Local images aka from your opencloud is not yet possible, as they aren't public accessible.

From the internet you can use this dialog:
*** you might need to update your csp file***
```
e.G. Content-Security-Policy: img-src * data: blob:;

```

<img width="1080" alt="Image" src="https://github.com/user-attachments/assets/920eb96f-f098-4292-a8d6-39980aa0bd97" />

<img width="1444" alt="Image" src="https://github.com/user-attachments/assets/69261305-536f-4b31-8828-1bf554950113" />

<img width="2010" alt="Image" src="https://github.com/user-attachments/assets/d5117bb1-078f-4946-9d17-e4b4e8b83851" />

AlexAndBear:
I updated the csp.yaml, so in order to display images in the markdown editor, admins don't need to adjust their csp file 
https://github.com/opencloud-eu/opencloud/pull/1101

DennisFaucher:
Thank you. I read the #1101 thread. 

What would be the process for inserting an image into markdown that is already on the OpenCloud server?

DennisFaucher:
This did not work for me. I am using the docker-compose.yml in ./deployments/examples/opencloud_full/docker-compose.yml

<img width="1212" alt="Image" src="https://github.com/user-attachments/assets/9b0d6210-2756-4918-a8fe-c96e05a9fb18" />

AlexAndBear:
after you adjusted the csp.yml you need to stop the container and restart it (docker compose down && docker compose up -d)

DennisFaucher:
Can you explain where that new directive goes in csp.yaml?
e.G. Content-Security-Policy: img-src * data: blob:;


```
directives:
  child-src:
    - '''self'''
  connect-src:
    - '''self'''
    - 'blob:'
[...]

PR #1101, opened 2025-06-24
feat: csp allow img src from everywhwere

AlexAndBear:
## Description
This PR updates the Content Security Policy (img-src) directive to allow all image sources (*). This change is necessary because we need to support embedding external images in the markdown editor. Markdown content often includes images from arbitrary URLs, and a restrictive img-src policy would block them from rendering properly.

## Related Issue

- Relates to https://github.com/opencloud-eu/opencloud/issues/1094

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/proxy/pkg/config/csp.yaml
+++ b/services/proxy/pkg/config/csp.yaml
@@ -16,10 +16,7 @@ directives:
     - 'blob:'
     - 'https://embed.diagrams.net/'
   img-src:
-    - '''self'''
-    - 'data:'
-    - 'blob:'
-    - 'https://raw.githubusercontent.com/opencloud-eu/awesome/'
+  - '''*'''
   manifest-src:
     - '''self'''
   media-src:


kulmann (approved):
Reasonable change 👍 Usually the same for forums etc... makes sense to do that for us as well.

micbar:
This can allow users to:

1. Do XSS attacks. Attackers can load images from random external domains which can inject javascript.
2. Data Exfiltration. Sensitive Data from openCloud can be transmittet to external image urls
3. Tracking
4. Using `*` allows images from `http://` sources which would cause mixed content issues

micbar:
I understand that users want to do fancy stuff. But I think we should have only images from opencloud server.

kulmann:
> I understand that users want to do fancy stuff. But I think we should have only images from opencloud server.

Then we need authenticated image caching via the opencloud server for images from external sources.

butonic:
um ... embedding images from anywhere? That is a step too far IMO. Images unfortunately can carry malware and I would prefer we do not open up this attac vector by default.

A more correct solution would be to use something like a `static.opencloud.com` or `img.opencloud.com` domain that can be used to fetch images that have been embedded into markdown files? The question is how we authenticate these requests.

Github generates a uuid for embedded images. The browser fetches them from sth like `https://cloud.githubusercontent.com/assets/711743/25689673/17296b42-30a8-11e7-9b62-90207b2a6686.png` ... no auth headers ... that is also problematic ...

🤔

kulmann:
Please understand: The UX is awful when I paste an external image-url and it won't be rendered. Let's find a proper solution for this. I'll create an issue.

butonic:
for reference, the attack vector is called `XSS Polyglot Payload`: https://www.researchgate.net/publication/386884386_Uncovering_XSS_Polyglot_Payload_Detection_with_Machine_Learning_Advancing_Web_Security_Against_Complex_Threats

butonic:
We could use our thumbnailing service to recreate the image.

kulmann:
https://github.com/opencloud-eu/opencloud/issues/1103

PR #1143, opened 2025-06-30
Make install.sh work with Linux aarch64

jnweiger:
My odroid M1 with ubuntu 24.04 reports this: uname -a
 Linux datenheim 6.6.0-odroid-arm64 #1 SMP PREEMPT Mon, 04 Nov 2024 09:52:06 +0000 aarch64 aarch64 aarch64 GNU/Linux
 
Without this patch, install.sh would download the x68_64 binary.

### DIFF
--- a/deployments/examples/bare-metal-simple/install.sh
+++ b/deployments/examples/bare-metal-simple/install.sh
@@ -69,6 +69,10 @@ if [[ $(uname -s) == "Darwin" && $(uname -m) == "arm64" ]]; then
     dlarch="arm64"
 fi
 
+if [[ $(uname -s) == "Linux" && $(uname -m) == "aarch64" ]]; then
+    dlarch="arm64"
+fi
+
 # ...results in the download file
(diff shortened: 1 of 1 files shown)


dragotin (approved):
Nice, thanks

deployments/examples/bare-metal-simple/install.sh
```diff
@@ -69,6 +69,10 @@ if [[ $(uname -s) == "Darwin" && $(uname -m) == "arm64" ]]; then
     dlarch="arm64"
 fi
 
+if [[ $(uname -s) == "Linux" && $(uname -m) == "aarch64" ]]; then
```
dragonchaser: can you maybe unifi the if-clause here and the one above into one?

jnweiger: @dragonchaser done.
I usually value readability and robustness over compactness and DRY. But not worth the discussion here.

jnweiger:
@dragonchaser re-review?

Issue #1166, opened 2025-07-03
Remove commit ID from released versions


ScharfViktor:
Steps: 
- run opencloud `OC_DOCKER_IMAGE=opencloudeu/opencloud-rolling` `OC_DOCKER_TAG=3.1.0`

Actual: 
<img width="177" alt="Image" src="https://github.com/user-attachments/assets/2fcedd19-61d8-4c09-bca5-f1a3ed8cb3b5" />

Expected: 
Released versions (tagged builds): `3.1.0`
Development/daily builds: 3.1.0-7dh9fd89 (keep commit ID)

PR #1167, opened 2025-07-03
delete commit id from release version

ScharfViktor:
related #1166

I believe it helps. If I correct got it 😀

```
// Parsed returns a semver Version
func Parsed() (version *semver.Version) {
	versionToParse := LatestTag
	// use the placeholder version if the tag is empty or when we are creating a daily build
	if Tag != "" && Tag != "daily" {
		versionToParse = Tag
	}
	version, err := semver.NewVersion(versionToParse)
	if err != nil {
		// this should never happen
		return &semver.Version{}
	}
	if String != "" {
		// We have no tagged version but a commitid
		nVersion, err := version.SetMetadata(String)
		if err != nil {
			return &semver.Version{}
		}
		version = &nVersion
	}
	return version
}
```

### For releases:

Tag = "v3.1.0" (from ldflags)
String = "" (empty, no commit hash added) // we set here empty string 
Result: 3.1.0 (clean version)

### For daily builds:

Tag = "daily"
String = "7dh9fd89" (commit hash)
Result: 3.1.0+dev-7dh9fd89 (with commit metadata)

### DIFF
--- a/.make/go.mk
+++ b/.make/go.mk
@@ -29,6 +29,8 @@ ifeq ($(VERSION), daily)
 	STRING ?= $(shell git rev-parse --short HEAD)
 else ifeq ($(VERSION),)
 	STRING ?= $(shell git rev-parse --short HEAD)
+else
+	STRING := 
 endif
 
 


micbar:
@ScharfViktor I think we should not change the code. The fix is in the pipeline.

`String` needs to be empty.

We had a wrong config for the release pipeline. the DOCKERLDFLAGS were wrong.

@fschade might help because that is all related to the complex Makefile inheritance.

micbar:
This has been the issue from the release pipeline.

You can see that `github.com/opencloud-eu/opencloud/pkg/version.String=7d6bc2a48` is set and
`github.com/opencloud-eu/opencloud/pkg/version.Tag="` is unset.

This needs to be fixed in the build environment.


```
#30 0.478 make: Entering directory '/opencloud/opencloud'
#30 3.336 GOOS=linux \
#30 3.336 GOARCH=arm64 \
#30 3.336 go build \
#30 3.336 	-tags 'netgo,disable_crypt,enable_vips' \
#30 3.336 	-buildmode=pie \
#30 3.336 	-trimpath \
#30 3.336 	-ldflags '-X google.golang.org/protobuf/reflect/protoregistry.conflictPolicy=warn -s -w -X "github.com/opencloud-eu/opencloud/pkg/version.String=7d6bc2a48" -X "github.com/opencloud-eu/opencloud/pkg/version.Tag=" -X "github.com/opencloud-eu/opencloud/pkg/version.Date=20250630" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseDataPathType=path" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseDataPathValue=/var/lib/opencloud" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseConfigPathType=path" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseConfigPathValue=/etc/opencloud" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseDataPathType=path" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseDataPathValue=/var/lib/opencloud" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseConfigPathType=path" -X "github.com/opencloud-eu/opencloud/pkg/config/defaults.BaseConfigPathValue=/etc/opencloud"' \
```

v-scharf:
> @ScharfViktor I think we should not change the code. The fix is in the pipeline.

The logic was modified in https://github.com/opencloud-eu/opencloud/pull/549
I believe this is just a logical improvement to handle the case properly.

Issue #1104, opened 2025-06-24
Research > Collabora: backend part for insert images from cloud (timeboxed 4pd)


db-ot:
# Description

### Initial Checklist

- [x] I understand this is a user story and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### User Story

> As a backend system, I want to expose an authenticated API that can be accessed from the Collabora frontend,
so that users can browse and insert images from OC - while researching how PROPFIND requests are used for resource discovery, since this mechanism is not yet fully understood.

## Acceptance Criteria

- ~~Investigate how PROPFIND requests are constructed and used in the Collabora frontend.~~
- ~~Explore how to expose an authenticated API endpoint for image browsing and insertion.~~
- Evaluate how to generate and return a presigned download URL for image access.

Note ( @kulmann ): the web ui does everything on its own regarding image selection for the user. We have a wrapper around the collabora iframe. The collabora iframe only sends a trigger to the wrapper that it wants to insert an image. The wrapper can respond async with an image URL. For this, web only needs a presigned url / url that can be used without auth headers for the image selected by the user. -> `downloadURL` prop in all authenticated webdav propfinds would be a nice solution. Needs research if this can be added without a serious performance drop.

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the OpenCloud charts is compatible

rhafer:
@kulmann Could you clarify what exactly is needed here? IIUC the whole PostMessage thing (`UI_InsertGraphic` / `Action_InsertGraphics`) is happening client side between the OpenCloud Web UI and the Collabora UI.

[...]

PR #1191, opened 2025-07-08
proxy(sign_url_auth): Allow to verify server signed URLs

rhafer:
With the ocdav service being able to provided signed download URLs we need the proxy to be able to verify the signatures.

This should also be a first step towards phasing out the weird ocs based client side signed urls.

Related Tickets: https://github.com/opencloud-eu/opencloud/issues/1104

This is still a bit of a PoC, but should be good enough for providing allowing some traction on: https://github.com/opencloud-eu/web/issues/704

Unfortunately in its current form it is a breaking change (a new configuration key is required) and would require a major version bump again. I'd be interested in suggestions about how we can make this non-breaking in a simple way.

To use this the new config variable `OC_URL_SIGNING_SHARED_SECRET` needs to be set. (Just use some randon string content)

### DIFF
--- a/services/ocdav/pkg/command/server.go
+++ b/services/ocdav/pkg/command/server.go
@@ -89,6 +89,7 @@ func Server(cfg *config.Config) *cli.Command {
 					ocdav.WithTraceProvider(traceProvider),
 					ocdav.RegisterTTL(registry.GetRegisterTTL()),
 					ocdav.RegisterInterval(registry.GetRegisterInterval()),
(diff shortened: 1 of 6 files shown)


AlexAndBear:
Please enable https://sdk.collaboraonline.com/docs/advanced_integration.html#enableinsertremoteimage

rhafer:
@AlexAndBear  should work now

image

AlexAndBear:
TYSM!

PR #1236, opened 2025-07-16
tests: Avoid static dates when creating shares

rhafer:
The expirationDate for new shares needs to be in the future. Using static dates in the tests means we need to update them every now an then. This adjusts the helpers to allow relative date formats like "+3 days" and "tomorrow" when creating shares.

### DIFF
--- a/tests/acceptance/features/apiSharingNgShareInvitation/updateShareInvitations.feature
+++ b/tests/acceptance/features/apiSharingNgShareInvitation/updateShareInvitations.feature
@@ -19,7 +19,7 @@ Feature: Update permission of a share
       | sharee             | Brian                |
       | shareType          | user                 |
       | permissionsRole    | Viewer               |
-      | expirationDateTime | 2025-07-15T14:00:00Z |
+      | expirationDateTime | tomorrow             |
     When user "Alice" updates the last resource share with the following properties using the Graph API:
       | space              | Personal             |
       | resource           | testfile.txt         |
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiSharingNgShareInvitation/updateShareInvitations.feature
```diff
@@ -717,7 +717,7 @@ Feature: Update permission of a share
       | sharee             | grp1                 |
       | shareType          | group                |
       | permissionsRole    | <permissions-role>   |
-      | expirationDateTime | 2025-07-15T14:00:00Z |
+      | expirationDateTime | tomorrow             |
     When user "Alice" updates the last drive share with the following using root endpoint of the Graph API:
       | expirationDateTime | 2200-07-15T14:00:00Z   |
```
AlexAndBear, line 722: As you already on it, and for consistency reasons, maybe you want to adjust that part and other occurrences as well ?

rhafer: Actually my intention was to fix the immediate problem, i.e. get the CI green again. Let's merge this first and I'll take at how hard it is to fix the other dates or create an issue if my php-fu is to weak.

rhafer: It seems to be more complicated that expected (at least for my noob level of php skills).  The test validates the update share response and expects the date to be exactly what was given in the request. 

@ScharfViktor do you think it is worth to invest more time there?

AlexAndBear: Maybe just simplify the process and adjust the tests to a different date. e.G 2200- which was done already in other tests.

individual-it: This specific test checks the date in the JSON response, so if you want to make it dynamic there one would have to teach the step `Then the JSON data of the response should match` to interpret `tomorrow`. e.g. by creating a new substitution.

Issue #877, opened 2025-05-19
[Epic] Multi tenancy


micbar:
# Description

## Version 1

- [ ] Every user has an ID which makes them part of a tenant (which is a subset) of all users in the directory (Tenant ID)
- [ ] Every user can only enumerate share recipients from its own tenant
- [ ] The backend enforces users to stay in their tenant (only one tenant per user)

butonic:
There are three possible points where we could check the tenant id:
- When searching recipients the list should be filtered by tenant
- When checking an invite the recipent, must be part of the same tenant
- When checking file access the space must be part of the same tenant

butonic:
We need to have some reporting for quota usage including versions and trash. Added an issue to track it.

db-ot:
Multi tenancy Epic here is MVP, 

remove admin role - https://github.com/opencloud-eu/opencloud-charts/issues/64 

admin role is effected by Multi tenancy but not part of the epic. As yesterday mentioned we will make here a notice for.

PR #1274, opened 2025-07-25
Tenant

aduffeck:
We have added support for tenantIds for users. Adapted the CreateShare Endpoints to respect theseif used.

refs #877


services/graph/pkg/identity/cs3.go
```diff
@@ -53,22 +53,43 @@ func (i *CS3) GetUser(ctx context.Context, userID string, _ *godata.GoDataReques
...
+		return nil, errorcode.New(errorcode.ServiceNotAvailable, err.Error())
+	case res.GetStatus().GetCode() == cs3rpc.Code_CODE_OK:
+		return CreateUserModelFromCS3(res.GetUser()), nil
+	case res.GetStatus().GetCode() == cs3rpc.Code_CODE_NOT_FOUND:
+		// If the user was not found by username, try to get it by user ID
+	default:
+		logger.Debug().Str("backend", "cs3").Err(err).Str("nameOrId", nameOrId).Msg("error sending get user by claim id grpc request")
+		return nil, errorcode.New(errorcode.GeneralException, res.GetStatus().GetMessage())
+
+	}
```
butonic, line 74: What if the search returns more than one user? names may not be unique

dragonchaser: refers to L58

dragonchaser: @butonic suggested to use upn to be safe

PR #1297, opened 2025-07-31
feat: introduce alias for thumbnail max dimensions and change default

kulmann:
## Description
Introduced an alias for the thumbnailer max dimensions, because setting max height and max width separately makes no sense at all. Photos in portrait and landscape orientation exist. Makes no sense to have a different max height than max width.
Also adjusted the default value, because 7680px is not enough for a 50MP photo.

## Motivation and Context
- easier config
- better defaults.

## Types of changes

- [ ] Technical debt

### DIFF
--- a/services/thumbnails/pkg/config/config.go
+++ b/services/thumbnails/pkg/config/config.go
@@ -44,7 +44,7 @@ type Thumbnail struct {
 	FontMapFile           string            `yaml:"font_map_file" env:"THUMBNAILS_TXT_FONTMAP_FILE" desc:"The path to a font file for txt thumbnails." introductionVersion:"1.0.0"`
 	TransferSecret        string            `yaml:"transfer_secret" env:"THUMBNAILS_TRANSFER_TOKEN" desc:"The secret to sign JWT to download the actual thumbnail file." introductionVersion:"1.0.0"`
 	DataEndpoint          string            `yaml:"data_endpoint" env:"THUMBNAILS_DATA_ENDPOINT" desc:"The HTTP endpoint where the actual thumbnail file can be downloaded." introductionVersion:"1.0.0"`
-	MaxInputWidth         int               `yaml:"max_input_width" env:"THUMBNAILS_MAX_INPUT_WIDTH" desc:"The maximum width of an input image which is being processed." introductionVersion:"1.0.0"`
-	MaxInputHeight        int               `yaml:"max_input_height" env:"THUMBNAILS_MAX_INPUT_HEIGHT" desc:"The maximum height of an input image which is being processed." introductionVersion:"1.0.0"`
+	MaxInputWidth         int               `yaml:"max_input_width" env:"THUMBNAILS_MAX_INPUT_WIDTH;THUMBNAILS_MAX_INPUT_DIMENSION" desc:"The maximum width of an input image which is being processed." introductionVersion:"1.0.0"`
+	MaxInputHeight        int               `yaml:"max_input_height" env:"THUMBNAILS_MAX_INPUT_HEIGHT;THUMBNAILS_MAX_INPUT_DIMENSION" desc:"The maximum height of an input image which is being processed." introductionVersion:"1.0.0"`
 	MaxInputImageFileSize string            `yaml:"max_input_image_file_size" env:"THUMBNAILS_MAX_INPUT_IMAGE_FILE_SIZE" desc:"The maximum file size of an input image which is being processed. Usable common abbreviations: [KB, KiB, MB, MiB, GB, GiB, TB, TiB, PB, PiB, EB, EiB], example: 2GB." introductionVersion:"1.0.0"`
 }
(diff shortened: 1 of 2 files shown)


kulmann:
@dragonchaser any comments about the var naming? "dimension" sounds like you wouldn't provide a number but a complete resolution as string (e.g. 9000x9000). But I don't have a better var name...

kulmann:
@dragonchaser I'd say it even makes sense to deprecate the max width and max height var names... but no idea how to do that. Do you agree? And if yes, could you help me with it?

Makes sense I write a proposal.

dragonchaser:
> @dragonchaser any comments about the var naming? "dimension" sounds like you wouldn't provide a number but a complete resolution as string (e.g. 9000x9000). But I don't have a better var name...

Yes right, but I have no alternative proposal.

services/thumbnails/pkg/config/config.go
```diff
@@ -44,7 +44,7 @@ type Thumbnail struct {
 	FontMapFile           string            `yaml:"font_map_file" env:"THUMBNAILS_TXT_FONTMAP_FILE" desc:"The path to a font file for txt thumbnails." introductionVersion:"1.0.0"`
 	TransferSecret        string            `yaml:"transfer_secret" env:"THUMBNAILS_TRANSFER_TOKEN" desc:"The secret to sign JWT to download the actual thumbnail file." introductionVersion:"1.0.0"`
 	DataEndpoint          string            `yaml:"data_endpoint" env:"THUMBNAILS_DATA_ENDPOINT" desc:"The HTTP endpoint where the actual thumbnail file can be downloaded." introductionVersion:"1.0.0"`
-	MaxInputWidth         int               `yaml:"max_input_width" env:"THUMBNAILS_MAX_INPUT_WIDTH" desc:"The maximum width of an input image which is being processed." introductionVersion:"1.0.0"`
-	MaxInputHeight        int               `yaml:"max_input_height" env:"THUMBNAILS_MAX_INPUT_HEIGHT" desc:"The maximum height of an input image which is being processed." introductionVersion:"1.0.0"`
+	MaxInputWidth         int               `yaml:"max_input_width" env:"THUMBNAILS_MAX_INPUT_WIDTH;THUMBNAILS_MAX_INPUT_DIMENSION" desc:"The maximum width of an input image which is being processed." introductionVersion:"1.0.0"`
```
dragonchaser, line 47:
```suggestion
	MaxInputWidth         int               `yaml:"max_input_width" env:"THUMBNAILS_MAX_INPUT_WIDTH" desc:"The maximum width of an input image which is being processed." introductionVersion:"1.0.0" deprecationVersion:"%%NEXT%%"`
```

services/thumbnails/pkg/config/config.go
```diff
@@ -44,7 +44,7 @@ type Thumbnail struct {
...
+	MaxInputHeight        int               `yaml:"max_input_height" env:"THUMBNAILS_MAX_INPUT_HEIGHT;THUMBNAILS_MAX_INPUT_DIMENSION" desc:"The maximum height of an input image which is being processed." introductionVersion:"1.0.0"`
```
dragonchaser, line 48:
```suggestion
	MaxInputHeight        int               `yaml:"max_input_height" env:"THUMBNAILS_MAX_INPUT_HEIGHT" desc:"The maximum height of an input image which is being processed." introductionVersion:"1.0.0" deprecationVersion:"%%NEXT%%"`
	MaxInputSideLength        int               `yaml:"max_input_side_length" env:"THUMBNAILS_MAX_INPUT_SIDE_LENGTH" desc:"The maximum side length of an input image which is being processed." introductionVersion:"%%NEXT%%"`
```

services/thumbnails/pkg/config/defaults/defaultconfig.go
```diff
@@ -55,8 +55,8 @@ func DefaultConfig() *config.Config {
 			RevaGateway:           shared.DefaultRevaConfig().Address,
 			CS3AllowInsecure:      false,
 			DataEndpoint:          "http://127.0.0.1:9186/thumbnails/data",
-			MaxInputWidth:         7680,
-			MaxInputHeight:        7680,
+			MaxInputWidth:         10000,
```
dragonchaser, line 58:
```suggestion
```

services/thumbnails/pkg/config/defaults/defaultconfig.go
```diff
@@ -55,8 +55,8 @@ func DefaultConfig() *config.Config {
...
+			MaxInputHeight:        10000,
```
dragonchaser, line 59:
```suggestion
			MaxInputSideLength:        10000,
```

dragonchaser (changes requested):

dragonchaser:
Unfortunately this needs more refactoring and rewriting of the values to the actual variables (I suggest if deprecated values are set to use the larger one).

AlexAndBear:
IHMO it would make more sense to have a env variable that limits the megapixel e.G maxInputMegaPixels and default to 50. By that we don't need to care about aspect ratio and so on

PR #1306, opened 2025-08-04
Improve indexing performance using batches

aduffeck:
Use batches to tremendously improve indexing/deletion performance (~50x).


services/search/pkg/engine/bleve.go
```diff
@@ -32,10 +34,16 @@ import (
 	searchQuery "github.com/opencloud-eu/opencloud/services/search/pkg/query"
 )
 
+const _batchSize = 500
```
dragonchaser, line 37: Do we really want to hard-code the batch-size? This might make sense to have a config var for that, depending on the systems usage (file-sizes etc.).

dragonchaser (changes requested):

dragonchaser (approved):

PR #1313, opened 2025-08-06
multiTenancyTests

ScharfViktor:
related https://github.com/opencloud-eu/opencloud/pull/1274

Test uses prepared LDAP users where we have users from two tenancy
tenancy1: `alice`, `brian` and `admin`(need for CI)
tenancy2: `carol` and `david`

How to run localy: 
- change https://github.com/opencloud-eu/opencloud/blob/main/devtools/deployments/opencloud_full/ldap.yml#L46 to `../../../tests/config/woodpecker/ldap:/ldifs`
- enable Ldap https://github.com/opencloud-eu/opencloud/blob/main/devtools/deployments/opencloud_full/.env#L254
- add envs to ldap.yml (opencloud->enviroment)
- - `OC_LDAP_USER_SCHEMA_TENANT_ID: "departmentNumber"`
- - `OC_LDAP_USER_ENABLED_ATTRIBUTE: ""`
- - coment out line`OC_LDAP_USER_FILTER: "(objectclass=inetOrgPerson)"`
- add envs to opencloud.yml (opencloud->enviroment)
- - `OC_MULTI_TENANT_ENABLED: "true"`
- - `GROUPS_DRIVER: "null"`
- - `GRAPH_IDENTITY_BACKEND: "cs3"`
- start opencloud `docker compose up -d`
- run test: `USE_PREPARED_LDAP_USERS=true TEST_SERVER_URL=https://cloud.opencloud.test BEHAT_FEATURE=tests/acceptance/features/apiTenancy/mutltiTenancy.feature make test-acceptance-api`

### DIFF
--- a/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
+++ b/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
@@ -1,21 +1,21 @@
 Feature: Multi-tenancy
-  I want to make sure that users from different companies are isolated from each other,
-  so that each tenants data and users remain private and secure.
+  I want to make sure that users from different tenants are isolated from each other,
+  so that each tenant's data and users remain private and secure.
 
   Note:
   All users are managed via LDAP and are assumed to exist.
   Tests will use existing users without creating or deleting them.
 
   Prepared LDAP users:
-    | user  | department | group                |
-    |-------|------------|----------------------|
-    | alice | 1a         | new-features-lovers  |
-    | brian | 1a         | -                    |
-    | carol | 2b         | -                    |
-    | david | 2b         | release-lover        |
+    | user  | tenant   | group                |
+    |-------|----------|----------------------|
+    | alice | tenant-1 | new-features-lovers  |
+    | brian | tenant-1 | -                    |
+    | carol | tenant-2 | -                    |
+    | david | tenant-2 | release-lover        |
 
 
-  Scenario: users from the same company can see each other
+  Scenario: users from the same tenant can see each other
     When user "Brian" searches for user "ali" using Graph API
     Then the HTTP status code should be "200"
     And the JSON data of the response should match
@@ -58,7 +58,7 @@ Feature: Multi-tenancy
       """
 
 
-  Scenario: users from different companies cannot see each other
+  Scenario: users from different tenants cannot see each other
     When user "David" searches for user "brian" using Graph API
     Then the HTTP status code should be "200"
     And the JSON data of the response should match
--- a/tests/config/woodpecker/ldap/20_users.ldif
+++ b/tests/config/woodpecker/ldap/20_users.ldif
@@ -10,7 +10,7 @@ cn: alice
 displayName: Alice Hansen
 description: Senior DevOps engineer responsible for cloud infrastructure automation.
 mail: alice@example.org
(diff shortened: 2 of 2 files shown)


v-scharf:
Test is failed but shouldn't

I tried to reproduce on fresh main and now it doesn't work as expected: 

steps: 

- change `openCloudTenantId` for `alan` and `lynn` in the `services/idm/ldif/demousers.ldif.tmpl`
- build `opencloud` and run `IDM_CREATE_DEMO_USERS=true OC_LDAP_USER_SCHEMA_TENANT_ID=openCloudTenantId OC_INSECURE=true PROXY_ENABLE_BASIC_AUTH=true opencloud/bin/opencloud server`
- `alan` tries to find `lynn` (same tenancy) - works fine 👍️
- `alan` tries to find `dennis` (different tenancy) - can find but shouldn't

@aduffeck @dragonchaser Maybe I forgot something?

v-scharf:
> @aduffeck @dragonchaser Maybe I forgot something?

I forgot to bump reva. re-run after rebase

tests/acceptance/features/apiTenancy/mutltiTenancy.feature
```diff
@@ -0,0 +1,77 @@
...
+              }
+            }
+          }
+        }
+      }
+      """
+
+
+  Scenario: users from different companies cannot see each other
+    When user "David" searches for user "brian" using Graph API
```
aduffeck, line 62: Maybe search for `ali` here as well? Otherwise it could happen that the test data gets changed in the future (e.g. Brian gets dropped) and then this test would give false positives.

individual-it (approved):

tests/acceptance/features/apiTenancy/mutltiTenancy.feature
```diff
@@ -0,0 +1,77 @@
...
+                }
+              }
+            }
+          }
+        }
+      }
+      """
+
+
+  Scenario: users from different companies cannot see each other
```
dragonchaser: nit picking: I'd prefer using the term tenants instead of companies, companies is confusing and might be used to describe things with separate instances.

tests/config/woodpecker/ldap/20_users.ldif
```diff
@@ -0,0 +1,74 @@
...
+objectClass: person
+objectClass: top
+uid: alice
+givenName: Alice
+sn: Hansen
+cn: alice
+displayName: Alice Hansen
+description: Senior DevOps engineer responsible for cloud infrastructure automation.
+mail: alice@example.org
+departmentNumber: cd22ea13-f6b4-4f5f-a2c2-69b5a0f07a1a
```
dragonchaser: another nit-pick: why use an uuid here? For tests a simple string should be sufficent (e.g.: tenant1), makes it easier to maintain/read

dragonchaser (changes requested):

dragonchaser (approved):

aduffeck (approved):
lgtm besides the nitpick

tests/acceptance/features/apiTenancy/mutltiTenancy.feature
v-scharf: I think this test cannot give false positives because when Brian gets dropped then -> we get a problem in the first test where we use `When user "Brian" searches for user "ali" using Graph API`

PR #1319, opened 2025-08-07
Drop the "InProgress" metric

aduffeck:
It didn't really make sense in multi-node setups because items could be started and finished on different nodes.

/cc @micbar

### DIFF
--- a/services/postprocessing/pkg/metrics/metrics.go
+++ b/services/postprocessing/pkg/metrics/metrics.go
@@ -36,12 +36,6 @@ var (
 		Name:      "events_redelivered",
 		Help:      "Number of redelivered events",
 	})
-	inProgress = promauto.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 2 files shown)


services/postprocessing/pkg/metrics/metrics.go
```diff
@@ -76,8 +69,6 @@ func New() *Metrics {
 		EventsOutstandingAcks: eventsOutstandingAcks,
 		EventsUnprocessed:     eventsUnprocessed,
 		EventsRedelivered:     eventsRedelivered,
-		InProgress:            inProgress,
-		Finished:              finished,
```
micbar, line 80: maybe one line too much.

aduffeck: Indeed! Fixed.

fschade (approved):

PR #1349, opened 2025-08-14
run most of the CI also on stable branches

individual-it:
## Description
I think any CI step that we run on the `main` branch, we also want to run on stable branches.
Is that a true assumption? @ScharfViktor @micbar 

## Related Issue
in preparation for running tests on cron jobs


individual-it:
my assumption was not correct, because that would also build and push docker images of the "stable" branch but the tag would be wrong.

PR #1352, opened 2025-08-14
Dockerfile cleanup

rhafer:
This tries to cleanup or multitude of Dockerfile a bit. By making the `make dev-docker` target use the `multiarch` Dockerfile and adding docker cache mounts for the go build caches, we no longer need specific Dockerfiles for the `dev-docker` target.


v-scharf:
I checked, and it works, but it took longer than building the arm64 container I need.
Perhaps we could leave an option of building a container for needed architecture?

v-scharf:
129,3s agains 5.8s

multiarch:
image

arm64:
image

rhafer:
> 129,3s agains 5.8s

The first run of this is expected to be some what slow. It needed to fetch an addtional image (the golang:alpine image wasn't in you local cache yet). Also it needed to do a full rebuild of the opencloud binary. Subsequent calls should be much faster because of the build cache.

Also I pushed another change to reduce the size of the docker build-context. That should shave off some time of the `transferring build context` step.

ScharfViktor (approved):
tested again. second running was 5.8s

PR #1408, opened 2025-08-27
fix(antivirus): the file bytesize differs if the file is larger than …

fschade:
…the max-scan-size which leads to a std-lib request validation error.

## Description
if the max-scan-size is set and the scan-mode is set to partial the actual file size will not match the partial file. That behavior leads to an error because the std-lib ( http/httputil, DumpRequestOut ) validates the body length which will not match the content-length in that scenario

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/antivirus/pkg/service/service.go
+++ b/services/antivirus/pkg/service/service.go
@@ -231,19 +231,21 @@ func (av Antivirus) process(ev events.StartPostprocessingStep) (scanners.Result,
 		return scanners.Result{ScanTime: time.Now()}, nil
 	}
 
+	filesize := ev.Filesize
 	headers := make(map[string]string)
 	switch {
 	case av.maxScanSize == 0:
 		// there is no size limit
 		break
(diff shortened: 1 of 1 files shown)


micbar:
I tested the fix on the cluster.

All files are scanned and get out of postprocessing.

I saw a new error, which seems to heal itself:

```
 {"level":"error","service":"antivirus","error":"read tcp 10.42.0.139:56182->10.43.109.232:1344: use of closed network connection\nclose tcp 10.42.0.139:56182->10.43.109.232:1344: use of closed network connection","uploadid":"66c2a684-8839-46da-9903-77e5f4ebc8ac","time":"2025-08-28T19:17:01Z","message":"error scanning file"}
```

@butonic Could that be an error by the icap service?

micbar (approved):
LGTM!

PR #1412, opened 2025-08-27
Correct app store URL

### LINKED ISSUE #1411: App store broken in release binaries
## Describe the bug

By default, the app store fails to load when using the release binaries without a csp.yaml.

### Cause

The embedded csp.yaml permits `opencloud-eu/awesome` but that repository is actually called `opencloud-eu/awesome-apps`. The URL is correct in the examples at `deployments/examples/opencloud_full/config/opencloud/csp.yaml`, but not in the embedded defaults at `services/proxy/pkg/config/csp.yaml`.

### Solution

A trivial PR that corrects the URLs in both `csp.yaml` and the single test that uses the same headers.

## Steps to reproduce

1. Minimally configure opencloud from a 3.3.0 release binary
2. Attempt to access app store from web UI
3. No apps listed, CSP errors in console

## Expected behavior

Working app store without any need for further configuration, just like the other installation methods.

## Actual behavior

Broken app store that requires further configuration.

## Setup

3.3.0 amd64 `opencloud init && opencloud server`

tylerlm:
## Description

This PR makes the app store work in the release binaries.

It replaces `awesome` with `awesome-apps` in `services/proxy/pkg/config/csp.yaml` and `tests/acceptance/features/coreApiWebdavOperations/downloadFile.feature` which includes the same headers.

## Related Issue

- Fixes #1411

## Motivation and Context

By default, the app store fails to load when using the release binaries. This PR assumes that is not intended behaviour.

## How Has This Been Tested?

- Rebuilt successfully, embedded csp.yaml is now correct, app store works using the resulting binary.

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


kulmann (approved):
Nice catch, thank you! 🥇

Issue #1414, opened 2025-08-28
Improve logging in supervised crashes


pbleser-oc:
## Is your feature request related to a problem? Please describe.
When running in supervised ("single binary") mode and a panic occurs in a service, the event is logged with the level `INFO` (instead of `ERROR` or `FATAL`), which makes it difficult to track down and is very likely to go unnoticed for system administrators:

```go
s.Log.Info().Str("event", e.String()).Msg(fmt.Sprintf("supervisor: %v", e.Map()["supervisor_name"]))
```

## Describe the solution you'd like
Log the event using the level `FATAL`, and exposing more of the attributes as part of the logging event payload.

PR #1415, opened 2025-08-28
enhancement(runtime): improve logging in supervised crashes

pbleser-oc:
## Description
When events occur in the supervisor, log them as `FATAL` when appropriate (panic and termination) instead of always using `INFO`, and include more information such as the service name and the current failures, etc..., rather than the generic message that was used so far.

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/1414

## Motivation and Context
Whenever a panic crash of the single binary occurs, the logging message that contains the reason and stacktrace is potentially hidden due to its `INFO` loglevel, which is typically filtered out in productive environments, or in development environments when the other services are too noisy.

With this change, it is correctly logged as `FATAL`, which should always be caught in productive environments.

## How Has This Been Tested?
Has only been tested locally by triggering a panic in another service on purpose.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt

It could potentially be seen as a breaking change if system administrators have already set up alerting filters based on the previous log event payload, which does change now (different log level, different attributes, different message.)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

There is no documentation to change because the backend error logs are currently not documented in the first place.

If we do consider the log messages "public API", then we might want to include documentation about this change in the release notes.


opencloud/pkg/runtime/service/service.go
```diff
@@ -374,7 +374,24 @@ func Start(ctx context.Context, o ...Option) error {
 					cancel()
 				}
 			}
-			s.Log.Info().Str("event", e.String()).Msg(fmt.Sprintf("supervisor: %v", e.Map()["supervisor_name"]))
+			switch ev := e.(type) {
+			case suture.EventServicePanic:
+				s.Log.Fatal().Str("event", e.String()).Str("service", ev.ServiceName).Str("supervisor", ev.SupervisorName).
+					Bool("restarting", ev.Restarting).Float64("failures", ev.CurrentFailures).Float64("threshold", ev.FailureThreshold).
+					Msgf("service panic: %v", ev.PanicMsg)
```
rhafer: Hm, not sure about this. Wouldn't it be more straightforward to just set `PassThroughPanics` on the `suture.Spec{}` ? If we really want that behavior.

pbleser-oc:
> I am a bit unsure about the `Fatal` log level. Isn't one of the main goals of the of the Suture module to restart service when they crash? If we log at the `Fatal` level that point is basically moot, isn't it?

Pushed a change to only use `FATAL` when the suture service decides not to restart, and using `ERROR` instead when it is going to restart the service.

pbleser-oc:
> For the panics I think using `PassThroughPanics` might be the setting we want to enable to terminate when one of the services panics.

Probably difficult to wager since it depends on how liberal our dependencies but also our own codebase is with the use of panics. I'm not sure all the vendored packages we have only make use of `panic()` when it's a fatal error that warrants a crash, something that should typically only happen when e.g. you fail to allocate memory (although I suppose that is opinionated as well), not when a configuration setting is wrong.

Arguably, and in favour of @rhafer 's  proposal to let panics pass through: can we at least be confident that all (or the vast majority) of calls to `panic()` everywhere are for issues that cannot be resolved by retrying (e.g. (static) configuration errors)? Because, then, there is really no point in restarting the service in the first place, if it's just going to fail again the next time it runs into the same code path...

opencloud/pkg/runtime/service/service.go
butonic: Panics are quite rare and AFAIR we were always the ones resposible for making a panic call, or reva. So `PassThroughPanics` is the rigth thing to do IMO. I really like the improvement in log level handling.

That being said, I would move the message to a dedicated property:
```suggestion
					Str("message", ev.PanicMsg).Msg("service panic")
```

butonic:
might need a rebase

micbar:
@butonic @dragonchaser @rhafer Let us merge this PR

rhafer:
@pbleser-oc  Would you mind rebasing this one?

pbleser-oc:
Done, rebased on 500487f2fadb58b0f53b9d1cba44c081c84dafe5

opencloud/pkg/runtime/service/service.go
```diff
@@ -382,7 +382,32 @@ func Start(ctx context.Context, o ...Option) error {
 					cancel()
 				}
 			}
-			s.Log.Info().Str("event", e.String()).Msg(fmt.Sprintf("supervisor: %v", e.Map()["supervisor_name"]))
```
micbar, line 385: Ci says that this was the only occurence of `fmt`. It needs to be removed from the imports to compile again.

pbleser-oc:
Fixed and added @butonic 's suggestion at https://github.com/opencloud-eu/opencloud/pull/1415#discussion_r2367479201

PR #1416, opened 2025-08-28
Nats connection names

butonic:
port of https://github.com/owncloud/ocis/pull/11144

first half of https://github.com/opencloud-eu/opencloud/pull/1478

### DIFF
--- a/pkg/generators/natsnames.go
+++ b/pkg/generators/natsnames.go
@@ -33,12 +33,9 @@ func GenerateConnectionName(service string, ntype NType) string {
 
 // firstNRunes returns the first n runes of a string
 func firstNRunes(s string, n int) string {
-	i := 0
-	for j := range s {
-		if i == n {
-			return s[:j]
-		}
-		i++
+	runes := []rune(s)
+	if n > len(runes) {
+		n = len(runes)
 	}
-	return s
+	return string(runes[:n])
 }


pkg/generators/natsnames.go
```diff
@@ -0,0 +1,38 @@
...
+
+func firstNRunes(s string, n int) string {
+	i := 0
+	for j := range s {
+		if i == n {
+			return s[:j]
+		}
+		i++
+	}
+	return s
```
rhafer: Is it me or is that somewhat overcomplicated?

```
if len(s) > n {
    return s[:n]
}
return s
```

:thinking:

butonic: `len(s)` returns the length of s in bytes, not in runes:
```go
s := "café"  // 5 bytes, 4 runes
n := 3
```
would return `s[:3]` which gives us 3 bytes, but since "é" is 2 bytes in UTF-8, this would truncate the character and give us invalid UTF-8. 😢

but the code is indeed hard to digest. Maybe this makes it clearer:
```go
func firstNRunes(s string, n int) string {    
    runes := []rune(s)
    if n > len(runes) {
        n = len(runes)
    }
    return string(runes[:n])
}
```

rhafer: Hm, true. I always forget about that 🤦 

To my defense: Currently `firstNRunes` is currently only called with the output of os.Hostname(). People using non-ASCII characters in a hostname should be punished 😄

butonic: I used the rune version to make it more obvious ... although I agree that in production this should never be a problem

PR #1420, opened 2025-08-29
[full-ci] run tests with VIPS enabled

individual-it:
## Description
Run acceptance tests with VIPS by default

## Related Issue
https://github.com/opencloud-eu/qa/issues/43
https://github.com/opencloud-eu/qa/issues/66

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


v-scharf:
strict comparing for image files works fine but for .txt file fails in CI (but works fine localy)



```
@need-build-with-VIPS
  Scenario Outline: it should update the preview content if the file content is updated (content with UTF chars)
    Given using <dav-path-version> DAV path
    And user "Alice" has uploaded file "filesForUpload/lorem.txt" to "/lorem.txt"
    And user "Alice" has uploaded file with content "ओनक्लाउड फाएल शेरिङ्ग एन्ड सिन्किङ" to "/lorem.txt"
    When user "Alice" downloads the preview of "/lorem.txt" with width "32" and height "32" using the WebDAV API
    Then the HTTP status code should be "200"
    And the downloaded image should be "32" pixels wide and "32" pixels high
    And the downloaded preview content should match with "unicode-fixture.png" fixtures preview content
    Examples:
      | dav-path-version |
      | old              |
      | new              |
      | spaces           |
```

Perhaps my `libvips` library on Mac works a little differently than in CI. @individual-it, could you generate the `unicode-fixture.png` file on your laptop using a similar version of `apt-get install libvips42`? Or I can change this check using a tolerance.

how to do it: 

- upload file like here: `And user "Alice" has uploaded file with content "ओनक्लाउड फाएल शेरिङ्ग एन्ड सिन्किङ" to "/lorem.txt"`
- `curl -vk -u alice:123456 -o unicode-fixture.png 'https://cloud.opencloud.test/remote.php/webdav/lorem.txt?x=32&y=32&preview=1'`
- replace file to `tests/acceptance/fixtures/unicode-fixture.png`

individual-it:
@ScharfViktor can you try with a bigger size preview? because for me (using the docker container) it does not work. It only returns squares: 
unicode-fixture-big

v-scharf:
I got the fixure image on a remote Linux computer with `libvips42` installed. The test passed locally, let's see what CI says.

individual-it:
yes, but now there are no special non-latin UTF characters.
grafik
I think this is a bug, the preview should also work with nepali and other characters

v-scharf:
> think this is a bug, the preview should also work with nepali and other characters

true, preview works only with latin and cyrillic

image

v-scharf:
but the test doesn't work. even when we use Latin

tests/acceptance/features/coreApiWebdavPreviews/previews.feature
```diff
@@ -236,21 +236,35 @@ Feature: previews of files downloaded through the webdav API
...
-
+  @need-build-with-VIPS @issue-1451
   Scenario Outline: it should update the preview content if the file content is updated (content with UTF chars)
     Given using <dav-path-version> DAV path
     And user "Alice" has uploaded file "filesForUpload/lorem.txt" to "/lorem.txt"
     And user "Alice" has uploaded file with content "ओनक्लाउड फाएल शेरिङ्ग एन्ड सिन्किङ" to "/lorem.txt"
-    When user "Alice" downloads the preview of "/lorem.txt" with width "32" and height "32" using the WebDAV API
+    When user "Alice" downloads the preview of "/lorem.txt" with width "1200" and height "1200" using the WebDAV API
     Then the HTTP status code should be "200"
-    And the downloaded image should be "32" pixels wide and "32" pixels high
```
individual-it, line 246: you are checking the size in every other test except of here, any special reason for that?
theoretically we don't need to check the size, because we are trying to match the file anyway. so if the size would be wrong the test would also fail

tests/acceptance/features/coreApiWebdavPreviews/previews.feature
```diff
@@ -236,21 +236,35 @@ Feature: previews of files downloaded through the webdav API
...
       | spaces           |
 
-
+  @need-build-with-VIPS @issue-1451
   Scenario Outline: it should update the preview content if the file content is updated (content with UTF chars)
     Given using <dav-path-version> DAV path
     And user "Alice" has uploaded file "filesForUpload/lorem.txt" to "/lorem.txt"
     And user "Alice" has uploaded file with content "ओनक्लाउड फाएल शेरिङ्ग एन्ड सिन्किङ" to "/lorem.txt"
-    When user "Alice" downloads the preview of "/lorem.txt" with width "32" and height "32" using the WebDAV API
+    When user "Alice" downloads the preview of "/lorem.txt" with width "1200" and height "1200" using the WebDAV API
```
individual-it, line 244: your added fixture is 32x32 px

tests/acceptance/features/coreApiWebdavPreviews/previews.feature
```diff
@@ -368,6 +382,29 @@ Feature: previews of files downloaded through the webdav API
       | spaces           | resize    | resize.png     |
       | spaces           | thumbnail | thumbnail.png  |
 
+  @need-build-with-VIPS
+  Scenario Outline: download previews of an image with different processors
+    Given using <dav-path-version> DAV path
+    And user "Alice" has uploaded file "filesForUpload/testavatar.jpg" to "/testimage.jpg"
+    When user "Alice" downloads the preview of "/testimage.jpg" with width "1200" and height "1200" and processor <processor> using the WebDAV API
+    Then the HTTP status code should be "200"
+    And the downloaded image should be "1920" pixels wide and "1920" pixels high
```
individual-it: why would that image be 1920px big if you have requested 1200px?

tests/acceptance/features/coreApiWebdavPreviews/previews.feature
v-scharf: I know, after fixing #1451 I'll generate new fixture file (without squares)

tests/acceptance/features/coreApiWebdavPreviews/previews.feature
v-scharf: I played with getting different image size and added extra test which I forgot to delete

PR #1443, opened 2025-09-02
Store user state in system storage

dragonchaser:


services/graph/pkg/config/config.go
```diff
@@ -42,15 +42,17 @@ type Config struct {
...
 	WebDavPath                      string `yaml:"webdav_path" env:"GRAPH_SPACES_WEBDAV_PATH" desc:"The WebDAV sub-path for spaces." introductionVersion:"1.0.0"`
 	DefaultQuota                    string `yaml:"default_quota" env:"GRAPH_SPACES_DEFAULT_QUOTA" desc:"The default quota in bytes." introductionVersion:"1.0.0"`
-	ExtendedSpacePropertiesCacheTTL int    `yaml:"extended_space_properties_cache_ttl" env:"GRAPH_SPACES_EXTENDED_SPACE_PROPERTIES_CACHE_TTL" desc:"Max TTL in seconds for the spaces property cache." introductionVersion:"1.0.0"`
-	UsersCacheTTL                   int    `yaml:"users_cache_ttl" env:"GRAPH_SPACES_USERS_CACHE_TTL" desc:"Max TTL in seconds for the spaces users cache." introductionVersion:"1.0.0"`
-	GroupsCacheTTL                  int    `yaml:"groups_cache_ttl" env:"GRAPH_SPACES_GROUPS_CACHE_TTL" desc:"Max TTL in seconds for the spaces groups cache." introductionVersion:"1.0.0"`
+	ExtendedSpacePropertiesCacheTTL int    `yaml:"extended_space_properties_STORE_ttl" env:"GRAPH_SPACES_EXTENDED_SPACE_PROPERTIES_STORE_TTL" desc:"Max TTL in seconds for the spaces property cache." introductionVersion:"1.0.0"`
```
rhafer: I guess the capital `STORE` in the yaml tag was some search and replace going wild?

Also, do we really want to rename this setting? Why? What about backwards compatibility?

The some goes true for the `GRAPH_SPACES_USERS_CACHE_TTL` to `GRAPH_SPACES_USERS_STORE_TTL` rename.

services/graph/pkg/service/v0/service.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/opencloud-eu/reva/v2/pkg/store"
 	"github.com/opencloud-eu/reva/v2/pkg/utils"
 	"github.com/opencloud-eu/reva/v2/pkg/utils/ldap"
+	"github.com/pkg/errors"
```
rhafer: Are you sure about this? What's wrong with the std `"errors"` package?

services/graph/pkg/service/v0/option.go
```diff
@@ -42,6 +43,7 @@ type Options struct {
 	SearchService            searchsvc.SearchProviderService
 	KeycloakClient           keycloak.Client
 	EventHistoryClient       ehsvc.EventHistoryService
+	Store                    microstore.Store
```
rhafer: What is this good for? I might be overlooking something, but is it actually used anywhere?

services/graph/pkg/config/config.go
```diff
@@ -169,3 +171,14 @@ type Metadata struct {
 	SystemUserIDP    string `yaml:"system_user_idp" env:"OC_SYSTEM_USER_IDP;GRAPH_SYSTEM_USER_IDP" desc:"IDP of the OpenCloud STORAGE-SYSTEM system user." introductionVersion:"%%NEXT%%"`
 	SystemUserAPIKey string `yaml:"system_user_api_key" env:"OC_SYSTEM_USER_API_KEY" desc:"API key for the STORAGE-SYSTEM system user." introductionVersion:"%%NEXT%%"`
 }
+
+// Store configures the store to use
+type Store struct {
+	Store        string        `yaml:"store" env:"OC_PERSISTENT_STORE;GRAPH_STORE" desc:"The type of the store. Supported values are: 'memory', 'nats-js-kv', 'redis-sentinel', 'noop'. See the text description for details." introductionVersion:"1.0.0"`
```
rhafer: AFAICS you're using the nats-kv module directly. Why do we need options to use a different store?

services/graph/pkg/config/config.go
```diff
@@ -169,3 +171,14 @@ type Metadata struct {
...
+	Nodes        []string      `yaml:"nodes" env:"OC_PERSISTENT_STORE_NODES;GRAPH_STORE_NODES" desc:"A list of nodes to access the configured store. This has no effect when 'memory' store is configured. Note that the behaviour how nodes are used is dependent on the library of the configured store. See the Environment Variable Types description for more details." introductionVersion:"1.0.0"`
+	Database     string        `yaml:"database" env:"GRAPH_STORE_DATABASE" desc:"The database name the configured store should use." introductionVersion:"1.0.0"`
+	Table        string        `yaml:"table" env:"GRAPH_STORE_TABLE" desc:"The database table the store should use." introductionVersion:"1.0.0"`
```
rhafer: AFAIK natkv doens't have the concept of tables. So we don't need this setting.

services/graph/pkg/config/config.go
```diff
@@ -169,3 +171,14 @@ type Metadata struct {
...
+	TTL          time.Duration `yaml:"ttl" env:"OC_PERSISTENT_STORE_TTL;GRAPH_STORE_TTL" desc:"Time to live for events in the store. See the Environment Variable Types description for more details." introductionVersion:"1.0.0"`
```
rhafer: What would we need the ttl for?

services/graph/pkg/server/http/option.go
```diff
@@ -22,6 +23,7 @@ type Options struct {
 	Flags         []cli.Flag
 	Namespace     string
 	TraceProvider trace.TracerProvider
+	Store         microstore.Store
```
rhafer: This store option is passed through several layer of the code, but I don't think it is used anywhere.

rhafer (changes requested):
Are you sure you pushed the right thing here? This looks weird. See my comments.

services/graph/pkg/service/v0/service.go
dragonchaser: Line 182 depends on errors.Wrapf which is old code and for some reason part of the included package.... confuses me a bit ATM

services/graph/pkg/command/server.go
```diff
@@ -44,6 +46,37 @@ func Server(cfg *config.Config) *cli.Command {
...
+
+			js, err := conn.JetStream()
+			if err != nil {
+				return err
+			}
+
+			kv, err := js.KeyValue(cfg.Store.Database)
+			if err != nil {
+				if !errors.Is(err, nats.ErrBucketNotFound) {
+					return errors.Wrapf(err, "Failed to get bucket (%s)", cfg.Store.Database)
```
rhafer: Please use `fmt.Errorf()` do get rid of the `"github.com/pkg/errors"` usage here. `github.com/pkg/errors` is unmaintained and mostly replaced by standard go features nowadays. 

I know were still using it in some other place but we should not introduce new code using it.

services/graph/pkg/command/server.go
```diff
@@ -44,6 +46,37 @@ func Server(cfg *config.Config) *cli.Command {
...
+				}
+
+				kv, err = js.CreateKeyValue(&nats.KeyValueConfig{
+					Bucket: cfg.Store.Database,
+				})
+				if err != nil {
+					return errors.Wrapf(err, "Failed to create bucket (%s)", cfg.Store.Database)
```
rhafer: Please use `fmt.Errorf()` here as well.

services/graph/pkg/config/config.go
```diff
@@ -168,3 +170,11 @@ type Metadata struct {
...
 }
+
+// Store configures the store to use
+type Store struct {
+	Nodes        []string `yaml:"nodes" env:"OC_PERSISTENT_STORE_NODES;GRAPH_STORE_NODES" desc:"A list of nodes to access the configured store. This has no effect when 'memory' store is configured. Note that the behaviour how nodes are used is dependent on the library of the configured store. See the Environment Variable Types description for more details." introductionVersion:"1.0.0"`
+	Database     string   `yaml:"database" env:"GRAPH_STORE_DATABASE" desc:"The database name the configured store should use." introductionVersion:"1.0.0"`
+	AuthUsername string   `yaml:"username" env:"OC_PERSISTENT_STORE_AUTH_USERNAME;GRAPH_STORE_AUTH_USERNAME" desc:"The username to authenticate with the store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"1.0.0"`
+	AuthPassword string   `yaml:"password" env:"OC_PERSISTENT_STORE_AUTH_PASSWORD;GRAPH_STORE_AUTH_PASSWORD" desc:"The password to authenticate with the store. Only applies when store type 'nats-js-kv' is configured." introductionVersion:"1.0.0"`
```
rhafer, line 179: It seems we're not using the `AuthUsername` and `AuthPassword` settings anywhere. Shouldn't we set the according nats options somewhere?

services/graph/pkg/command/server.go
```diff
@@ -44,6 +46,37 @@ func Server(cfg *config.Config) *cli.Command {
...
+			//Connect to NATS servers
+			natsOptions := nats.Options{
+				Servers: cfg.Store.Nodes,
+			}
+			conn, err := natsOptions.Connect()
+			if err != nil {
+				return err
+			}
+
+			js, err := conn.JetStream()
```
rhafer: Upstream nats nowadays recommends to use the `jetstream` module for jetstream (see: https://github.com/nats-io/nats.go/blob/main/jetstream/README.md#basic-usage-of-kv-bucket). We've e.g. already switched the reva events to using that module.

The APIs you're using here are considered legacy (https://github.com/nats-io/nats.go/blob/main/legacy_jetstream.md). IMO we shouldn't add new code using it.

rhafer (changes requested):
I've some concerns regarding the usage of the legacy jetstream API. I think we should use the new APIs for new code we introduce (see my inline comments).

Also it seems you forgot to add the nats mocks to the PR, please add them (or the adequate ones for the newer API).

fschade:
i took it over as good as i can till @dragonchaser is back, 
THX @rhafer for the detailed review!

rhafer (approved):
Thanks @fschade  for jumping in. Looks good to me now. Let's see what the ci thinks of it.

PR #1474, opened 2025-09-08
remove opencloud_full from the read.me and add opencloud-compose instead

Svanvith:
Changes like in the docs because the files in the docs were generated from the read.me files here


Svanvith:
Was by accident

rhafer:
> Was by accident

I guess you also need to restore the original eicar zip file as well.

rhafer (approved):

PR #1478, opened 2025-09-08
Fix graceful shutdown

butonic:
backport of https://github.com/owncloud/ocis/pull/11295
includes https://github.com/opencloud-eu/opencloud/pull/1416 so we should merge that first


butonic:
I rebased this PR after the pending PRs got merged. It does include a reva bump which also pulls in other vendor changes.

Other than that the PR changes every service to use the runner package. This allows a clean graceful shutdown of every service.

It can be tested by sending a SIGTERM to the process or hitting Ctrl-C. The log will then show that all services are shutting down.

Without this PR the first reva service completing his Stop function would `exit(0)` preventing other services from shutting down properly.

micbar:
> It can be tested by sending a SIGTERM to the process or hitting Ctrl-C. The log will then show that all services are shutting down.

On which log level do we log that? Info level?

opencloud/pkg/runtime/service/service.go
```diff
@@ -501,20 +502,48 @@ func (s *Service) List(_ struct{}, reply *string) error {
...
+	}()
+
 	for sName := range s.serviceToken {
 		for i := range s.serviceToken[sName] {
-			if err := s.Supervisor.Remove(s.serviceToken[sName][i]); err != nil {
-				s.Log.Error().Err(err).Str("service", "runtime service").Msgf("terminating with signal: %v", s)
-			}
+			wg.Add(1)
+			go func() {
+				s.Log.Warn().Msgf("call supervisor RemoveAndWait for %s", sName)
```
micbar: I do not understand the Log Levels. Why is that warn level?

I can see the consistency of L524 and L529, makes sense to log that at the same level, that is the happy path when everything is doing well. 527 is the error path, fine with that.

opencloud/pkg/runtime/service/service.go
```diff
@@ -501,20 +502,48 @@ func (s *Service) List(_ struct{}, reply *string) error {
...
+
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-time.After(_defaultInterruptTimeoutDuration):
+		s.Log.Fatal().Msg("ocis graceful shutdown timeout reached, terminating")
```
micbar: makes sense.

opencloud/pkg/runtime/service/service.go
```diff
@@ -501,20 +502,48 @@ func (s *Service) List(_ struct{}, reply *string) error {
...
+	case <-done:
+		s.Log.Info().Msg("all ocis services gracefully stopped")
```
micbar: L517 is logging on Info level. This line too, while L524 and L529 are logging on warn level?

opencloud/pkg/runtime/service/service.go
butonic: moving to debug, and dropping msgf

PR #1495, opened 2025-09-11
Allow disabling search servers

butonic:
This PR introduces two new env vars: SEARCH_GRPC_DISABLED and SEARCH_EVENTS_DISABLED. They allow disabling the grpc request handler and the events listener. This can be used to split the functionality when trying to run an api handler pod and a separate event handler pod in kubernetes or on docker-compose scenarios.

Builds upon https://github.com/opencloud-eu/opencloud/pull/1478 aand should be rebased once that has been merged.

I also had to move the Searcher initialization to the start command and property inject the dependency to the two servers.


butonic:
@fschade I rebased this... locally I can search. Waiting for CI...

fschade (approved):

PR #1518, opened 2025-09-18
enhancement(docs): describe what and why ADRs

fschade:
## Description
ADRs are a good way to document decisions and provide some background of the hows and whys.
We started using them already, but in my opinion there where some weak spots on when to write such ADRs and how to structure them.

https://github.com/opencloud-eu/internal/issues/178

### DIFF
--- a/docs/adr/README.md
+++ b/docs/adr/README.md
@@ -33,7 +33,7 @@ Use an ADR to document decisions which are significant, such as:
 
 - **Location**: Store all ADRs as Markdown files in this folder.
 - **Format**: Use [Markdown](https://commonmark.org/).
-- **Naming**: Use a consistent naming convention, e.g., `0001-descriptive-title.md`.
+- **Naming**: Adhere to the naming convention, e.g., `0001-descriptive-title.md`.
 
 ### ADR Template
 


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -8,7 +8,10 @@ assignees: ''
 ---
 
 ## Is your feature request related to a problem? Please describe.
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+A clear and concise description of what the problem is.
+Ex. I'm always frustrated when [...].
+
+Please consider writing an [ADR](../../docs/adr) (Architecture Decision Record) if the feature has architectural implications.
```
aduffeck, line 14: Is an ADR really something regular users would initiate? To me it feels more like something the team would start based on input from the users.

docs/adr/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+* It substantially affects the architecture, design, or direction of OpenCloud.
+* It involves trade-offs between multiple options.
+* It needs Team consensus or input from multiple stakeholders.
+
+## Writing ADRs
+
+- **Location**: Store all ADRs as Markdown files in this folder.
+- **Format**: Use [Markdown](https://commonmark.org/).
+- **Naming**: Use a consistent naming convention, e.g., `0001-descriptive-title.md`.
```
aduffeck, line 36:
```suggestion
- **Naming**: Adhere to the naming convention, e.g., `0001-descriptive-title.md`.
```

.github/ISSUE_TEMPLATE/feature_request.md
fschade: good point, i also was not sure if or if not... i reverted.

PR #1527, opened 2025-09-22
fix: make dev docker builds possible with reva replace

micbar:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/opencloud/docker/Dockerfile.multiarch
+++ b/opencloud/docker/Dockerfile.multiarch
@@ -7,11 +7,11 @@ ARG STRING
 RUN apk add bash make git curl gcc musl-dev libc-dev binutils-gold inotify-tools vips-dev
 
 WORKDIR /opencloud
-RUN --mount=type=bind,target=/opencloud \
+RUN --mount=type=bind,target=/opencloud,rw\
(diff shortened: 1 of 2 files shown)


opencloud/docker/Dockerfile.multiarch
```diff
@@ -7,11 +7,11 @@ ARG STRING
 RUN apk add bash make git curl gcc musl-dev libc-dev binutils-gold inotify-tools vips-dev
 
 WORKDIR /opencloud
-RUN --mount=type=bind,target=/opencloud \
+RUN --mount=type=bind,target=/opencloud,rw\
```
rhafer, line 10: Should we rename this to make it obvious that this is not the opencloud git base directory. But one level up?  How about `/src`?

micbar: Ci is not passing because the folder has a different name in ci.

rhafer:
#1778 was merged instead of this.

Issue #897, opened 2025-05-20
Duplicate OIDC authentication code


butonic:
Currenty, the opencloud proxy implements its own oidc authentication and ignores the implementation of the oidc auth manager in reva.

This prevents customizing the user authentication by replacing the CS3 auth manager that e.g. knows how to identify lightweight accounts based on email prefixes.

IIRC this has been known for a while: https://github.com/owncloud/ocis/issues/4701

@rhafer why auth-bearer and not auth-oidc?

butonic:
cc @glpatcern @jessegeens @micbar

rhafer:
> [@rhafer](https://github.com/rhafer) why auth-bearer and not auth-oidc?

I guess that is basically the same. In OpenCloud (as well as in oCIS) the service for the reva `oidc` auth manager is currently called `auth-bearer` (https://github.com/opencloud-eu/opencloud/tree/main/services/auth-bearer) for what ever reason. As it is basically unmaintained I guess we can also rename it to `auth-oidc`.

rhafer:
BTW, I don't think this should be treated as a bug. It probably fits into the technical debt category (that would be the Type:Maintenance label)

PR #1536, opened 2025-09-22
allow running auth-bearer as part of the runtime

butonic:
I wondered why this feature failed:
```
  @env-config @issue-10661
  Scenario: check extra services readiness
    Given the following configs have been set:
      | config                 | value        |
      | OC_ADD_RUN_SERVICES  | auth-bearer  |
      | AUTH_BEARER_DEBUG_ADDR | 0.0.0.0:9149 |
    When a user requests these URLs with "GET" and no authentication
      | endpoint                               | service     |
      | http://%base_url_hostname%:9149/readyz | auth-bearer |
    Then the HTTP status code of responses on all endpoints should be "200"
```
The auth-bearer service is not started when `OC_ADD_RUN_SERVICES="auth-bearer"`. This pr changes that.

Maybe we should log an error if the config contains an unknown service?


butonic:
then we should delete the code

rhafer:
either that or finally make it useful again (#897)

PR #1540, opened 2025-09-23
cleanup: Remove some unused code

rhafer:
This removes some remains of the old continous-deployment systems


flimmy:
> Is there any deployment test left? Maybe we should have one of the dev deployments still in here to know if a change breakes the setup.
> 
> As we currently dont have any automated tests for the opencloud-compose repo, should we keep the test in here, and rename the path?

nevermind, i see it isn't even in use now.

flimmy:
@pbleser-oc FYI

PR #1548, opened 2025-09-23
adr: use eduation api for multi-tenancy provisioning

rhafer:
@butonic @micbar This is trying to summarize what we discussed last week around the provisioning API for multi-tenancy.

Please comment.

### DIFF
--- a/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
+++ b/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
@@ -64,6 +64,9 @@ Use the existing Education API of the Graph Service.
 * Allows integration with existing (de)provisioning systems
 * hopefully keeps the implementation effort low
 
+Note: For now this means that the auto-provisioning feature will not be available for
+multi-tenant setups. We might want to revisit this in the future.
+
 ### Implementation Steps
 
 * re-vive the existing Education API implementation and run it as a separate service
--- a/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
+++ b/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
@@ -2,7 +2,7 @@
 title: "Use the graph education API for multi-tenant user provisioning"
 ---
 
-* Status: proposed
+* Status: approved
 * Deciders: [@micbar, @butonic, @rhafer]
 * Date: 2025-09-23
 
--- a/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
+++ b/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
@@ -1,0 +1,76 @@
+---
+title: "Use the graph education API for multi-tenant user provisioning"
+---
+
+* Status: proposed
+* Deciders: [@micbar, @butonic, @rhafer]
+* Date: 2025-09-23
+
+Reference: https://github.com/opencloud-eu/opencloud/issues/877
+
+## Context and Problem Statement
+
+With the current multi-tenancy implementation, the user-management is mostly external
+to the OpenCloud instance. Up to [now](../0001-simple-multi-tenancy-using-a-single-opencloud-instance.md)
+we relied on some external LDAP server providing the users including their tenant assignment.
+We'd like multi-tenancy to also work in environments where no such LDAP server is not available.
+
+## Decision Drivers
+
+* Mulit-tenancy must work without some existing external (as in not managed by us) LDAP server
+* keep the implementation effort low
+* allow integration with existing (de)provisioning systems
+
+## Considered Options
+
+### Use the auto-provisioning feature of OpenCloud
+
+We already have basic auto-provsioning features implemented in OpenCloud.
+Currently this is not tenant-aware, but it could be extended to support that.
+This would require some changes in the way that the users are managed by the
+auto-proviosioning code.
+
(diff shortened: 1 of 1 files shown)


docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+* Date: 2025-09-23
+
+Reference: https://github.com/opencloud-eu/opencloud/issues/877
+
+## Context and Problem Statement
+
+With the current multi-tenancy implementation, the user-management is mostly external
+to the OpenCloud instance. Up to [now](../0001-simple-multi-tenancy-using-a-single-opencloud-instance.md)
+we relied on some external LDAP server providing the users including their tenant assignment.
+We'd like multi-tenancy to also work in environments where no such LDAP server is not available.
```
butonic:
```suggestion
We'd like multi-tenancy to also work in environments where no such LDAP server is available.
```

docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+### Use the auto-provisioning feature of OpenCloud
+
+We already have basic auto-provsioning features implemented in OpenCloud.
+Currently this is not tenant-aware, but it could be extended to support that.
+This would require some changes in the way that the users are managed by the
+auto-proviosioning code.
+
+The auto-provisioning code does currently use the "normal" graph API to create
+users. That API is not tenant-aware and would need to be significantly changed
+to support multi-tenancy. However currently there is not real need to put
```
butonic:
```suggestion
to support multi-tenancy. However currently there is no real need to put
```

docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+tenant-awareness into that API (and it would drive us even further a away from
+compatibility with the MS Graph API). We could also switch away from the Graph API
+for auto-provisioning and use some direct calls to the underlying LDAP server.
+
+Also using the auto-provisioning feature means that user are only created
```
butonic:
```suggestion
Also, using the auto-provisioning feature means that users are only created
```

docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+when they first login. This means it is not possible to share files with users that
+have not yet logged in. This is a significant limitation.
+
+Also we don't currently have any de-provisioning features implemented.
+
+### Use the existing Eudcation API of the Graph Service
+
+We already implemented the Graph Education API in OpenCloud (based on the MS Graph Education API).
+This, appart from the somewhat different naming, does already bring most of what is needed
```
butonic:
```suggestion
This, apart from the somewhat different naming, does already bring most of what is needed
```

docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Implementation Steps
+
+* re-vive the existing Education API implementation and run it as a separate service
+* (maybe) allow to create tenants with a customer specified ID. The tenant id might also be
+  part of the user's claims (provided by the customer's identity provider). It would be better
+  if the tenant ids in our system match the tenant ids in the customer's identity provider.
+* For de-provisioning to work we need to implement a way to lookup users by an external ID as
+  that is only unique identfier the customer's system knows for a user. While the MS Graph API
+  already provides an `externalId` Attribtue we don't currently support that on our APIs.
```
butonic:
```suggestion
  already provides an `externalId` Attribute we don't currently support that on our APIs.
```

rhafer:
> I think this summarizes our discussion well. Should we clarify wether or not the autoprovisioning mode is to be used in combination with external provisioning? We should be provisioning the same user data if the claims contain all necessary properties.

I wrote this under the assumption that auto-provisioning would be turned off in this scenario. All user changes would need to be provisioned before the first login (and changes would also be done explicitly via the provisioning API). If that wasn't clear from the text, the I guess I need to add that. OTOH if we come to the conclusion that we need auto-provisioning to work in combination with the external provisioning we should write that down here as well.

>  It might be worth a dedicted ADR to discuss if we want to support only (auto)provisioning and drop the external LDAP integration (I'd love that).

Yeah. But let's keep that topic for another time.

docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md
```diff
@@ -0,0 +1,76 @@
...
+## Context and Problem Statement
+
+With the current multi-tenancy implementation, the user-management is mostly external
+to the OpenCloud instance. Up to [now](../0001-simple-multi-tenancy-using-a-single-opencloud-instance.md)
+we relied on some external LDAP server providing the users including their tenant assignment.
+We'd like multi-tenancy to also work in environments where no such LDAP server is not available.
+
+## Decision Drivers
+
+* Mulit-tenancy must work without some existing external (as in not managed by us) LDAP server
```
micbar:
```suggestion
* Multi-tenancy must work without some existing external (as in not managed by us) LDAP server
```

micbar (approved):

PR #1553, opened 2025-09-24
[full-ci] getting personal space by userId instead of userName

ScharfViktor:
- Removed `getUserIdByUserName()` → replaced with `getAttributeOfCreatedUser($user, 'id')`, which reduces the number of requests to `/graph/v1.0/users`.

- Added `getPersonalSpace()`, which retrieves a space by `userId` (instead of by name) via `/graph/v1.0/users/{userId}?%24select=&%24expand=drive`, ensuring the correct personal space is found even in case of duplicates (e.g., soft-deleted users).

### DIFF
--- a/tests/acceptance/bootstrap/GraphContext.php
+++ b/tests/acceptance/bootstrap/GraphContext.php
@@ -353,7 +353,7 @@ class GraphContext implements Context {
 	 * @throws GuzzleException
 	 */
 	public function theUserDeletesAUserUsingTheGraphAPI(string $byUser, string $user): void {
-		$userId = $this->featureContext->getUserIdByUserName($user);
+		$userId = $this->featureContext->getAttributeOfCreatedUser($user, 'id');
 		$this->featureContext->setResponse($this->deleteUserByUserIdUsingTheGraphApi($userId, $byUser));
 	}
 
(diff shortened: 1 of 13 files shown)


rhafer (approved):
I am by no means a PHP coder, so I didn't really review the code style an stuff. But the way you get the personal space now looks fine to me.

tests/acceptance/bootstrap/GraphContext.php
```diff
@@ -2455,7 +2454,7 @@ public function userTriesToExportGdprReportOfAnotherUserUsingGraphApi(
 	 */
 	public function getAssignedRole(string $user): ResponseInterface {
 		$userId = $this->featureContext->getAttributeOfCreatedUser($user, 'id')
-		?: $this->featureContext->getUserIdByUserName($user);
+		?: $this->featureContext->getAttributeOfCreatedUser($user, 'id');
```
individual-it: looks like we deliberately had both ways here, first try  `getAttributeOfCreatedUser` and if that fails use `getUserIdByUserName`.
but if there is no case where the user exists in the system but not in `$createdUsers[]` or `$createdRemoteUsers[]`, the line should be
only: `$userId = $this->featureContext->getAttributeOfCreatedUser($user, 'id');`

v-scharf: I deleted if->else
If a user was created (given step or during the test) ->  it is placed in $createdUsers[]. I took care of that here https://github.com/opencloud-eu/opencloud/pull/1553/files/078d40916d16495ba9ca3fbd68215be15052abf6#diff-cd22268dcad22d1ff7738bd89ac4da8543b277eabaf0bf4cadbdb56c5a8023d8R678-R692

individual-it, line 2457:
```suggestion
$userId = $this->featureContext->getAttributeOfCreatedUser($user, 'id');
```

individual-it (changes requested):

PR #1554, opened 2025-09-24
Introduce a multi-tenancy enablement configuration switch

### LINKED ISSUE #1458: Introduce a multi-tenancy enablement configuration switch
Currently there is not explict config setting to switch on multi-tenancy. We should consider adding one, to:

- make sure the graph service only start when the `cs3` identity backend it selected
- reject users without a tenantid
- ...

rhafer:
Closes #1458

For reviewing it's probably easier to look into the individual commits.

~~requires: https://github.com/opencloud-eu/reva/pull/360~~


dragonchaser (approved):
lgtm

PR #1562, opened 2025-09-25
attempt to fix broken soft delete

dragonchaser:


rhafer:
@dragonchaser Is this still needed?

dragonchaser:
> @dragonchaser Is this still needed?

nope

PR #1574, opened 2025-09-26
[full-ci] Enhance getSpaceByName: check local cache before Graph API calls

ScharfViktor:
- Added `getCreatedSpace()` method to fetch spaces from locally saved cache
- Updated `getSpaceByName()` to:
  - Check local cache first
  - Fetch personal space via Graph API if not in cache
  - Fall back to listing all available spaces via Graph API for users/admins
- Reduced unnecessary API calls and improved performance
- Added inline comments explaining the logic


tests/acceptance/bootstrap/SpacesContext.php
```diff
@@ -181,9 +182,19 @@ public function getSpaceByNameFromResponse(string $name, ?ResponseInterface $res
...
+			return $space;
+		}
+		// If it's a personal space and not found in the getCreatedSpace() — fetch via Graph API
+		// GET /graph/v1.0/users/{userId}?$select=&$expand=drive
+		if ($spaceName === "Personal" && $user !== 'admin') {
+			return $this->getPersonalSpace($user);
 		}
+		// If the space is not found in the getCreatedSpace() and it's not personal — call API to list available spaces
+		// Admin: GET /graph/v1.0/drives
+		// Regular user: GET /graph/v1.0/users/{userId}/drives
```
rhafer: There is no  `/graph/v1.0/users/{userId}/drives`  endpoint in the graph API. Are you really using that anywhere? AFAICS `listAllAvailableSpacesOfUser` is using `/graph/v1.0/me/drives`. So this might just be a bug in the comment.

tests/acceptance/bootstrap/SpacesContext.php
```diff
@@ -181,9 +182,19 @@ public function getSpaceByNameFromResponse(string $name, ?ResponseInterface $res
...
 	public function getSpaceByName(string $user, string $spaceName): array {
-		if ($spaceName === "Personal") {
-			$spaceName = $this->featureContext->getUserDisplayName($user);
+		// First, try to find the space in locally saved (previously created) spaces — no API call
+		$space = $this->getCreatedSpace($spaceName);
+		if ($space !== null) {
+			return $space;
+		}
+		// If it's a personal space and not found in the getCreatedSpace() — fetch via Graph API
+		// GET /graph/v1.0/users/{userId}?$select=&$expand=drive
```
rhafer, line 191: If you really just need the personal space of a specific user, `getPersonalSpace` you could use `/graph/v1.0/me/drive` (as the user) or `graph/v1.0/users/{userid}/drive` (with should work for admin and "normal" users). Might be a bit simpler, but feel free to ignore.

tests/acceptance/bootstrap/SpacesContext.php
v-scharf: no, sorry. it's bug in comment. should be `/graph/v1.0/me/drives`

tests/acceptance/bootstrap/SpacesContext.php
v-scharf: `graph/v1.0/users/{userid}/drive` doesn't work to me/ so I leave `/graph/v1.0/users/{userId}?$select=&$expand=drive` which is already exist and work

rhafer (approved):

PR #1582, opened 2025-09-29
change: use 404 as status when thumbnail can not be fetched

micbar:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


services/thumbnails/pkg/service/grpc/v0/service.go
```diff
@@ -160,7 +160,7 @@ func (g Thumbnail) handleCS3Source(ctx context.Context, req *thumbnailssvc.GetTh
 	case errors.Is(err, terrors.ErrImageTooLarge):
 		return "", merrors.Forbidden(g.serviceID, "%s", err.Error())
 	case err != nil:
-		return "", merrors.InternalServerError(g.serviceID, "could not get image from source: %s", err.Error())
+		return "", merrors.NotFound(g.serviceID, "could not get image from source: %s", err.Error())
```
rhafer: This has the side-effect of increasing the loglevel in the logging middleware/decorator from `Warn` to `Debug` for the errors that happen here. This means we would lose probably helpful log output in case of errors.

Should we adapt https://github.com/opencloud-eu/opencloud/blob/main/services/thumbnails/pkg/service/grpc/v0/decorators/logging.go#L39 accordingly?

butonic: Hm, for the thumbail service we should keep the noise low. If there is a problem users just do not see thumbnails. I think it is safe to be quiet about this by default. If there really is a problem, admins can set the log level for the thumbnailer to debug.

Issue #1475, opened 2025-09-08
We can't add default CSP rules reliably


kulmann:
## Describe the bug

When I want to add a new CSP rule to the set of [default CSP rules](https://github.com/opencloud-eu/opencloud/blob/main/services/proxy/pkg/config/csp.yaml), I can't be certain that it reaches all OpenCloud installations.

## Steps to reproduce

1. Release OpenCloud
2. Wait until admins have set up their own OpenCloud instances with customized `csp.yaml` files
3. Release another version of OpenCloud which needs a new CSP rule (e.g. our upcoming download-server so that we can check in the web ui, whether the installed version is up to date).

## Expected behavior

When I add a new CSP rule to the default csp.yaml it gets merged into the custom csp.yaml of all the existing OpenCloud instances which have their own custom csp.yaml.

## Actual behavior

If an OpenCloud instance already has a custom csp.yaml it will always overrule the default csp.yaml - as in: fully replace. There is no merging.

## Additional context

Kind of relates to this feature request: https://github.com/opencloud-eu/opencloud/issues/1388 where we want to be able to add CSP rules via web ui apps.

kulmann:
Please note that this is P2 because we want to bring an update check to the web ui, which will require an additional CSP rule for contacting the download server from the web ui. We can only do that for the default, not for OpenCloud instances which deviated from the default.

tbsbdr:
blocks https://github.com/opencloud-eu/web/issues/1158

dragonchaser:
@tbsbdr && @kulmann what csp.yaml should have higher priority? The custom user generated one or the one distributed by us (in case of overlapping configurations)?

kulmann:
@dragonchaser I would always say that user generated config takes precedence over our defaults. But: I think it's not possible to have conflicts, right? if both our defaults and the user generated csp.yaml have an entry that is exactly the same we only need it once (avoid the duplication). But everything else needs to be merged. Do you have examples for "conflicts" in mind where this is not true?

dragonchaser:
@kulmann that is exactly my point, example:

we have a csp config for service xzy, user has a csp config for service xzy, I'd always pick the users config, assuming the user is the source of truth and throw away our config

kulmann:
> I'd always pick the users config [...] and throw away our config

[...]

PR #1617, opened 2025-10-06
load two yaml configs

dragonchaser:
We have added the ability to merge custom CSP rules configuration with the provided ones through `PROXY_CSP_CONFIG_FILE_LOCATION` (and its yaml equivalent `csp_config_file_location`) and the ability to completely override the CSP rules configuration through `PROXY_CSP_CONFIG_FILE_OVERRIDE_LOCATION` (and its yaml equivalent `csp_config_file_override_location`)

refs #1475

### DIFF
--- a/services/proxy/pkg/middleware/security.go
+++ b/services/proxy/pkg/middleware/security.go
@@ -19,11 +19,7 @@ func LoadCSPConfig(proxyCfg *config.Config) (*config.CSP, error) {
 	if err != nil {
 		return nil, err
 	}
-	if proxyCfg.CSPConfigFileOverrideLocation == "" {
-		return loadCSPConfig(yamlContent, customYamlContent)
-	} else {
-		return loadCSPConfig(yamlContent, []byte{})
-	}
+	return loadCSPConfig(yamlContent, customYamlContent)
 }
 
 // LoadCSPConfig loads CSP header configuration from a yaml file.


butonic:
nice!  hm ... I'm nut sure this produces what we expect, when an admin has currently provided a csp rule file where he dropped eg the rules to github.  would we not then merge them back because they are in our default? ... I guess that is fine, because he can now overrule them ... but we need to document this propertly ... as it might cause unwanted security sideeffects.

I am aware of https://github.com/opencloud-eu/opencloud/issues/1475#issuecomment-3502269154 ... but why don't we introduce a PROXY_CSP_CONFIG_CUSTOMIZATIONS_FILE that allows adding additional rules. If someone really wants to get rid of the rules we provide he can point PROXY_CSP_CONFIG_FILE_LOCATION to an empty file and use his PROXY_CSP_CONFIG_CUSTOMIZATIONS_FILE. Of course he will then have to deal with updates himself.

I find that clearer than an PROXY_CSP_CONFIG_FILE_OVERRIDE_LOCATION. 

hm ... I'm not against this ... I am just worried about admins screaming "security incident!!!"

services/proxy/pkg/middleware/security.go
```diff
@@ -3,30 +3,52 @@ package middleware
...
-	yamlContent, err := loadCSPYaml(proxyCfg)
+	presetYamlContent, customYamlContent, err := loadCSPYaml(proxyCfg)
 	if err != nil {
 		return nil, err
 	}
-	return loadCSPConfig(yamlContent)
+	if proxyCfg.CSPConfigFileOverrideLocation == "" {
+		return loadCSPConfig(presetYamlContent, customYamlContent)
+	} else {
+		return loadCSPConfig(presetYamlContent, []byte{})
```
kulmann: This line should only return the yaml content as loaded from the override path.

services/proxy/pkg/config/config.go
```diff
@@ -24,27 +24,28 @@ type Config struct {
...
+	InsecureBackends              bool                `yaml:"insecure_backends" env:"PROXY_INSECURE_BACKENDS" desc:"Disable TLS certificate validation for all HTTP backend connections." introductionVersion:"1.0.0"`
+	BackendHTTPSCACert            string              `yaml:"backend_https_cacert" env:"PROXY_HTTPS_CACERT" desc:"Path/File for the root CA certificate used to validate the server’s TLS certificate for https enabled backend services." introductionVersion:"1.0.0"`
+	AuthMiddleware                AuthMiddleware      `yaml:"auth_middleware"`
+	PoliciesMiddleware            PoliciesMiddleware  `yaml:"policies_middleware"`
+	CSPConfigFileLocation         string              `yaml:"csp_config_file_location" env:"PROXY_CSP_CONFIG_FILE_LOCATION" desc:"The location of the CSP configuration file." introductionVersion:"1.0.0"`
+	CSPConfigFileOverrideLocation string              `yaml:"csp_config_file_override_location" env:"PROXY_CSP_CONFIG_FILE_OVERRIDE_LOCATION" desc:"The location of the CSP configuration file override." introductionVersion:"%%NEXT%%"`
```
kulmann, line 47: Except for an empty string check this var is completely unused. I would expect that
- either you work with three different yaml contents (default, custom, customOverride),
- or that the override var is a boolean and only makes sure that the default yaml is ignored entirely

I'd be fine with both. The behaviour described in the issue is the first behaviour. For that you would need to load the override yaml content and make use of it in `services/proxy/pkg/middleware/security.go` in line 25 (replace `presetYamlContent` with `customOverrideYamlContent`)

kulmann:
> I find that clearer than an PROXY_CSP_CONFIG_FILE_OVERRIDE_LOCATION.
> 
> hm ... I'm not against this ... I am just worried about admins screaming "security incident!!!"

... but it would still leave the update check broken for an unknown number of existing instances. Which is the point of my bug report and the main reason for this PR.

But I agree that we need clear documentation about all of this.

services/proxy/pkg/middleware/security.go
dragonchaser: I have renamed the variable and commited the missing part: https://github.com/dragonchaser/opencloud/blob/2241efc9d0c81f03ec2d23f5b09aee241688f473/services/proxy/pkg/middleware/security.go#L129-L132

services/proxy/pkg/config/config.go
dragonchaser: I have opted for using 3 files

services/proxy/pkg/middleware/security.go
```diff
@@ -15,14 +15,14 @@ import (
 
 // LoadCSPConfig loads CSP header configuration from a yaml file.
 func LoadCSPConfig(proxyCfg *config.Config) (*config.CSP, error) {
-	presetYamlContent, customYamlContent, err := loadCSPYaml(proxyCfg)
+	yamlContent, customYamlContent, err := loadCSPYaml(proxyCfg)
 	if err != nil {
 		return nil, err
 	}
 	if proxyCfg.CSPConfigFileOverrideLocation == "" {
```
kulmann: we don't need the if/else branching. we only need line 23. `loadCSPYaml` always returns the correct content for merging.

dragonchaser: correct

PR #1625, opened 2025-10-07
fix(settings): env var precedence

rhafer:
Fix the ordering of the env vars for the 'set_default_assignments' setting. The service specific variable ('SETTINGS_SETUP_DEFAULT_ASSIGNMENTS') should take precedence over other variable ('IDM_CREATE_DEMO_USERS' in this case).

Alternative fix to: https://github.com/opencloud-eu/opencloud-compose/pull/109

### DIFF
--- a/services/settings/pkg/config/config.go
+++ b/services/settings/pkg/config/config.go
@@ -33,7 +33,7 @@ type Config struct {
 
 	TokenManager *TokenManager `yaml:"token_manager"`
 
-	SetupDefaultAssignments bool `yaml:"set_default_assignments" env:"SETTINGS_SETUP_DEFAULT_ASSIGNMENTS;IDM_CREATE_DEMO_USERS" desc:"The default role assignments the demo users should be setup." introductionVersion:"1.0.0"`
+	SetupDefaultAssignments bool `yaml:"set_default_assignments" env:"IDM_CREATE_DEMO_USERS;SETTINGS_SETUP_DEFAULT_ASSIGNMENTS" desc:"The default role assignments the demo users should be setup." introductionVersion:"1.0.0"`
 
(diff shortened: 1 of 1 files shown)


schweigisito (approved):
I verified your fix by building a "dev" image based on latest `opencloud/main` including your change and running

```
$ export OC_DOCKER_IMAGE="opencloudeu/opencloud"; export OC_DOCKER_TAG="dev"; docker compose --env-file .env.1625 down -v && docker compose --env-file .env.1625 up -d
```

My `.env.1625` file used

```
grep -v ^# .env.1625|grep -v ^$|sort|egrep 'DEMO|COMPOSE_FILE'
COMPOSE_FILE=docker-compose.yml:traefik/opencloud.yml:idm/ldap-keycloak.yml:traefik/ldap-keycloak.yml
DEMO_USERS=true
```

BTW, I was not able to build a running dev image from your forked 'rhafer:env-var-precedence` branch.

rhafer:
>  I was not able to build a running dev image from your forked 'rhafer:env-var-precedence` branch.

How does it  fail? It seems to work for me.

schweigisito:
> How does it fail? It seems to work for me.

Must be something on my side probably.

If I use your branch, and build a dev image, using `make -C opencloud dev-docker` and try to deploy this with our compose and connect to it via browser, it just get a 404.

If I do exactly the same on the opencloud-eu/opencloud:main (applying your single change), all works as expected.

But, if I compare your branch with the opencloud:main, I exactly have this single change. I don't understand why, yet :-/

rhafer:
> If I use your branch, and build a dev image, using `make -C opencloud dev-docker` and try to deploy this with our compose and connect to it via browser, it just get a 404.

Try running `make generate` after you switch branches, before running `make -C opencloud dev-docker`. That might fix it.

schweigisito:
Yes, this worked! Thanks :)

PR #1639, opened 2025-10-09
[full-ci] deleted unused step definitions

ScharfViktor:
These are unused `Given` steps because we replaced all to sharing NG but forgot to drop them


v-scharf:
`api-bacis` `core-api-3` are green. Other pipelines failed due to a CI issue.

PR #1653, opened 2025-10-13
feat(collaboration): add cssVariables to Collabora open-in-app request

JammingBen:
The idea is to send the css variables as part of `form_parameters` in the `/app/open` request. This way, the web client can use them to style Collabora. This is specified in the Collabora docs for theming: https://sdk.collaboraonline.com/docs/theming.html#content-of-hidden-field-css-variables

Needs https://github.com/opencloud-eu/reva/pull/377 to work.

refs https://github.com/opencloud-eu/web/issues/516


JammingBen:
Turns out, we don't need to roundtrip via the server here -> closing.

PR #1658, opened 2025-10-14
graph: Make the nats dependency optional, fix readyness check

rhafer:
It should be possible to run the graph service without NATs. Standalone for the Provisioning API.

Also the LDAP dependency is optional in some cases (GRAPH_IDENTITY_BACKEND=cs3)


rhafer:
I am force merging this now despite the failing ci run. The previous run was green apart from the unit test, which have succeeded now. All the test that are failing now succeeded in the previous run. The current failures (e.g. https://ci.opencloud.eu/repos/3/pipeline/660/232) are in no way related to the code changes.

micbar:
The daily image has been built. I tested it. It works fine.

Issue #1523, opened 2025-09-19
Data loss from a restored folder


tammi-23:
## Describe the bug
It is possible to create two folders with the same name. This leads to a conflict and to strange behavior. 

## Steps to reproduce
1. Create a Folder A with a File B
2. First Delete the File B
3. Delete Folder A
4. Restore File B from Trashbin (here the Folder from 3. is recreated)
5. Restore Folder A (from Trashbin) and choose "keep both"
6. Click on the Folder with 0 kb (there is the File inside)
7. Go back and click on both folders (they both has the file inside)
8. Rename the empty folder (now one of both folders is not usable after the user is navigating into the folder)

## Expected behavior
Both folder are usable or the 2nd folder is created with an other name and there are no conflicts

## Actual behavior
Inside one of both folder it is not possible to create files (no new and no upload available)

"Resource not found
We went looking everywhere, but were unable to find the selected resource.
[Go to »Personal« page](https://demo.opencloud.eu/files/spaces/personal)" is displayed

<img width="987" height="719" alt="Image" src="https://github.com/user-attachments/assets/63a527a2-5683-4cf3-8e66-ee23ff40b531" />

v-scharf:
interesting, web has correct destination with prefix `(1)` `https://host.docker.internal:9200/remote.php/dav/spaces/27e7e0f6-cd60-4ad8-9684-223e0eb7c39f%24c0f63a6e-db39-4156-979b-cbf69af40473/new%20(1)`
we need a get second folder `a (1)`

v-scharf:
tested on demo instance https://demo.opencloud.eu/ `OpenCloud 3.5.0 OpenCloud Web UI 4.0.0`

Steps: 

- create `New folder` with file
- delete `New folder`
- create new empty `New folder`
- restore old `New folder` from trash using "keep both"

Actual: 2 `New folder` folders,  both are empty. No way to restore file which was inside

Expected: user has two folders: - new empty `New folder` and restored folder with file `New folder (1)`

fschade:
bug is posix fs related, decomposed (`export STORAGE_USERS_DRIVER=decomposed`) works as designed!

posix fs has another critical error case (which is similar):


## Steps

- Create a Folder A with a File B, and File C
- First Delete the File B
- Delete Folder A
- Restore File B from Trashbin (here the Folder from 3. is recreated)
- Restore Folder A (from Trashbin) and choose "keep both"

##  Result

FILE C is gone

[...]

PR #1660, opened 2025-10-14
apiTest-coverage for #1523

ScharfViktor:
related #1523

but tests are green without fixing. see explanation here https://github.com/opencloud-eu/opencloud/issues/1523#issuecomment-3401278660


v-scharf:
need reva bump with https://github.com/opencloud-eu/reva/pull/381

fschade (approved):

Issue #1597, opened 2025-10-02
Re-vive the existing Education API


db-ot:
# Userstory

As a developer, I want the existing Education API implementation to be revived and run as a separate service so that it can be maintained and scaled independently.

## Description

The current Education API implementation should be reactivated and deployed as its own standalone service. This separation improves maintainability, allows independent scaling, and prepares the service for future extensions.

## Acceptance Criteria

- Existing Education API implementation is restored and functional
- Education API runs as a separate service
- Documentation updated on how to run and access the service
- Create docker compose ( OC DEV Tools )

### Information Details

here are information about API and provisioning etc.

https://github.com/opencloud-eu/opencloud/blob/main/docs/adr/0002-use-education-api-for-multitenant-user-provisioning.md

In order to make the API easier to understand we should probably make a few attributes optional that only have a meaning in the School Context:

- Schools/Tenants: We likely won't need the `schoolNumber` attribute, which in our current implemenation is `required` . In MS Graph itself it optional.
- EducationUsers: 
  - The `primaryRole` attribute is currently required, we don't need that for the MultiTenant use case.
  - The `identities` attribute is supposed to contain the `sub` and `issuer` claim of the user. We/The provider might not even know those when users are provisioned. Those should be optional as well. (At least when an `externalId` (https://github.com/opencloud-eu/opencloud/issues/1599) is set in the request)

micbar:
https://github.com/opencloud-eu/opencloud-charts/pull/69#issue-3410757290

rhafer:
@dragonchaser  @fschade 

Here's a PR with a compose setup deploying OpenCloud, keycloak, OpenLDAP and the provisioning API for testing and development purposes:

https://github.com/opencloud-eu/opencloud/pull/1637

PR #1663, opened 2025-10-14
multi-tenancy: Optional attributes on provision API

rhafer:
For the provisioning API we want the `schoolNumber` attribute to be optional for a new School/Tenant. Also the user attributes `primaryRole` and `identities` are made optional. See #1597

### DIFF
--- a/services/graph/pkg/service/v0/educationuser.go
+++ b/services/graph/pkg/service/v0/educationuser.go
@@ -76,18 +76,7 @@ func (g Graph) PostEducationUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	identities, ok := u.GetIdentitiesOk()
-	if !ok {
-		logger.Debug().Err(err).Interface("user", u).Msg("could not create education user: missing required Collection: 'identities'")
-		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "missing required Attribute: 'identities'")
-		return
-	}
-	if len(identities) < 1 {
-		logger.Debug().Err(err).Interface("user", u).Msg("could not create education user: missing entry in Collection: 'identities'")
-		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "missing required Collection: 'identities'")
-		return
-	}
-	for i, identity := range identities {
+	for i, identity := range u.GetIdentities() {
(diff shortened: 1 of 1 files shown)


services/graph/pkg/identity/ldap_education_school.go
```diff
@@ -723,18 +727,22 @@ func (i *LDAP) createSchoolModelFromLDAP(e *ldap.Entry) *libregraph.EducationSch
 	if err != nil && !errors.Is(err, errNotSet) {
 		i.logger.Error().Err(err).Str("dn", e.DN).Msg("Error reading termination date for LDAP entry")
 	}
-	if id != "" && displayName != "" && schoolNumber != "" {
-		school := libregraph.NewEducationSchool()
-		school.SetDisplayName(displayName)
+
+	if id == "" || displayName == "" {
+		i.logger.Warn().Str("dn", e.DN).Str("id", id).Str("displayName", displayName).Msg("Invalid School. Missing required attribute")
```
micbar, line 732: Just out of curiosity, why do we use the Warn() log level?

rhafer: Good question. I didn't change that so it was added quite a while ago. I guess we could lower it to `Info` or even `Debug` because it's not really an error, but just caused by a user sending an invalid request.

Which one would you prefer?

services/graph/pkg/identity/ldap_education_school.go
```diff
@@ -119,16 +119,18 @@ func (i *LDAP) CreateEducationSchool(ctx context.Context, school libregraph.Educ
...
-	default:
-		logger.Error().Err(err).Str("schoolNumber", school.GetSchoolNumber()).Msg("error looking up school by number")
-		return nil, errorcode.New(errorcode.GeneralException, "error looking up school by number")
+	if school.HasSchoolNumber() {
+		_, err := i.getSchoolByNumber(school.GetSchoolNumber())
+		switch err {
+		case nil:
+			logger.Debug().Err(errSchoolNumberExists).Str("schoolNumber", school.GetSchoolNumber()).Msg("duplicate school number")
+			return nil, errSchoolNumberExists
+		case ErrNotFound:
```
fschade, line 128:
```go
errors.Is(err, ErrNotFound)
```

fschade (approved):

PR #1676, opened 2025-10-17
[full-ci] feat: implement OIDC authentication option

ScharfViktor:
### New Features
- **OIDC login Support**: Tests can now use bearer token authentication by setting `USE_BEARER_TOKEN=true` environment variable
- Tokens are cached per user to avoid regenerating on each request
- Expired tokens are refreshed using refresh tokens - need for admin (token TTL = 5min, after refresh token)

### Usage
Enable bearer token authentication in tests:
`USE_BEARER_TOKEN=true make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/..`

I decided to use the USE_BEARER_TOKEN variable first.
Basic Auth remains the default for now.
Once all tests work with OIDC, I’ll make it the default authentication method.

### Improvements

- During debugging, I noticed that after creating a user (we check 201), we additionally make a GET request to check for the user's existence, which I consider unnecessary. I deleted it
- reworked creating multiple users methods to using `HttpRequestHelper->sendRequest()`
- checking that share is syncing in some tests (were failed in CI)
- adjusted setup in Ci for multitenancy tests and updated description how to run tests localy https://github.com/opencloud-eu/opencloud/pull/1313#issue-3296964130

### DIFF
--- a/tests/acceptance/bootstrap/FeatureContext.php
+++ b/tests/acceptance/bootstrap/FeatureContext.php
@@ -52,7 +52,6 @@ use Swaggest\JsonSchema\Exception\NumericException;
 use Swaggest\JsonSchema\Exception\ObjectException;
 use Swaggest\JsonSchema\Exception\StringException;
 use Swaggest\JsonSchema\Exception\TypeException;
-use TestHelpers\TokenHelper;
 
 require_once 'bootstrap.php';
 
--- a/tests/acceptance/bootstrap/GraphContext.php
+++ b/tests/acceptance/bootstrap/GraphContext.php
@@ -11,7 +11,6 @@ use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\TableNode;
 use GuzzleHttp\Exception\GuzzleException;
-use PhpCsFixer\Doctrine\Annotation\Token;
 use PHPUnit\Framework\Assert;
 use Psr\Http\Message\ResponseInterface;
 use TestHelpers\GraphHelper;


tests/acceptance/TestHelpers/TokenHelper.php
```diff
@@ -0,0 +1,392 @@
...
+			[
+			'headers' => [
+				'Kopano-Konnect-XSRF' => '1',
+				'Referer' => $baseUrl,
+				'Content-Type' => 'application/json'
+			],
+			'json' => [
+				'params' => [$username, $password, '1'],
+				'hello' => [
+					'scope' => 'openid profile email',
```
rhafer: are you sure you get a refresh token back with this values for `scope` AFAIK lico will only return a refresh token when the `offline_access` scope is added to the scopes?

rhafer (commented):
Looks good to me. Just a question related to the refresh tokens and the requested scopes.

tests/acceptance/TestHelpers/TokenHelper.php
v-scharf: it was copy paste from `web`. I can fix it if you tell me how

but it works: 
I delete users after each test, except for `admin` -> admin updates their token every 5 minutes.
If `refresh token` didn't work, the tests would stop working after 5 minutes.

rhafer: That is really weird because lico really does not return a `refresh_token` here. When I run the steps manually I just get an `id_token` and an `access_token` back:

```
 curl 'https://cloud.collabora-test.opencloud.rocks/konnect/v1/token'   -X POST   -H 'Content-Type: application/x-www-form-urlencoded'    --data-raw 'grant_type=authorization_code&redirect_uri=https%3A%2F%2Fcloud.collabora-test.opencloud.rocks%2Foidc-callback.html&code=G-yipFrxJ1fpzlTf6HqLhfcWtr1otlOZ&client_id=web' 
{
  "access_token": "...",
  "token_type": "Bearer",
  "id_token": "...",
  "expires_in": 300
}
```
And if you look in the browser console when logging in, there is also no refresh_token returned.

I really wonder why the your code would work. Are you sure the server is actually returning a refresh_token for you. I'd be surprised if you do. You might be hitting some weird malfunction in lico.

v-scharf: There's some confusion here. I'll show you tomorrow beter.
`makeLoginRequest()` uses `/signin/v1/identifier/_/logon`
I get token here `getToken()` `/konnect/v1/token` and I get refresh token as well. see here

Screenshot 2025-10-23 at 20 24 30

saw-jan: The `getCode` method includes `offline_access` scope; therefore, the corresponding token response might have included refresh_token in it.

tests/acceptance/TestHelpers/TokenHelper.php
```diff
@@ -0,0 +1,392 @@
...
+	 * @param string $password
+	 * @param string $url
+	 *
+	 * @return array ['access_token' => string, 'refresh_token' => string, 'expires_at' => int]
+	 * @throws GuzzleException
+	 * @throws Exception
+	 */
+	public static function getTokens(string $username, string $password, string $url): array {
+		// Extract base URL. I need to send $url to get correct server in case of multiple servers (ocm suite)
+		$baseUrl = preg_replace('#(https?://[^/]+).*#', '$1', $url);
```
saw-jan: we can have a separate method to parse the baseUrl and use it wherever needed.

tests/acceptance/TestHelpers/TokenHelper.php
```diff
@@ -0,0 +1,392 @@
...
+	/**
+	 * Clear cached tokens for a specific user
+	 *
+	 * @param string $username
+	 * @param string $url
+	 *
+	 * @return void
+	 */
+	public static function clearUserTokens(string $username, string $url): void {
+		$baseUrl = preg_replace('#(https?://[^/]+).*#', '$1', $url);
```
saw-jan: same as above

tests/acceptance/TestHelpers/TokenHelper.php
rhafer: Argh. Thanks for the pointer. I overlooked that. @ScharfViktor could you please add `offline_access` here (makeLoginRequest) as well? Even if just for the sake of consistency.

v-scharf: done

tests/acceptance/bootstrap/FeatureContext.php
```diff
@@ -52,6 +52,7 @@
 use Swaggest\JsonSchema\Exception\ObjectException;
 use Swaggest\JsonSchema\Exception\StringException;
 use Swaggest\JsonSchema\Exception\TypeException;
+use TestHelpers\TokenHelper;
```
saw-jan: unused import?

tests/acceptance/bootstrap/GraphContext.php
```diff
@@ -11,12 +11,14 @@
 use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\TableNode;
 use GuzzleHttp\Exception\GuzzleException;
+use PhpCsFixer\Doctrine\Annotation\Token;
```
saw-jan: unused import?

tests/acceptance/bootstrap/FeatureContext.php
v-scharf: yes 👍 , I deleted it

PR #1724, opened 2025-10-28
fix: only search LDAP group by name

### LINKED ISSUE #1673: LDAP searches can return weird matches based on an ID match
## Describe the bug
LDAP group searches (performed when trying to add a new member to a space, for instance) do a substring search for both the `cn` and the `entryUUID`.  This leads to weird experiences where results are returned that do not match the query in an obvious way.

## Steps to reproduce
1. Have a group with an `entryUUID` that contains a string such as "beef" in it.  This might be difficult because these are generated by the LDAP provider.
2. Try to add "Beef Tobin" to a space

## Expected behavior
Only see an entry with a name that contains "beef" (a person named Beef Tobin)

## Actual behavior
See an entry that does not contain "beef" for the group that has the `entryUUID` with the substring "beef".

## Additional context
This came up when I discovered I couldn't add someone to a Space.  After some debugging, it turned out that my LDAP provider ([lldap](https://github.com/lldap/lldap), which isn't uncommon for self-hosters to run) did not support substring queries on `entryUUID`, and so the LDAP query failed.  I [filed an issue there](https://github.com/lldap/lldap/issues/1327) to add that support, but they rightly pushed back that the behavior is odd.

I think it's reasonable to match an an exact `entryUUID` hit in LDAP so power users/admins can get the exact group they want if they know it, but it can be jarring and produce unexpected results to end users since searches start with as few as three characters.

sdwilsh:
## Description
Searching an LDAP group by `entryUUID` does not make a lot of sense.  This change drops that query.

## Related Issue
- Fixes #1673

## Motivation and Context
As described in the related Issue, my LDAP backend, [LLDAP](https://github.com/lldap/lldap) does not support a substring query on `entryUUID`, and actually returns an error that completely breaks my ability to add anyone to a Space.

This is a breaking change if anybody actually depended on this behavior, but as @rhafer pointed out in the related Issue, other LDAP servers like OpenLDAP doesn't support this query either (but fail with no results), so it is dependent upon the LDAP server that is being used.

## How Has This Been Tested?
- `cd services/graph/pkg/identity && go test`

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/graph/pkg/identity/ldap_group.go
+++ b/services/graph/pkg/identity/ldap_group.go
@@ -83,9 +83,8 @@ func (i *LDAP) GetGroups(ctx context.Context, oreq *godata.GoDataRequest) ([]*li
 	if search != "" {
 		search = ldap.EscapeFilter(search)
 		groupFilter = fmt.Sprintf(
-			"(|(%s=*%s*)(%s=*%s*))",
+			"(%s=*%s*)",
 			i.groupAttributeMap.name, search,
-			i.groupAttributeMap.id, search,
 		)
 	}
 	groupFilter = fmt.Sprintf("(&%s(objectClass=%s)%s)", i.groupFilter, i.groupObjectClass, groupFilter)
--- a/services/graph/pkg/identity/ldap_group_test.go
+++ b/services/graph/pkg/identity/ldap_group_test.go
@@ -305,7 +305,7 @@ func TestGetGroupsSearch(t *testing.T) {
 	// only match if the filter contains the search term unquoted
 	lm.On("Search", mock.MatchedBy(
 		func(req *ldap.SearchRequest) bool {
-			return req.Filter == "(&(objectClass=groupOfNames)(|(cn=*term*)(entryUUID=*term*)))"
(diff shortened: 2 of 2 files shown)


micbar:
@rhafer FYI

rhafer:
@sdwilsh Thanks a lot for your contribution!

> This is a breaking change if anybody actually depended on this behavior, but as @rhafer pointed out in the related Issue, other LDAP servers like OpenLDAP doesn't support this query either (but fail with no results), so it is dependent upon the LDAP server that is being used.

Actually this is not a breaking change. According to the LDAP Standards the UUID Syntax (as used for any UUID valued attribute) does not have a substring matching rule defined. So any substring filter for an attribute of a UUID type evaluates to `undefined`. And for the concrete filter that means that it will be ignored because it is ORed with another filter. (at least on a standards conform LDAP server). I removed the `Breaking Change` marker.

Additionally adding the GroupId to the filter for the group search seems to have been a bug from the beginning. Did doesn't really makes senses.

PR #1732, opened 2025-10-28
make user cache tenant aware

### LINKED ISSUE #1462: User cache in graph service is not multi-tenant safe
## Describe the bug

The graph service maintains a simple in-memory cache for looking up users by userid (https://github.com/opencloud-eu/opencloud/blob/main/services/graph/pkg/identity/cache.go#L87), that cache does currently not take the tenant-ids of the requesting and the requested user into account.

So a users from one tenant might get some information about a user from a different tenant if it gets to know the other users userid and if the user entry is already cached (because a different user looked it up before).

(Similar issues would exist for the groups cache, once we enable groups for multi-tenancy)

dragonchaser:
This PR makes the user cache in the identity package of the graph service tenant aware
fixes #1462

### DIFF
--- a/services/graph/pkg/identity/cache/cache.go
+++ b/services/graph/pkg/identity/cache/cache.go
@@ -116,11 +116,6 @@ func (cache IdentityCache) GetCS3User(ctx context.Context, tenantId, userid stri
 			}
 			return nil, errorcode.New(errorcode.GeneralException, err.Error())
 		}
-		// check if the user is in the correct tenant
-		// if not we need to return before the cache is touched
-		if user.GetId().GetTenantId() != tenantId {
-			return nil, identity.ErrNotFound
-		}
 
 		cache.users.Set(tenantId+"|"+userid, user, ttlcache.DefaultTTL)
 	} else {
--- a/services/graph/pkg/identity/cache/cache.go
+++ b/services/graph/pkg/identity/cache/cache.go
@@ -107,6 +107,7 @@ func (cache IdentityCache) GetCS3User(ctx context.Context, tenantId, userid stri
 		}
 		cs3UserID := &cs3User.UserId{
 			OpaqueId: userid,
+			TenantId: tenantId,
 		}
 		user, err = revautils.GetUserNoGroups(ctx, cs3UserID, gatewayClient)
 		if err != nil {
@@ -123,9 +124,6 @@ func (cache IdentityCache) GetCS3User(ctx context.Context, tenantId, userid stri
 
 		cache.users.Set(tenantId+"|"+userid, user, ttlcache.DefaultTTL)
 	} else {
-		if item.Value().GetId().GetTenantId() != tenantId {
-			return nil, identity.ErrNotFound
-		}
 		user = item.Value()
 	}
 	return user, nil
--- a/services/graph/pkg/identity/cache/cache.go
+++ b/services/graph/pkg/identity/cache/cache.go
@@ -1,4 +1,4 @@
-package identity
+package cache
 
 import (
 	"context"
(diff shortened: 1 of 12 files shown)


services/graph/pkg/identity/cache/cache.go
```diff
@@ -106,13 +111,21 @@ func (cache IdentityCache) GetCS3User(ctx context.Context, userid string) (*cs3U
...
+				return nil, identity.ErrNotFound
 			}
 			return nil, errorcode.New(errorcode.GeneralException, err.Error())
 		}
-		cache.users.Set(userid, user, ttlcache.DefaultTTL)
+		// check if the user is in the correct tenant
+		// if not we need to return before the cache is touched
+		if user.GetId().GetTenantId() != tenantId {
+			return nil, identity.ErrNotFound
+		}
```
rhafer: `GetUserNoGroups()` is already tenant aware. You could just add the tenantId to the `c3UserID` above and `GetUserNoGroups()` would only return a user if the tenantid matches.

services/graph/pkg/identity/cache/cache.go
```diff
@@ -106,13 +111,21 @@ func (cache IdentityCache) GetCS3User(ctx context.Context, userid string) (*cs3U
...
 
+		cache.users.Set(tenantId+"|"+userid, user, ttlcache.DefaultTTL)
 	} else {
+		if item.Value().GetId().GetTenantId() != tenantId {
+			return nil, identity.ErrNotFound
```
rhafer: Why is this needed? The tenantid id is already part of the cache key. So any user returned should be from that tenant, no?

services/graph/pkg/service/v0/api_driveitem_permissions.go
```diff
@@ -184,7 +185,8 @@ func (s DriveItemPermissionsService) Invite(ctx context.Context, resourceId *sto
 		cTime = createShareResponse.GetShare().GetCtime()
 		expiration = createShareResponse.GetShare().GetExpiration()
 	default:
-		user, err := s.identityCache.GetCS3User(ctx, objectID)
+		// TODO: get tenantId from revactx.ContextGetUser(ctx), maybe we need to extent the user struct
```
rhafer: Shouldn't we address this before merging? Unless I am overlooking something the tenantId should already be part of the User returned by `ContextGetUser()` (inside the `UserId` member?

services/graph/pkg/service/v0/api_driveitem_permissions.go
```diff
@@ -259,14 +261,15 @@ func (s DriveItemPermissionsService) Invite(ctx context.Context, resourceId *sto
 	}
 
 	if user, ok := revactx.ContextGetUser(ctx); ok {
-		identity, err := userIdToIdentity(ctx, s.identityCache, user.GetId().GetOpaqueId())
+		// TODO: get tenantId from revactx.ContextGetUser(ctx), maybe we need to extent the user struct
+		userIdentity, err := userIdToIdentity(ctx, s.identityCache, "", user.GetId().GetOpaqueId())
```
rhafer: The tenantId should already be there. `user.GetId().GetTenantId()`

services/graph/pkg/service/v0/base.go
```diff
@@ -170,7 +171,8 @@ func (g BaseGraphService) cs3SpacePermissionsToLibreGraph(ctx context.Context, s
 			}
 			isGroup = true
 		} else {
-			cs3Identity, err = userIdToIdentity(ctx, g.identityCache, tmp)
+			// TODO: get tenantId from revactx.ContextGetUser(ctx), maybe we need to extent the user struct
```
rhafer: same here

rhafer (changes requested):
Looks fine so far. Just some thoughts regarding the remaining TODO comments and a small simplification.

rhafer (approved):

PR #1773, opened 2025-11-04
introduce read http timeout listener

butonic:
add a timeout listener that uses a sliding window for a read timeout.

currently uses a hardcoded timeout of 3sec. we should make that configurable


micbar:
@fschade @butonic What is the status of this PR?

It not needed, please close.

PR #1781, opened 2025-11-05
fix: set global signing secret fallback correctly

rhafer:
When falling back to the transfer secret we need to set the global cfg.URLSigningSecret as well otherwise the Validate() step fails.

### DIFF
--- a/pkg/config/parser/parse.go
+++ b/pkg/config/parser/parse.go
@@ -102,7 +102,7 @@ func EnsureCommons(cfg *config.Config) {
 
 	// make sure url signing secret is set and copy it to the commons part
 	// fall back to transfer secret for url signing secret to avoid
-	// issues when upgradin from an older release
+	// issues when upgrading from an older release
 	if cfg.URLSigningSecret == "" {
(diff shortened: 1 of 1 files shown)


pkg/config/parser/parse.go
```diff
@@ -100,14 +100,13 @@ func EnsureCommons(cfg *config.Config) {
...
-	// copy url signing secret to the commons part if set
-	if cfg.URLSigningSecret != "" {
-		cfg.Commons.URLSigningSecret = cfg.URLSigningSecret
-	} else {
-		// fall back to transfer secret for url signing secret to avoid
-		// issues when upgradin from an older release
-		cfg.Commons.URLSigningSecret = cfg.TransferSecret
+	// make sure url signing secret is set and copy it to the commons part
+	// fall back to transfer secret for url signing secret to avoid
+	// issues when upgradin from an older release
```
kulmann:
```suggestion
	// issues when upgrading from an older release
```

micbar (approved):

PR #1809, opened 2025-11-07
check propfing after renaming data in file system

ScharfViktor:
test for checking propfing response body after renaming file in the file system


rhafer:
rebased to re-run with latest changes

PR #1844, opened 2025-11-12
docs: Fix auth-app examples in README

rhafer:
We return UUIDs for the token not the actually token hash since at least OpenCloud 2.0.0

Related: #1832


rhafer:
Merging, ignoring the unrelated CI failure.

PR #1856, opened 2025-11-13
benchmark client enhancements

butonic:
this allow reading file content from disk as with curl, eg. to upload a blob as txt to test the virus scanner
```
go run ./opencloud/cmd/opencloud benchmark client  -X PUT -k --data-binary @/tmp/binary-blob -H 'Content-Type: application/octet-stream' -j 10 -u admin:admin 'https://localhost:9200/dav/spaces/storage-users-1$d0e6c8e8-deef-4a6f-945e-75aba6a2f5a8/test.txt'
```

### DIFF
--- a/opencloud/pkg/command/benchmark.go
+++ b/opencloud/pkg/command/benchmark.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"bytes"
+	"context"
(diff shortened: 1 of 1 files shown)


butonic:
cc @dragotin

PR #1873, opened 2025-11-18
autodecode base64 encoded uuid userid

butonic:
Keycloak base64 encodes the AD binary objectGUID. We can automatically detect that. No need for a new env var.


micbar:
@rhafer @butonic 

I added a test for that.

Not a beauty, but the mock User Backend now validates the given claim value during "GetUserByClaim"

rhafer:
@butonic @micbar Is there any documentation for this "feature" in Keycloak e.g. under which circumstances a value is considered "binary"?

butonic:
superseded by https://github.com/opencloud-eu/opencloud/pull/1901

Issue #1643, opened 2025-10-10
Request hangs when less bytes are send than advertised by `Content-Length` header


individual-it:
## Describe the bug
When the TUS upload request contains less bytes than given in the `Content-Length` header, the request hangs or comes back with 502 depending on the deployment

## Steps to reproduce
1. create a TUS resource e.g. `curl -vk -XPOST 'https://localhost:9200/remote.php/dav/spaces/6ee4ff97-5185-4a54-8df7-447299e75dfa$cb651e9f-eebb-4827-95eb-f3cea34bb6d4' -uadmin:admin -H"Tus-Resumable: 1.0.0" -H"Upload-Length: 10" -H"Upload-Metadata: filename ZmlsZS50eHQ=" -H"Content-Type: application/offset+octet-stream"`
2. send a PATCH request to the location given by the server, but set the `Content-Lenght` header to `10` but only send 9 bytes: `curl -vk -XPATCH "$LOCATION" -H'Content-Type: application/offset+octet-stream' -H'Content-Length: 10' -H'Upload-Offset: 0' -H'Tus-Resumable: 1.0.0' -d"012345678"`

## Expected behavior
4xx error

## Actual behavior
on bare metal installation the request just hangs
in the docker compose full example a 502 error comes back

log in the server:
```
{"level":"error","service":"storage-users","host.name":"deepthought3","pkg":"rhttp","datatx":"tus","method":"PATCH","path":"/8e04450a-3455-4738-96cb-c3d7cfcf6087","requestId":"","id":"8e04450a-3455-4738-96cb-c3d7cfcf6087","error":"ERR_READ_TIMEOUT: timeout while reading request body","time":"2025-10-10T12:49:39+05:45","message":"BodyReadError"}
{"level":"error","service":"storage-users","host.name":"deepthought3","pkg":"rhttp","traceid":"8cefab162ea7ca27b1fdfb68aae32f2f","host":"127.0.0.1","method":"PATCH","uri":"/data/tus/8e04450a-3455-4738-96cb-c3d7cfcf6087","url":"/8e04450a-3455-4738-96cb-c3d7cfcf6087","proto":"HTTP/1.1","status":500,"size":53,"start":"10/Oct/2025:12:48:39 +0545","end":"10/Oct/2025:12:49:39 +0545","time_ns":60001145254,"time":"2025-10-10T12:49:39+05:45","message":"http"}
[...]

PR #1883, opened 2025-11-20
enhancement(proxy): add read deadline and filter chain middleware

fschade:
## Description
as described [here](https://github.com/opencloud-eu/opencloud/issues/1643) openCloud does not close dangling uploads which can result in unnecessary unclosed connections.

Usually openCloud is used behind a proxy such as nginx which has timout settings for such cases, e.g.:

- proxy_connect_timeout
- proxy_send_timeout
- proxy_read_timeout
- send_timeout

which works and fixes that!

Anyway in my opinion openCloud should also have such options just in case.

As a starting-point the PR introduces such tools and already fixes it for the described upload case [here](https://github.com/opencloud-eu/opencloud/issues/1643).

The difference to the nginx timeout is that this implementation refreshes the timeout after each read of the request body which is specially helpful for large uploads with probably many chunks!

In a nutshell it contains the following tools:

- a middleware that allows to inspect and apply middleware chains per request 
- a middleware that sets a deadline for uploads and refreshes that after each read cycle

unfortunately the last uploaded chunk could take some time to finish because we have to calculate the checksum here and if the file is large the calculation will take some time, thats the reason behind the long upload deadline default value (120s).

as discussed today @butonic and @rhafer we maybe can send the checksum calculation to the background and not as part of the last upload chunk request! NOT PART OF THIS PR!

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/1643

## Motivation and Context
close unused or dangling requests

## How Has This Been Tested?
```bash
#!/usr/bin/env bash

echo 1
LOCATION=$(curl -k -I -s -o /dev/null -w '%header{location}' -XPOST 'https://localhost:9200/remote.php/dav/spaces/281457a5-5279-47d7-b55e-9f514c2df42b%24b6b9f660-e084-4dda-9689-76b75777381e' -uadmin:admin -H"Tus-Resumable: 1.0.0" -H"Upload-Length: 10" -H"Upload-Metadata: filename ZmlsZS50eHQ=" -H"Content-Type: application/offset+octet-stream")
echo 2
echo "$LOCATION"
echo "========================="
echo `date`: PATCH start
echo "========================="
curl -vk -XPATCH "$LOCATION" -H'Content-Type: application/offset+octet-stream' -H'Content-Length: 10' -H'Upload-Offset: 0' -H'Tus-Resumable: 1.0.0' -d"012345678"
echo "========================="
echo `date`: PATCH end
echo "========================="
echo 3
```

## Types of changes
- [X] Technical debt

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/proxy/pkg/command/server.go
+++ b/services/proxy/pkg/command/server.go
@@ -6,11 +6,23 @@ import (
 	"fmt"
 	"net/http"
 	"os/signal"
+	"slices"
 	"time"
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	chimiddleware "github.com/go-chi/chi/v5/middleware"
 	"github.com/justinas/alice"
+	"github.com/opencloud-eu/reva/v2/pkg/events"
+	"github.com/opencloud-eu/reva/v2/pkg/events/stream"
+	"github.com/opencloud-eu/reva/v2/pkg/rgrpc/todo/pool"
+	"github.com/opencloud-eu/reva/v2/pkg/signedurl"
+	"github.com/opencloud-eu/reva/v2/pkg/store"
+	"github.com/urfave/cli/v2"
+	"go-micro.dev/v4/selector"
+	microstore "go-micro.dev/v4/store"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/opencloud-eu/opencloud/pkg/config/configlog"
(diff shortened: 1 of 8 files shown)


micbar:
I have still the question if we need that. Isn't our message always "Do not run a go http server without a reverse proxy"?

fschade:
> I have still the question if we need that. Isn't our message always "Do not run a go http server without a reverse proxy"?

thats the question, if... we close this pr, if not... we should have some guard like that

PR #1900, opened 2025-11-24
chore: bump web to v4.2.1-rc.1

kulmann:
Bump web to v4.2.1-rc.1


ScharfViktor (approved):

.woodpecker.env
```diff
@@ -1,4 +1,4 @@
 # The test runner source for UI tests
-WEB_COMMITID=6abffcc9cff31c46a341105eb6030fec56338126
```
v-scharf, line 2: `fatal: reference is not a tree: 3d7367cfb1abe288d0fc0b0b1cc494a7747bcaf6` in the CI

something is wrong with commitid

ScharfViktor (changes requested):

.woodpecker.env
kulmann: good spot, thank you. the .woodpecker.env was still set to `main` instead of `stable-4.2`.

PR #1908, opened 2025-11-25
bump pkg/tracing semconf to 1.37.0

butonic:
this is the only place where we use this old version


butonic:
superseded by https://github.com/opencloud-eu/opencloud/pull/1918

PR #1927, opened 2025-11-27
[full-ci] revaBump-v2.40.1

ScharfViktor:
reva bump for rc.3

PR #1928, opened 2025-11-27
bump reva to 2.40.0

dragonchaser:
This PR bumps reva to 2.40.0


butonic (approved):

rhafer:
I think this can be closed. #1927 is more current and has already some test fixes.

PR #1954, opened 2025-12-02
migrate from urfave/cli to spf13/cobra

dragonchaser:
We have replaced urfave/cli by spf13/cobra, because urfave had issues with race conditions and we could not fix them.

**Hints for reviewers:**

- the PR is huge, but not that huge ("only" 5k lines of change, most of it duplicate refactorings, the rest ist vendoring from `spf13/cobra`, `spf13/viper` and `spf13/pflag` no business logic has been touched)

- displaying help of consumed service commands is tricky, I replace the `RunE` function in case there is none to make sure that the help is displayed, if the function already has a `RunE` function, it might be needed to add `cmd.Help()` there. (replace function: https://github.com/dragonchaser/opencloud/blob/22857e0cdc264701db0fc937b4ef2e595d3f8cd8/opencloud/pkg/command/root.go#L26-L36)

**TODO:**

- [X] fix https://github.com/dragonchaser/opencloud/blob/22857e0cdc264701db0fc937b4ef2e595d3f8cd8/opencloud/pkg/command/root.go#L26-L36 the nil check is somehow bogus, subcommand help is not displayed
- [X] fix `make test-with-race` ci hickups... there are actual race conditions
- [X] make ci-green
- [x] please verify that all `Required: true` in urfave/cli commands are migrated to cobra like:
```go
err := corbraCommandVariable.MarkFlagRequired("<flagname>")
if err != nil {
	fmt.Println(err)
}
```
- [x] fix Flag("...").Value calls and use type safe cobra substitution, `cmd.Flags().GetBool("..."), ...`
  i think I replaced all.... but just make sure

- trigger docs team to update docs, because https://github.com/dragonchaser/opencloud/blob/22857e0cdc264701db0fc937b4ef2e595d3f8cd8/opencloud/pkg/command/init.go#L44-L86 the shorthand commands have changed (background: cobra only supports single character shorthands, urfave multi-char shorthands)
- `oc init` command can not be interrupted with `ctrl+c` when running interactively without parameters
- make `cmd.Context()` in cobra inherit from the commands context to accept ctrl+c

### DIFF
--- a/tests/acceptance/features/apiAntivirus/antivirus.feature
+++ b/tests/acceptance/features/apiAntivirus/antivirus.feature
@@ -191,6 +191,7 @@ Feature: antivirus
     And user "Brian" should get a notification with subject "Virus found" and message:
       | message   |
       | <message> |
+    When the user waits for "10" seconds for postprocessing to finish
     And as "Brian" file "/Shares/uploadFolder/<new-file-name>" should not exist
     And as "Alice" file "/uploadFolder/<new-file-name>" should not exist
     Examples:


dragonchaser:
There is still an issue with the output of the subcommand usage from `opencloud/bin/opencloud` all commands are there, can be output with `opencloud/bin/opencloud sse --help` but running `opencloud/bin/opencloud sse` no usage is displayed.

**Update:** https://github.com/opencloud-eu/opencloud/pull/1954/commits/dde1a4334ebe0075bebcaec6f1bba71ed1b4accc fixes the issue

services/postprocessing/pkg/command/postprocessing.go
```diff
@@ -55,14 +39,16 @@ func RestartPostprocessing(cfg *config.Config) *cli.Command {
...
+			step := ""
 			if uid == "" {
-				step = c.String("step")
+				step, _ = cmd.Flags().GetString("step")
 			}
 
+			restart, _ := cmd.Flags().GetBool("restart")
 			var ev events.Unmarshaller
 			switch {
-			case c.Bool("retrigger"):
```
butonic, line 65: how did this ever work? the flag was registered as `restart`

fschade: w00t 👀

PR #1994, opened 2025-12-09
Groupware

pbleser-oc:
## Description
Ongoing implementation of the Groupware backend service, which initially happened on its own branch but should now occur on the `main` branch.

Changes are almost exclusively on packages that are distinct to the Groupware backend, namely
* `pkg/jmap`, `pkg/jscalendar`  and `pkg/jscontact`: contain the implementation of the [JMAP protocol](https://www.rfc-editor.org/rfc/rfc8620.html) and data models for Core, Mail, Contacts, Calendars, Blobs, ...
* `services/groupware`: contains the Groupware backend service which is currently configured to not be started by default, and must be explicitly included in the `START_ADDITIONAL_SERVICES` property

Changes to common areas include:
* `devtools/deployments/opencloud_full`: addition of a container of the [Stalwart](https://stalw.art/) mail server which is used for Groupware functionality
* `pkg/structs`: add more helper functions that are then used in `pkg/jmap` and `services/groupware`

Note that it is not fully functional yet and is going to be under continued and ongoing development along with its UI counterparts.

Specifically, the following aspects are lacking and only implemented in a skeletal way as a proof of concept:
* API documentation: originally developed using [go-swagger](https://github.com/go-swagger/go-swagger) and [redocly-cli](https://redocly.com/docs/cli), but which is having severe limitations and might be superseded by an [ongoing implementation](https://github.com/opencloud-eu/groupware-apidocs) of a custom source code parser which renders OpenAPI
* unit tests: very little coverage if at all, which is mostly due to most of the implementation of the Groupware backend doing relatively little logic and is mostly implementing the JMAP model
* integration tests: some coverage exists that performs tests against a Stalwart container using testcontainers, but will be made more exhaustive
* k6 black box tests: none implemented yet
* system documentation: none implemented yet, too early to do so

The Groupware backend is not meant to be used productively yet.


services/auth-api/pkg/service/http/v0/service.go
```diff
@@ -0,0 +1,266 @@
...
+			otelchi.WithChiRoutes(m),
+			otelchi.WithTracerProvider(options.TraceProvider),
+			otelchi.WithPropagators(tracing.GetPropagator()),
+			otelchi.WithTraceResponseHeaders(otelchi.TraceHeaderConfig{}),
+		),
+	)
+
+	svc, err := NewAuthenticationApi(options.Config, &options.Logger, options.Metrics, options.TraceProvider, m)
+	if err != nil {
+		panic(err) // TODO p.bleser what to do when we encounter an error in a NewService() ?
```
butonic: The authentication api should be passed in as an option so the creation cannot fail here.

services/groupware/DEVELOPER.md
```diff
@@ -0,0 +1,761 @@
...
+* <u>Bind Password:</u> `admin`
+* <u>Base DN:</u> `ou=users,dc=opencloud,dc=eu`
+* <u>Host:</u> `localhost`
+* <u>LDAP Port:</u> `389`
+* <u>LDAPS Port:</u> `636`
+
+Run the following command on your host (requires the `ldap-tools` package with the `ldapsearch` CLI tool), which should output a list of DNs of demo users:
+
+```bash
+ldapsearch -h localhost -D 'cn=admin,dc=opencloud,dc=eu' \
```
butonic: this only works with the url for me
```suggestion
ldapsearch -H ldap://localhost -D 'cn=admin,dc=opencloud,dc=eu' \
```

services/groupware/DEVELOPER.md
```diff
@@ -0,0 +1,761 @@
...
+> [!NOTE]
+> Only available in the [&ldquo;production&rdquo; setup](#prod-setup)
+
+To check whether it works correctly, the following `curl` command:
+
+```bash
+curl -ks -D- -X POST \
+"https://keycloak.opencloud.test/realms/openCloud/protocol/openid-connect/token" \
+-d username=alan -d password=demo -d grant_type=password \
+-d client_id=groupware -d scope=openid
```
butonic: when working from top to bottom this fails. using a `client_id=web` works ... so I need to add a client id for the goupware somewhere ...

services/groupware/DEVELOPER.md
```diff
@@ -0,0 +1,761 @@
...
+enter the following command:
+
+```bash
+A LOGIN alan demo
+```
+
+to which one should receive the following response:
+
+```java
+A OK [CAPABILITY IMAP4rev2 ...] Authentication successful
```
butonic, line 794: hm, I need to dig into:
```
stalwart-1            | 2025-12-10T16:07:05Z TRACE Raw IMAP input received (imap.raw-input) listenerId = "imaptls", localPort = 993, remoteIp = 172.39.0.1, remotePort = 45034, size = 19, contents = "A LOGIN alan demo\r\n"
stalwart-1            | 2025-12-10T16:07:05Z ERROR LDAP error (store.ldap-error) listenerId = "imaptls", localPort = 993, remoteIp = 172.39.0.1, remotePort = 45034, reason = "I/O error: Connection refused (os error 111)", causedBy = "crates/directory/src/core/dispatch.rs:25", id = "A"
```

devtools/deployments/opencloud_full/.env
```diff
@@ -305,8 +305,15 @@ KEYCLOAK_ADMIN_PASSWORD=
 # Leaving it default stores data in docker internal volumes.
 #RADICALE_DATA_DIR=/your/local/radicale/data
 
+### Stalwart Settings ###
+# Note: the leading colon is required to enable the service.
+#STALWART=:stalwart.yml
+# Domain of Stalwart
+# Defaults to "stalwart.opencloud.test"
+STALWART_DOMAIN=
+
```
butonic, line 320:
```suggestion
STALWART_DOMAIN=
# LDAP config to use. Can either be idmldap (the built in IdP) or ldap (when using keycloak).
STALWART_AUTH_DIRECTORY=idmldap
```

services/groupware/DEVELOPER.md
```diff
@@ -0,0 +1,761 @@
...
+```
+
+> [!TIP]
+> Until everything is documented, the complete list of URI routes can be found in \
+[`$OCDIR/opencloud/services/groupware/pkg/groupware/groupware_route.go`](./pkg/groupware/groupware_route.go)
+
+## Services
+
+### Stalwart
+
```
butonic, line 1125: add a section about the STALWART_AUTH_DIRECTORY env var

services/groupware/pkg/groupware/api.go
```diff
@@ -0,0 +1,19 @@
+package groupware
```
butonic, line 1: I think we should drop the `groupware_` prefix from files in this package

pkg/jmap/api.go
```diff
@@ -0,0 +1,49 @@
+package jmap
```
butonic, line 1: I think we should drop the `jmap_` prefix from all files in this package

devtools/deployments/opencloud_full/opencloud.yml
```diff
@@ -58,6 +58,11 @@ services:
       COMPANION_DOMAIN: ${COMPANION_DOMAIN:-companion.opencloud.test}
       # enable to allow using the banned passwords list
       OC_PASSWORD_POLICY_BANNED_PASSWORDS_LIST: banned-password-list.txt
+      IDM_REVASVC_PASSWORD: "admin"
+      AUTH_BASIC_LDAP_BIND_PASSWORD: "admin"
+      USERS_LDAP_BIND_PASSWORD: "admin"
+      GROUPS_LDAP_BIND_PASSWORD: "admin"
+      IDM_LDAPS_ADDR: 0.0.0.0:9235
```
butonic, line 65: I configured everything on `.compose.test` to prevent collisions with my `.opencloud.test` deployment. The stalwart URL then needs to be set for the groupware:
```suggestion
      IDM_LDAPS_ADDR: 0.0.0.0:9235
      GROUPWARE_JMAP_BASE_URL: https://${STALWART_DOMAIN:-stalwart.opencloud.test}
```

butonic (changes requested):
Ok, I still don't have a menu entry for mail in the web UI even though the config.json contains the `mail` app. I assume that is a problem with my setup/config. in any case, I can see the mail UI when navigation to `{$OC_URL}/mail` manually. \o/

I did run into some minor issues when following the docs. If you could address them ~~(and enlighten me on how to get menu to show the groupware apps)~~ I'm happy to merge this. I mostly want others to be able follow the DEVELOPMENT.md and have a working setup. Kudos for that, btw.

The groupware service itself follows our ... boilerplate ... service code ... and implements the JMAP handling. I assume that will have to evolve, but we can merge it and iterate.

> [!TIP]
> the web repo has [a compose file](https://github.com/opencloud-eu/web/blob/main/docker-compose.yml) with all the apps enabled. That gave me the final hint to get the menu entries

devtools/deployments/opencloud_full/opencloud.yml
butonic: oh and we need to set `FRONTEND_GROUPWARE_ENABLED: "true"` and enable the `mail` app in the web config when stalwart is enabled ... but that connot be configured with a simple env var ... 😞 we need to replace the config file then ...

services/groupware/DEVELOPER.md
pbleser-oc: ✅ done in 10d1a99a14b5d714432af974f3fa7e228591e032

services/groupware/pkg/groupware/api.go
pbleser-oc: ✅ done in 19947b12815d231dc7e5b218f1b0b76e08d1bf7b

devtools/deployments/opencloud_full/.env
pbleser-oc: Done in 67dd937a02828b866dd8955f6eace15f8916d3cc

services/groupware/DEVELOPER.md
pbleser-oc: I did add a JSON file `devtools/deployments/opencloud_full/config/keycloak/clients/groupware.json`, expecting it to be automatically picked up, but that might require re-creating the Keycloak container... ?
In any case, that was a beginner's mistake, there is really no reason to have a distinct client ID for the groupware, as the Web UI is going to use `web` instead.

Removed that config file and updated the documentation accordingly to use `web` instead, in cfbbe02e64cb362da9eec64580c53beab7df736b

Issue #1580, opened 2025-09-29
create a buildflag for stable / rolling version


db-ot:
## Userstory

As a DEV, i want to know what kind of type the release is.

# Acceptance Criteria

- create a flag in the build process for the release type ( rolling or stable )
- report build type alongside the version

Issue #1549, opened 2025-09-24
Announce if running a stable or rolling version


kulmann:
# Description

## User Stories

* > As a web dev, I want know the info if the opencloud instance is running on a rolling or stable version, so that I can request the correct channel from the update server for displaying if an update is available.

## Value

Display update availability in the web ui:
https://github.com/opencloud-eu/web/issues/1158

## Acceptance Criteria
- along with the current version, opencloud announces the update channel (rolling or stable)
- can we use the already existing `edition` field for this?
- information must be contained in the binary, not via configuration

## Open questions
- how can we prevent jailing rolling instance admins into the stable channel when they update to the next production release?

## Definition of ready
- [ ] Everybody needs to understand the value written in the user story
- [ ] Acceptance criteria have to be defined
- [ ] All dependencies of the user story need to be identified
- [ ] Feature should be seen from an end user perspective
- [ ] Story has to be estimated
- [ ] Story points need to be less than 20

## Definition of done
- Functional requirements
  - [ ] Functionality described in the user story works
  - [ ] Acceptance criteria are fulfilled
- Quality
  - [ ] Code review happened
  - [ ] CI is green (that includes new and existing automated tests)
  - [ ] Critical code received unit tests by the developer
- Non-functional requirements
  - [ ] No sonar cloud issues
- Configuration changes
  - [ ] The next branch of the OpenCloud charts is compatible

kulmann:
depends on https://github.com/opencloud-eu/opencloud/issues/1580

kulmann:
this is already done 👍

PR #2001, opened 2025-12-09
enhancement: introduce release channel capability

fschade:
## Description
set the edition capability to a build time flag which represents the release channel! Available channels are:

- `dev`: local builds and builds without ldflags `$(OC_REPO)/pkg/version.Edition=$(EDITION)`
- `rolling`
- `stable`
- `lts-0.0.0`, the contained version number must follow semver!

unfortunately oc had the public config sets for setting the edition as part of the config or as env variable such as:

- `config:edition`
- `env:OC_EDITION`
- `env:FRONTEND_EDITION`
- `env:OCDAV_EDITION`

the ability to configure that value has been removed as part of the pr, deprecating the env is not easy and it should not be in use anyway, but who knows..., @micbar lets discuss, i still can change the behavior!

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/1580
- Enables https://github.com/opencloud-eu/opencloud/issues/1549

## Motivation and Context
Having detailed information of which build channel the current server binary is originated from, web also requested such a capability to display it. 

## Todos:

- [x] add the `Env:EDITION` to the each related block of our woodpecker.yml
- [x] add the `Arg:EDITION` to all related docker images
- [x] check web, just in case
- [x] get green light from @TheOneRing 

## Questions:

- Q: Should our make step automatically parse the current branch to determine the current `Env:CHANNEL`, should work out of the box for `stable`, but what about rolling and nightly?
  - A: ✅
    - ci binary builds are always of type `edition:rolling`
    - ci docker production builds are always of type `edition:stable`
    - ci docker other builds (rolling, and daily) are always of type `edition:rolling`

- Q: if the channel is unknown running the binary fails (i added a validation for that), current allowed values are: `rolling, rolling, stable, and lts-0.0.0`, in other words, it panics with a message, ok or maybe we should just log a warning!? ..., @kulmann, @micbar 
  - A: ✅ @flimmy and myself decided not to panic, it logs an error instead and falls back to its edition default value which is `dev`

- Q: any woodpecker magic needed? Tags, manuall triggered ci runs, ..., @flimmy 
  - A: ✅ nope, default type is daily which results in an rolling edition

- Q: removing `env:OC_EDITION`, `...` without deprecating it is ok!?
  - A: - ✅ undocumented

## How Has This Been Tested?
- manually checking the capability (edition)

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


fschade:
@kulmann, @db-ot FYI

flimmy (approved):
looking good

PR #2060, opened 2025-12-18
Fix edition error and image arg default value

fschade:
## Description
the edition validation does not mention the configured edition (go build flag) which results in:

```bash
{"level":"error","error":"unknown edition channel 'dev'","time":"2025-12-18T12:56:14.607885675Z","caller":"github.com/opencloud-eu/opencloud/pkg/version/version.go:55","message":"falling back to dev"}
```

## Motivation and Context
if a configured edition is not allowed the error should mention the used edition, e.g.

```bash
{"level":"error","error":"unknown edition channel 'foo'","time":"2025-12-18T12:56:14.607885675Z","caller":"github.com/opencloud-eu/opencloud/pkg/version/version.go:55","message":"falling back to dev"}
```

also, the current EDITION ARG in Dockerfile.multiarch has no default value which always results in a logged error if the edition is not provided.

## How Has This Been Tested?
- local docker builds
- unit-tests

## Types of changes
- [ ] Technical debt

### DIFF
--- a/opencloud/docker/Dockerfile.multiarch
+++ b/opencloud/docker/Dockerfile.multiarch
@@ -3,7 +3,7 @@ ARG TARGETOS
 ARG TARGETARCH
 ARG VERSION
 ARG STRING
-ARG EDITION
+ARG EDITION="dev"
 
 RUN apk add bash make git curl gcc musl-dev libc-dev binutils-gold inotify-tools vips-dev
 


pkg/version/version.go
```diff
@@ -79,8 +79,11 @@ func initEdition() error {
 		_, err := semver.NewVersion(editionParts[1])
 		return err == nil
 	}) {
-		Edition = Dev
-		return fmt.Errorf(`unknown edition channel "%s"`, Edition)
+		defer func() {
+			Edition = Dev
+		}()
+
+		return fmt.Errorf(`unknown edition channel '%s'`, Edition)
```
butonic, line 86: That defer is super confusing. Why do we need to reset Edition to Dev in the first place?
```suggestion
		err := fmt.Errorf(`unknown edition channel '%s'`, Edition)

		// reset to Dev edition
		Edition = Dev

		return err
```

fschade: fine for me too

PR #2072, opened 2025-12-21
feat: support desktop and mobile specific  `client_id` and `scopes`

pat-s:
## Description

Enables identity providers that require separate OIDC clients per application type (like Authentik, Kanidm, Zitadel) to work with OpenCloud clients.

- Add desktop-specific OIDC issuer relation (`http://openid.net/specs/connect/1.0/issuer/desktop`)
- Add mobile-specific OIDC issuer relation (`http://openid.net/specs/connect/1.0/issuer/mobile`)
- Support optional `client_id` property in WebFinger link responses

## Configuration

```yaml
WEBFINGER_OIDC_ISSUER_DESKTOP: "https://idp.example.com"
WEBFINGER_OIDC_CLIENT_ID_DESKTOP: "desktop-client-id"
WEBFINGER_OIDC_ISSUER_MOBILE: "https://idp.example.com"
WEBFINGER_OIDC_CLIENT_ID_MOBILE: "mobile-client-id"
```

## Motivation and Context

This problems solves https://github.com/opencloud-eu/desktop/issues/246 and is a complementary PR to https://github.com/opencloud-eu/desktop/pull/766. 

Besides the PR for the desktop app, similar changes are needed for the iOS and Android sources.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


kaivol:
I don't think we should allow/encourage using different OIDC issuers (and client ids) for web and desktop/mobile clients. 
There really is no reason to do so, and in fact it is incorrect to assume that users from different issuers represent the same entity. 

---

Also, I would add the OIDC scopes to request during authentication, just like it is already possible for the web client with `WEB_OIDC_SCOPE`. This would also fix https://github.com/opencloud-eu/desktop/issues/217.

mavgit1:
> I don't think we should allow/encourage using different OIDC issuers (and client ids) for web and desktop/mobile clients. There really is no reason to do so, and in fact it is incorrect to assume that users from different issuers represent the same entity.
> 
> Also, I would add the OIDC scopes to request during authentication, just like it is already possible for the web client with `WEB_OIDC_SCOPE`. This would also fix [opencloud-eu/desktop#217](https://github.com/opencloud-eu/desktop/issues/217).

You are right that using different Issuers would break user identity (tough that is necessary to register unique urls from and idp like Authentik), but having **separate Client IDs for Web and Native apps is actually standard OIDC practice.**

Some IDPs (like Google Identity or Zitadel) strictly enforce this separation for security reasons (handling localhost vs https redirects) and completely refuse to mix Web and Native configurations in a single Client ID. As long as the Issuer remains the same, the user identity stays consistent.

So having the option to provide 2 different Client IDs is necessary so that Owncloud can be used with standard security practices / even work at all with some IDPs.

kaivol:
> having **separate Client IDs for Web and Native apps is actually standard OIDC practice.**

Could you elaborate on this point? During my use of OIDC (mostly in the self-hosting area), I never encountered such applications. 
I can't think of any type of attack that would be prevented by this practice. 

But anyway, if there are servers that make such requirements (different _Client ID_ per redirect URI, IIUC), we should of course support that. 
Is one _Client ID_ for all native clients sufficient, or do we need different client for different platforms?

---



> using different Issuers would break user identity (tough that is necessary to register unique urls from and idp like Authentik)

Could you explain what you mean by that? As far as I know, Authentik supports defining multiple redirect URIs per client.

mavgit1:
> > having **separate Client IDs for Web and Native apps is actually standard OIDC practice.**
> 
> Could you elaborate on this point? During my use of OIDC (mostly in the self-hosting area), I never encountered such applications. I can't think of any type of attack that would be prevented by this practice.
> 

image

Eg. in Zitadel you are forced to create a seperate app both for WEB and Native. 

You cannot add a local redirect uri like oc://android.opencloud.eu or even http://127.0.0.1 to a WEB app. Vice versa you cannot add a normal url like https://your-domain.example.com/oidc-callback.html to a native app.

OpenCloud even reccomends having 3 seperate clients. One for Mobile App, one for Dekstop App and one for WEB. So it would be compatible with Zitadel and many others if simply the CLIENT-ID could be set manually.

> > using different Issuers would break user identity (tough that is necessary to register unique urls from and idp like Authentik)
> 
> Could you explain what you mean by that? As far as I know, Authentik supports defining multiple redirect URIs per client.

Yes authentik supports different URIs per client and I think even the renaming of client ID (thus being able to name the client fixed default names from OpenCloud) but each App is a different issuer URL:

See currently open issue:

https://github.com/goauthentik/authentik/issues/7251

butonic:
AFAICT there are two problems:
## 1. client_id registration

We try to follow OIDC best practices, which is why all of our clients have a different client_id. Each client has his own redirect urls, token lifetimes, etc. We currently hardcode these client_ids for our clients:
- OpenCloud Web App: `web`
- OpenCloud Desktop Client: `OpenCloudDesktop` 
- OpenCloud Android App: `OpenCloudAndroid`
- OpenCloud iOS App: `OpenCloudIOS`
- For more details see the [Client Configuration docs](https://docs.opencloud.eu/docs/admin/configuration/authentication-and-user-management/external-idp/#client-configuration).

Can Authentik, Kanidm, Zitadel and Authelia be configured to use these `client_id`s? If not, I would consider that a bug in them.

👀 ... yeah, Zitadel seems to assume every add developer will hardcode their generated client id in their codebase: https://zitadel.com/docs/examples/login/flutter. How is that supposed to work when an open source project develops an app and someone tries to use it with Zitadel. It seems the readonly client_id is by design: https://github.com/zitadel/zitadel/issues/10149

> [!TIP]
>For ADFS powershell allows setting the `client_id` with, eg:
> ```
> Add-AdfsClient -Name "OpenCloud Desktop" -ClientId "OpenCloudDesktop" -RedirectUri "http://localhost:*"`
> ```

The *spec* way of getting a shared `client_id` (all iOS devices get `OpenCloudIOS`, all android devices get `OpenCloudAndroid`, etc) would be [Dynamic Client Registration (RFC 7591) A.4.2 Client ID Shared among All Instances of Client Software](https://www.rfc-editor.org/rfc/rfc7591#appendix-A.4.2), but AFAICT it would also require [Software Statements](https://www.rfc-editor.org/rfc/rfc7591#section-2.3): our clients would at least send a `software_id` property in the registration request that the IdP can use to look up the corresponding `client_id`. Not even keycloak supports that, yet.

Without Software Statements dynamic client registration would lead to on client per android app, which does not add any security benefits (sessions can already be killed individually) at the cost of a huge management overhead. So, Dynamic Client Registration is currently not an option, IMO.

If there are IdPs that currently do not support configuring the client_id we would have to touch all our clients and the server to dynamically look up the client_id.

Maybe the time is better spent on documenting how to configure or change the `client_id` with different IdPs? We had hoped that OIDC would standardize a few things ... but it seems it left out a few implementation bits. 😞 

Honestly, I would prefer if admins would not have to configure anyting manually, however, either they have to configure opencloud to use the client ids generated by the IdP , or they have to configure the IdP to use our hardcoded client ids. That is, until dynamic client registration with software statements is widely supported. 

## 2. multiple issuer urls
I guess here it is the other way around: OpenCloud currently only supports a singe OIDC issuer. Our clients discover the IdP via a `GET https://cloud.opencloud.test/.well-known/openid-configuration` request that we proxy to the IdP configured on the server side with `OC_OIDC_ISSUER` (defaults to `OC_URL` for the built in IdP).

When different Issuer URLs are used for different clients, where should we proxy the request to? We would have to route the request based on the user agent ... 🫨 

----

The webfinger approach seems to be a pragmatic solution to both problems. However, there are several problems:
1. we cannot just invent new OIDC issuer relations in the openid namespace (http://openid.net/specs/connect/1.0/issuer/desktop). 
2. we have at least four client ids: web, OpenCloudDesktop, OpenCloudAndroid, OpenCloudIOS. So we need at least four env vars to configure all of these.
3. multiple issuers is still a problem

Tools like cyberduck or rclone will not be able to look up the configuration using webfinger because it does not follow any official spec ... but end users have to provide the client id and maybe secret on the cli anyway.

I had internally proposed to use properties in our webfinger response like this:
```json

{
  "subject": "acct:alice@tenant.opencloud.example",
  "links": [
    {
      "rel": "http://openid.net/specs/connect/1.0/issuer",
      "href": "https://idp.tenant.example.com"
    }
  ],
  "properties": {
    "client_ids": {
      "web": "opencloud-web-123",
      "desktop": "opencloud-desktop-456",
      "android": "opencloud-789",
      "ios": "opencloud-ios-abc"
    },
    "scopes": {
      "web": ["openid", "profile", "email"],
      "desktop": ["openid", "profile", "email", "roles"]
    },
    "config_url": "https://cloud.opencloud.test/config.json"
  }
}
```

But this does not include the issuer.

Maybe we should add a platform query parameter to the webfinger request:
```http
GET /.well-known/webfinger?resource=https://drive.opencloud.test&platform=android

Response:
{
  "subject": "https://drive.opencloud.test",
  "links": [
    {
      "rel": "http://openid.net/specs/connect/1.0/issuer",
      "href": "https://idp.example.com/tenants/acme/clients/opencloud-android"
    }
  ],
  "properties": {
    "client_id": "opencloud-android"
    "scopes": ["openid", "profile", "email"],
    "config_url": "https://cloud.opencloud.test/android-config.json"
  }
}
```
```http
GET /.well-known/webfinger?resource=https://drive.opencloud.test&platform=ios

Response:
{
  "subject": "https://drive.opencloud.test",
  "links": [
    {
      "rel": "http://openid.net/specs/connect/1.0/issuer",
      "href": "https://idp.example.com/tenants/acme/clients/opencloud-ios"
    }
  ],
  "properties": {
    "client_id": "opencloud-ios",
    "scopes": ["openid", "profile", "email", "roles"],
    "config_url": "https://cloud.opencloud.test/ios-config.json"
  }
}
```

This way:
- we don't invent new relations
- It is fully backwards compatible
- every client only sees the information he requested
- we can return the client_id
- we can return a different issuer href if necessary
- we can return differnet scopes the clients should request

@rhafer @micbar opinions?

micbar:
@butonic I think your reasoning about the webfinger spec makes sense to me.

Using a platform parameter could solve this.

butonic:
Regarding multiple issuers: I'm not sure this can work.

If the clients use different issuers, I have to assume they will receive different iss claims in the auth token / user info endpoint as well. Then the server will see four different users because the iss is different. They will not even see the same personal space ... and I am not even sure our proxy would use the issuer url from the token ... if it is a jwt, then yes. otherwise we will use the configured issuer to make a userinfo lookup ... with an access token from a different issuer ... AFAICT this is not how openid connect is supposed to work.

Maybe this works if the idp always returns the same issuer url in the userinfo endpoint, regardless of the issuer that created the access token... but that seems to be even more of a glitch then.

I might just be ignorant. Which IdP uses different issuers per client? Maybe they can give some reasoning.

And no, this has nothing to do with pairwise identifiers that creates a new sub for every relying party so they cannot crossreference user metadata - by design - for security reasons ...

mavgit1:
> I might just be ignorant. Which IdP uses different issuers per client? Maybe they can give some reasoning.
> 

https://github.com/goauthentik/authentik/issues/7251


this is the issue where they talk about solving that for authentik 


currently i think authentik and maybe? kandim or something else have that

they just have a different and specific issuer path for every app that was registered

pat-s:
> Can Authentik, Kanidm, Zitadel and Authelia be configured to use these client_ids? If not, I would consider that a bug in them.

In the overall discussion I am wondering: it reads as if (almost) all IDP tools would do this wrong (since ever) and should change and OpenCloud is having a very strong opinion on what is right or wrong for that detail specifically. 

The world has seen other tools before OpenCloud which attempted to work with OIDC through desktop/web/mobile Apps.
I don't understand why this topic is filled with such strong opinions instead of working/complying with what exists (which is surely also not just a random insecure implementation by design).

On top, this topic itself *should* be of high importance for the overall adoption and distribution of OpenCloud itself. However, it is treated as a "maybe some day" issue for a long time already, which isn't really understandable when looking at it from a broader perspective.

---

PS: This is a PR. The discussion about the general issue should happen in the linked issue or a dedicated discussion thread. I'd like to know if the PR here will ever be considered or whether I should close it.

mavgit1:
Why don't the Mobile / Desktop clients just request some sort of discovery json from the server?

That would not cause any further config for an enduser connecting to own cloud and the Url / https traffic is allready trusted by a CA or the user is using owncloud local only anyways. There is no security gained by hardcoding it id argue.

I think that would be a pragmatic way to solve it which also adds no complex logic to the webfinger response.

mavgit1:
Also on a sidenote:

If users are in a B2B scenario where a single IDP instance exists but multiple Own Cloud instances are used, using hard-coded ID's would cause a conflict because Client_ID goes beyond an Organization in many IDP's. This is probably also the reason why you can manually set it in some and not in others.

There are two kind of approaches "islands" like eg. Keycloak or Authentik - there an Organization is isolated. Eg. multiple Relams can have the same ID i think. You'd only have an issue if you wanted the same app twice in the same realm.

And more like a "Shared" approach - eg. in Zitadel, Auth0, Google, Okta - an App can be shared across Organizations. Thus the ID must be unique and not pre-defined. 3 Organizations can use the same app with the same client_id and just be granted access to it. Thus, a pre defined ID would prevent users from adding a second instance of the same app to the entire IDP.

butonic:
I reference the proposed query parameter in https://github.com/opencloud-eu/desktop/issues/246#issuecomment-3723824039.

@pat-s feel free to either rewrite this PR or create a new PR if you agree with using a query param.

kaivol:
> I reference the proposed query parameter in [opencloud-eu/desktop#246 (comment)](https://github.com/opencloud-eu/desktop/issues/246#issuecomment-3723824039).

I could take care of implementing these changes if @pat-s is not already working on it.

pat-s:
@kaivol Thanks for suggesting to help. I'll continue myself first in the next days and see if I can get there.

Firstyear:
> I can't think of any type of attack that would be prevented by this practice.

Client ID's also have client *secrets* associated in confidential clients. You don't want to be sharing secrets.

If you re-use client ideas, there are some possible ways to also intercept and replay token exchange calls (which is why you should have PKCE on all client flows, even confidential ones).

Additionally, Client ID's tend to be bound to a cryptographic key domain - the signing keys that issue tokens. By sharing a client ID, you allow tokens from one client to be used directly to another. 

In the case here, you probably want a public client id that can be configured in your desktop/mobile in someway.

kaivol:
@Firstyear 
> Client ID's also have client _secrets_ associated in confidential clients. You don't want to be sharing secrets.

Yes, of course, I should have clarified that I'm talking about non-confidential clients without client secrets. 

> Additionally, Client ID's tend to be bound to a cryptographic key domain - the signing keys that issue tokens. By sharing a client ID, you allow tokens from one client to be used directly to another.

That's not really a problem given the way OpenCloud works with different clients (all clients grant the same rights, access to all files of the user), is it?

Chanceium:
Can we get this implemented lol...

rhafer:
> Regarding multiple issuers: I'm not sure this can work.

I agree. But in the cases where the issuer URL is different per client id (AFAIK this boils down to Authentik) all clients should probably use the same client-id (this has been requested many of the Authentik related discussion/tickets anyway).

When this, and the related client PRs, are merged, a setup like that should be easily to accomplish.

I don't think we should add support of different issuer url per client. Because that seems really wrong with regards to the OIDC spec. It specifically says that the combination of the `issuer` and the `subject`  are the only claims that you can rely upon as a stable identifier for a user. By that definition we'd have to consider a user with `sub: user1` and `iss: http://issuer/desktop` to be a different user than the user with `sub: user1` and `iss: http://issuer/web`

But as said if we go forward with this PR people can just tweak the configuration so that the webfinger endpoint returns the same client id value for all four client (which should work with authentik)

services/webfinger/pkg/relations/openid_discovery.go
```diff
@@ -8,7 +8,9 @@ import (
 )
 
 const (
-	OpenIDConnectRel = "http://openid.net/specs/connect/1.0/issuer"
+	OpenIDConnectRel        = "http://openid.net/specs/connect/1.0/issuer"
+	OpenIDConnectDesktopRel = "http://openid.net/specs/connect/1.0/issuer/desktop"
+	OpenIDConnectMobileRel  = "http://openid.net/specs/connect/1.0/issuer/mobile"
```
rhafer: I don't think we should invent new relations here. (As also mentioned by @butonic in the PR discussion)

Apart from the fact that OpenCloud itself does currently only support talking a single IDP and configuring multiple different issuer URL would effectively mean we'd need to add multi-IDP support, I really think we don't need that.

As outlined [here](https://github.com/opencloud-eu/opencloud/pull/2072#issuecomment-3777549184) support for multiple issuers is not even needed. In the case of Authentik where one application (OpenCloud) can just have a single clientid we should just use the same client id for all clients.

If the clients fetch their client id via webfinger that is easy to be achived without adding the complexity of multiple issuer urls.

services/webfinger/pkg/config/config.go
```diff
@@ -20,6 +20,10 @@ type Config struct {
...
+	DesktopClientID string `yaml:"desktop_client_id" env:"WEBFINGER_OIDC_CLIENT_ID_DESKTOP" desc:"The OIDC client ID for desktop clients. When set along with WEBFINGER_OIDC_ISSUER_DESKTOP, this client ID will be provided to desktop clients via webfinger properties." introductionVersion:"%%NEXT%%"`
+	MobileIDP       string `yaml:"mobile_idp" env:"WEBFINGER_OIDC_ISSUER_MOBILE" desc:"The identity provider href for mobile clients. When set, mobile clients will use this issuer instead of the default IDP. This allows configuring separate OIDC clients for web and mobile applications." introductionVersion:"%%NEXT%%"`
+	MobileClientID  string `yaml:"mobile_client_id" env:"WEBFINGER_OIDC_CLIENT_ID_MOBILE" desc:"The OIDC client ID for mobile clients. When set along with WEBFINGER_OIDC_ISSUER_MOBILE, this client ID will be provided to mobile clients via webfinger properties." introductionVersion:"%%NEXT%%"`
```
rhafer: Additionally to the `clientid` I think we also need to be able to set a list of `scopes` per client. So that every client is able to figure out with exact scopes it needs to request for the access_-/id_token.

This is needed for stuff like automatic role assignment and group provisioning to work in a more generic way and not just for keycloak. (See e.g. https://github.com/opencloud-eu/desktop/issues/217#issuecomment-3007974759)

rhafer (changes requested):
@pat-s Thanks a lot for the effort you've spent on this. That is highly appreciated.

As @butonic mentioned, it would be nice to change the `WebFingerHandler` to support the additional query parameter `platform`. And based on that return the platform specific `clientid` and list of `scopes` in the `properties` part of the response.

To stay backwards compatible and not break the current clients, the response when the `platform` parameter is missing should stay unchanged for now.

With those changes I think we could move forward and merge this PR. After that  obviously all clients need to be adjust to use the new discovery mechnism.

pat-s:
@rhafer Happy to hand over the finishing touches to a team member. I don't need to do them myself, especially as the two PR across repos must stay in sync somehow to work together.

Do you (or somebody else) from the team take over?

rhafer:
> @rhafer Happy to hand over the finishing touches to a team member. I don't need to do them myself, especially as the two PR across repos must stay in sync somehow to work together.
> 
> Do you (or somebody else) from the team take over?

I've reworked the PR to what I think would fulfill all the requirements in the various related tickets and pushed it to your branch. To discover the correct issuer, clientid and scope the clients could just do a webfinger request like this:

```
curl -k 'https://localhost:9200/.well-known/webfinger?resource=https://localhost:9200&rel=http://openid.net/specs/connect/1.0/issuer&platform=desktop'
```

The result is something like this:
```
{
  "subject": "https://localhost:9200",
  "properties": {
    "http://opencloud.eu/ns/oidc/client_id": "OpenCloudDesktop",
    "http://opencloud.eu/ns/oidc/scopes": [
      "openid",
      "profile",
      "email",
      "offline_access"
    ]
  },
  "links": [
    {
      "rel": "http://openid.net/specs/connect/1.0/issuer",
      "href": "https://localhost:9200"
    }
  ]
}
```

services/webfinger/pkg/config/config.go
```diff
@@ -17,11 +17,21 @@ type Config struct {
...
+	DesktopClientID     string     `yaml:"desktop_client_id" env:"WEBFINGER_DESKTOP_OIDC_CLIENT_ID" desc:"The OIDC client ID for the OpenCloud desktop application." introductionVersion:"%%NEXT%%"`
+	DesktopClientScopes []string   `yaml:"desktop_client_scopes" env:"WEBFINGER_DESKTOP_OIDC_CLIENT_SCOPES" desc:"The OIDC client scopes the OpenCloud desktop application should request." introductionVersion:"%%NEXT%%"`
+	IOSClientID         string     `yaml:"ios_client_id" env:"WEBFINGER_IOS_OIDC_CLIENT_ID" desc:"The OIDC client ID for the IOS app." introductionVersion:"%%NEXT%%"`
+	IOSClientScopes     []string   `yaml:"ios_client_scopes" env:"WEBFINGER_IOS_OIDC_CLIENT_SCOPES" desc:"The OIDC client scopes the IOS app should request." introductionVersion:"%%NEXT%%"`
+	WebClientID         string     `yaml:"web_client_id" env:"WEBFINGER_WEB_OIDC_CLIENT_ID" desc:"The OIDC client ID for the OpenCloud web client." introductionVersion:"%%NEXT%%"`
+	WebClientScopes     []string   `yaml:"web_client_scopes" env:"WEBFINGER_WEB_OIDC_CLIENT_SCOPES" desc:"The OIDC client scopes the OpenCloud web client should request." introductionVersion:"%%NEXT%%"`
```
kaivol: Should we integrate this with the existing environment variables like `OC_OIDC_CLIENT_ID` and `WEB_OIDC_SCOPE`?

I think `AndroidClientID`, `DesktopClientID`, etc. could fall back to `OC_OIDC_CLIENT_ID`, and `AndroidClientScopes`, `DesktopClientScopes`, etc., as well as the existing `WEB_OIDC_SCOPE` could fall back to a new `OC_OIDC_SCOPE` variable. 

This would reduce the required environment variables for setups that use a single OIDC issuer for all clients.

kaivol:
The existing code related to additional OIDC scopes for the `web` service uses the singular "scope",
https://github.com/opencloud-eu/opencloud/blob/e0b465342f1389d91869bafa07f4d230c92493e0/services/web/pkg/config/config.go#L74
while this PR uses the plural "scopes". We should probably use the same wording for both.

rhafer:
> while this PR uses the plural "scopes". We should probably use the same wording for bo

> The existing code related to additional OIDC scopes for the `web` service uses the singular "scope",
> 
> https://github.com/opencloud-eu/opencloud/blob/e0b465342f1389d91869bafa07f4d230c92493e0/services/web/pkg/config/config.go#L74
> 
> while this PR uses the plural "scopes". We should probably use the same wording for both.

I'd prefer the plural from for the new config values, because it really is a multi-valued attribute. But I think adding support for `WEB_OIDC_SCOPE` (singular) for the web settings as a fallback for backwards compatibility makes sense. To keep exiting setups working. The intention is also adapt the web frontend use the webfinger based approach for discovering the OIDC settings and remove the `WEB_OIDC_SCOPE` and `WEB_OIDC_CLIENT_ID` settings from `services/web/pkg/config/config.go` when that happened.

PR #2099, opened 2026-01-01
[POC] ci: skip previously passed workflows on pipeline restart

saw-jan:
### How does it work?
1. First CI pipeline run, save the current pipeline number to s3 using the **commit hash as a key**
2. Second CI pipeline run (**_Restart on the same commit_**), check the previously saved pipeline number in s3
3. If found, get the pipeline info using woodpecker API and save the info to s3 inside the same commit hash key path
4. Then, in each test workflow, check if this was passed previously (if passed, then it will be marked as skip)
5. Each step in the test workflow will be skipped if `SKIP_WORKFLOW=true` is found in `.woodpecker.env`

Notes:
- If the pipeline is running with a new commit or there is a change in the commit, then there will be no saved pipeline number for that and the workflows will continue normally as the first run

### Implemented workflows
This behaviour is implemented in all the test workflows.
- API tests (local and core)
- web-e2e tests
- cs3-API tests
- wopi-validator tests
- unit tests
- k6 tests

### Working example
✔️ When the workflow passed previously (steps skipped):
  - Previous pipeline: https://ci.opencloud.rocks/repos/3/pipeline/717/61
  - Current restarted pipeline: https://ci.opencloud.rocks/repos/3/pipeline/718/61

❌ When the workflow failed previously (steps run):
  - Previous pipeline: https://ci.opencloud.rocks/repos/3/pipeline/717/54
  - Current restarted pipeline: https://ci.opencloud.rocks/repos/3/pipeline/718/54
## Related Issue
- resolves https://github.com/opencloud-eu/qa/issues/55

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -570,6 +570,13 @@ def main(ctx):
         ),
     )
 
+    test_pipelines.append(
+        pipelineDependsOn(
+            purgePipelineInfoCache(),
+            testPipelines(ctx),
+        ),
+    )
+
     pipelines = test_pipelines + build_release_pipelines + notifyMatrix(ctx)
 
     pipelineSanityChecks(pipelines)
@@ -2680,6 +2687,9 @@ def purgeOpencloudWebBuildCache(ctx):
 
 def purgeGoBinCache(ctx):
     return purgeCache("purge_go_bin_cache", "dev/opencloud/go-bin", 14)
+
+def purgePipelineInfoCache():
+    return purgeCache("purge_pipeline_info_cache", "public/opencloud/pipelines", 14)
 
 def pipelineSanityChecks(pipelines):
     """pipelineSanityChecks helps the CI developers to find errors before running it


.woodpecker.star
```diff
@@ -561,14 +565,72 @@ def main(ctx):
...
+        "when": [
+            {
+                "event": ["push", "manual"],
+                "branch": ["main", "stable-*"],
+            },
+            event["tag"],
+            event["cron"],
+            event["pull_request"],
+        ],
+    }]
```
v-scharf: I didn't find where we use it. Can we remove?

fschade:
Cool :), quick question, following scenario:

1st run
workflow A is green
workflow B is red

the dev fixes the bug which made pipeline B fail

woodpecker starts a new run
workflow A is skipped (it was green before)
workflow B is green


this sounds really good, but what happens if the fix which made workflow B green introduced a new bug on workflow A?

v-scharf:
then it will be a new `$CI_COMMIT_SHA` INFO_URL="https://s3.ci.opencloud.eu/public/$CI_REPO_NAME/pipelines/$CI_COMMIT_SHA/pipeline_info.json"
it will not be found (404), and we will run all tests again

fschade:
Ahhhh thanks @ScharfViktor so, it only applies for restarts? That’s really helpful, thanks for explaining 🤗

.woodpecker.star
saw-jan: that one was just for testing. it will be removed

saw-jan:
> then it will be a new `$CI_COMMIT_SHA` INFO_URL="https://s3.ci.opencloud.eu/public/$CI_REPO_NAME/pipelines/$CI_COMMIT_SHA/pipeline_info.json" it will not be found (404), and we will run all tests again

yeah, correct. We only want to restart the failing workflows and the change in commit would mean full CI run

saw-jan:
>Could you please write the logic in the pr description
that we keep info about previous running in s3 and evaluate it when restarting piplene and decide skip or not

Updated the PR description

ScharfViktor (approved):

Issue #144, opened 2025-01-29
Missing onPremisesSamAccountName in user search response


ScharfViktor:
## Steps to reproduce
1. create user `Brian-04f`
2. `einstein` tries to share to new user. search by: `Brian-04f`
3. `admin` tries to share to new user. search by: `Brian-04f`

## Expected behavior
both users have `onPremisesSamAccountName=Brian-04f` in the response body

## Actual behavior
`GET https://opencloud-server:9200/graph/v1.0/users?%24search=%22Brian-04f%22&%24orderby=displayName` 
response body doesn't contain `onPremisesSamAccountName` for `einstein`. `admin` can see `onPremisesSamAccountName` and `email`

if I run opencloud with `OC_SHOW_USER_EMAIL_IN_RESULTS=true`
`einstein` has `onPremisesSamAccountName` but it's empty

![Image](https://github.com/user-attachments/assets/1654872b-1de5-485e-8cd3-667819b50ac2)

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OC_XXX=somevalue
OC_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

## Additional context

cc @rhafer

PR #2104, opened 2026-01-05
fix: Show username in unprivileged search results

rhafer:
`onPremisesSamAccountName` is a mandatory attribute according to the spec. There's no harm in returning it in the search results also for unprivileged users.

Fixes: #144

### DIFF
--- a/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
+++ b/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
@@ -45,7 +45,7 @@ Feature: Multi-tenancy
                   "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
                 },
                 "onPremisesSamAccountName": {
-                  "const": "Alice"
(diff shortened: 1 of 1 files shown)


butonic:
the existing tests expect an empty string, but I think we can now expect `alice`?
```diff
diff --git a/tests/acceptance/features/apiTenancy/mutltiTenancy.feature b/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
index e989a0789b9..21a796c1a4e 100644
--- a/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
+++ b/tests/acceptance/features/apiTenancy/mutltiTenancy.feature
@@ -45,7 +45,7 @@ Feature: Multi-tenancy
                   "pattern": "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
                 },
                 "onPremisesSamAccountName": {
-                  "const": ""
+                  "const": "alice"
                 },
                 "userType": {
                   "const": "Member"
```

v-scharf:
> the existing tests expect an empty string, but I think we can now expect `alice`?

yes, I'll adapt tests

PR #2114, opened 2026-01-07
consolidate log config

dragonchaser:
We have consolidated the log configuration.
The following per service config options have been completely removed:

- `*_LOG_PRETTY`
- `*_LOG_COLOR`
- `*_LOG_FILE`

Please use the following global configuration options for this:

- `OC_LOG_PRETTY`
- `OC_LOG_COLOR`
- `OC_LOG_FILE`

Log level can still be configured per service, use `*_LOG_LEVEL` (e.g. `FRONTEND_LOG_LEVEL`) for this.
The log level can also be still set on a global scale using `OC_LOG_LEVEL`.

### DIFF
--- a/pkg/shared/shared_types.go
+++ b/pkg/shared/shared_types.go
@@ -12,7 +12,7 @@ type EnvBinding struct {
 
 // Log defines the available logging configuration.
 type Log struct {
-	Level  string `yaml:"loglevel" env:"OC_LOG_LEVEL" desc:"The log level. Valid values are: 'panic', 'fatal', 'error', 'warn', 'info', 'debug', 'trace'." introductionVersion:"1.0.0"`
+	Level  string `yaml:"level" env:"OC_LOG_LEVEL" desc:"The log level. Valid values are: 'panic', 'fatal', 'error', 'warn', 'info', 'debug', 'trace'." introductionVersion:"1.0.0"`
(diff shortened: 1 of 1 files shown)


pkg/shared/logging.go
```diff
@@ -0,0 +1,14 @@
+package shared
+
+import "github.com/opencloud-eu/opencloud/pkg/log"
```
butonic: maybe put this file into the into the `github.com/opencloud-eu/opencloud/pkg/log` package in a config.go?

services/activitylog/pkg/config/defaults/defaultconfig.go
```diff
@@ -59,18 +59,9 @@ func DefaultConfig() *config.Config {
...
-			Level:  cfg.Commons.Log.Level,
-			Pretty: cfg.Commons.Log.Pretty,
-			Color:  cfg.Commons.Log.Color,
-			File:   cfg.Commons.Log.File,
-		}
-	} else if cfg.Log == nil {
-		cfg.Log = &config.Log{}
+	if cfg.LogLevel == "" {
+		cfg.LogLevel = "error"
 	}
```
butonic, line 64: I think this can never happen because we set it in the defaultconfig of every service.

dragonchaser: I do not think we do that. EnsureDefaults is part of the defaultconfig....

butonic (approved):

PR #2183, opened 2026-01-20
adr: Add thoughts about guest users

rhafer:
The approach to take has not been decided yet. This just reflects the current research/concept work.

### DIFF
--- a/docs/adr/0004-guest-users.md
+++ b/docs/adr/0004-guest-users.md
@@ -1,12 +1,19 @@
 ---
-title: "Guest users in OpenCloud"
+title: "Inviting Guest users as fully integrated user in OpenCloud"
 ---
 
-* Status: pending
+* Status: postponed
 * Deciders: []
 * Date: 2026-01-20
 
 Reference: https://github.com/opencloud-eu/opencloud/issues/2111
+
+## Important Disclaimer
+
+The approach discussed here has been postponed (as of March, 2026) in favor
+of a different solution that does not require full-blown user-accounts in
+OpenCloud. That approach is currently tracked here:
+https://github.com/opencloud-eu/opencloud/issues/2513
 
 ## Context and Problem statement
 


docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+#### Implementation Obstacles
+
+    - Permissions on spaces are currently not tracked in the share
+      manager, the are purely managed via grants. So currently the share
+      manager service currently does not know anything about (invited)
+      users being assigned to spaces
+
+## Additional thoughts
+
+If OpenCloud were responsible for allocating the UserIDs of all users
```
micbar, line 194: Good point! I was wondering also, if that could simplify things in the future and let us get rid of the LDAP dependency.

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+
+Reference: https://github.com/opencloud-eu/opencloud/issues/2111
+
+## Context and Problem statement
+
+To allow collaboration with external Users (Users that don't yet have an
+account in the IDP, and might be external to the organization), it should
+be possible to invite "Guest Users" into and OpenCloud instance.
+
+## Requirements
```
micbar, line 24: I agree with the requirements. Maybe we need to add that the process can be asynchonous and take some time because we are crossing a lot of boundaries between different systems.

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+* Deciders: []
+* Date: 2026-01-20
+
+Reference: https://github.com/opencloud-eu/opencloud/issues/2111
+
+## Context and Problem statement
+
+To allow collaboration with external Users (Users that don't yet have an
+account in the IDP, and might be external to the organization), it should
+be possible to invite "Guest Users" into and OpenCloud instance.
```
butonic:
```suggestion
be possible to invite "Guest Users" into an OpenCloud instance.
```

docs/adr/0004-guest-users.md
butonic: AFAICT always having opencloud generate a userid would allow us to get rid of the shared ldap deployment mode. The ldap server would only be used to find recipients (users or groups) in an organization. We always only send invitations into some form of inbox of a user. This could literally be an email. Or an internal nats queue with invites.
Then, when he follows the invite link, we not only provision their personal space (guest users don't have one) but the invite they followed (the creat home call can add the grant). Invites are actually not a new concept. IMO it is just a better name for a pending share (they have three states: pending, accepted and declined).
When sharing with 'internal' users the invite service can take the responsibility of creating grants and accepting them instead of creating invites.

Anyway, IIRC we bounced around the idea of exchanging the sub+iss, basic auth or app password credentials in the proxy with a userid generated by opencloud years ago already. It has seeveral benefits:
1. we can assign credentials and multiple identities (as in multiple Identity Providers) to an account. This allows [migrating accounts from one IdP to another](https://openid.net/wordpress-content/uploads/2016/08/draft-account-migration-01.html) as the grants on disk can keep the same userid.
2. we can add a `scim_id` as an identitiy, which would make some use cases follow the standard integration of [SCIM and OpenID Connect](https://openid.net/specs/openid-connect-scim-profile-1_0.html) we could use SCIM to provision a guest account in the Identity management system.

The last point addresses the problem that if a user shares with a guest, aka an email address, that needs to trigger an onboarding process for the new guest. If we just create a new user in a keycloak that we have write permission to we are back at the same shadow it user management as before.

We could use [OpenID Connect Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html) to find the external issuer and trust that to authenticate users. OIDC in theory is federated. However, in practice our clients would have to [dynamically register with the guests IdP](https://openid.net/specs/openid-connect-registration-1_0.html) ... which does not seem to be widely supported, yet.

I think we should use a list of trusted identity providers, this would allow a single instance to use multiple identity providers, eg for multi tenancy use cases or when organization merge and multiple idps exist for a wile or to better reflect the sovereignty of organizations.

If no idp is responsible for the guest email, we can use a fallback idp that is only used for guest accounts. We already have the webfinger service that we can use for the issuer discovery.

So ... yes, please ... make opencloud generate a userid.

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+
+- the audit trail of the external user accessing the resource needs to
+  be maintained, that means sharing via a password protected public link
+  is not sufficient as access to that one is tracked as if the creator
+  of the link accessed the resource
+- external users need to be authenticated just like "normal" users, when
+  accessing the shared resource (including the possibility to use 2FA)
+- the ability to invite external users is tied to a separate permission
+  (e.g. "can invite guest users")
+- make it work with all (most) of the user-management configurations we support
```
dragotin: Not for Lico, right?

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+- once a guest user is created, is everyone in the organization able to
+  share with that guest user?
+- what are guest user allowed to do? (are they able to share, lookup
+  other users?)
+- should this be really tied just to creating shares? Or should we have
+  a more generic "invite user" feature that can be used in other
+  contexts as well?
+- What if the user already exists but used a different mail address in
+  his account (e.g. sub-addressing?).
+
```
dragotin, line 71: Can guest users be converted to "regular" users and keep their shares?

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+### UserIDs
+
+- Every user in OpenCloud needs to have a userid assigned
+- Sharing, as many other features, needs that userid for storing the
+  share (share service) and for assigning the grants on the shared
+  resources (storage provider)
+- When an external IDP is used the generation of that userid is usually
+  not in control of OpenCloud (exception User-Autoprovisioning, or when
+  the Provisioning/Education API is used), but where taking the userid
+  from some LDAP Attribute maintained in the external system
```
dragotin:
```suggestion
- When an external IDP is used, the generation of that userid is usually
  not in control of OpenCloud (exception User-Autoprovisioning, or when
  the Provisioning/Education API is used). In that case, the userid is taken
  from a LDAP Attribute maintained in the external system
```

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+- OpenCloud provides many different ways to consume user-accounts. Guest
+  users are supposed to be working with all/most of them:
+  - Internal IDP, with internal LDAP service
+  - External IDP, with external LDAP service
+  - External IDP, with manual provisioning via the
+    Education/Provisioning APIs (to a local OpenCloud specific LDAP
+    service) - e.g. in multi-tenant setups
+  - External IDP, with User-Autoprovisioning (also to a local OpenCloud
+    specific LDAP service)
+  - everything in-between and outside of the above
```
dragotin, line 95: I suggest to explicitly not support the internal IDP

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Re-vitalize the PoC implementation of the invitations service and finalize it (<https://github.com/opencloud-eu/opencloud/blob/main/services/invitations/README.md>)
+
+- Implements parts of the MSGraph Invitation Specification
+  (<https://learn.microsoft.com/en-us/graph/api/resources/invitation?view=graph-rest-1.0>)
+- Currently there's just a single backend that allows creating users,
+  using the Keycloak Admin API
+- As part of the user creation keycloak triggers an email to be sent to
+  the invited user to get him to verify his email address and set a
+  password. This is not really and invitation email.
```
dragotin, line 115: That is not enough. We need to be in control of that EMail probably.

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+
+    - somewhat complex
+    - likely requires changes to the CS3 APIs
+
+#### Implementation Obstacles
+
+    - Permissions on spaces are currently not tracked in the share
+      manager, the are purely managed via grants. So currently the share
+      manager service currently does not know anything about (invited)
+      users being assigned to spaces
```
dragotin, line 190: Why does the share manager need to know about invited or even accepted guest users? Is't that just a second call to the invitation manager to get that info where needed?

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Additional thoughts
+
+If OpenCloud were responsible for allocating the UserIDs of all users
+the solution sketch above would likely loose some of its complexity. We
+would "roll" the userid for the invited user already when creating the
+invite. And could skip the step of creating a "pending" Share with an
+invitation assigned. As we have an ID already we could just create a
+"normal" share an even populated the grants on the filesystem for that
+share (or space)
```
dragotin:
```suggestion
invite. That would allow to skip the step of creating a "pending" Share with an
invitation assigned. As we have an ID already, we could just create a
"normal" share and even populate the grants on the filesystem for that
share (or space)
```

docs/adr/0004-guest-users.md
rhafer: Ideally yes. I guess we should add that to the requirements.

docs/adr/0004-guest-users.md
rhafer: ok

docs/adr/0004-guest-users.md
rhafer: What exactly do you mean by that?

docs/adr/0004-guest-users.md
rhafer: > Is't that just a second call to the invitation manager to get that info where needed?

Could you please elaborate on that?

docs/adr/0003-guest-users.md
tbsbdr: fine for me.

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  (e.g. "can invite guest users")
+- make it work with all (most) of the user-management configurations we support
+- avoid creating "Shadow IT" Infrastructure, e.g. we don't want to
+  create/maintain a separate IDP instance just for Guest User that would
+  allow bypassing corporate rules for Identity Management
+
+## Questions still to be answered
+
+- what's the life cycle of a guest user?
+  - Who's responsible for deprovisioning?
```
tbsbdr, line 65: I suggest users with the role "admin"

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  - Do guest users expire after a certain time?
```
tbsbdr, line 66: default: guest user accounts expire after 120 days of inactivity (=no login), can be configured. the invitation expires after 30 days per default. can be configured.

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  - Do we need to keep track of who invited whom and when? (not just in
```
tbsbdr, line 67: yes. there needs to be a lowlevel option to manage guest users eg. via cmd.
Guest user management must also be possible via a dashboard which is by default an external system with these options in the ui (same options should be availbale via cmd):

list all email adresses:
- email adress (=guest user)
- status (invitation accepted / pending) 
- invited by
- invitation date
- if accepted: invitation accepted date
- last login
- days left until expiration + absolute date (updates after every login) 

actions:
- export list as CSV
- deactivate/activate login (triggers infomail to guestuser)
- delete invite/user (triggers infomail to guestuser)

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  allow bypassing corporate rules for Identity Management
+
+## Questions still to be answered
+
+- what's the life cycle of a guest user?
+  - Who's responsible for deprovisioning?
+  - Do guest users expire after a certain time?
+  - Do we need to keep track of who invited whom and when? (not just in
+    the audit log?)
+- who can see the list of guest users?
```
tbsbdr: list of all guest users: users with the role oc-admin

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+- once a guest user is created, is everyone in the organization able to
```
tbsbdr: yes

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  share with that guest user?
+- what are guest user allowed to do? (are they able to share, lookup
```
tbsbdr: - guestusers **can not share** or invite other users to a space or create public links. (primary focus of the feature is to provide a simple way to grant external, authorized access. anything else like resharing would undermine regular user accounts).
- ~the default should be, that guestusers **can not see who else has access**.~ postponed for now
- guestusers **can use the desktop and mobile** client to access their shares or spaces
- ~guest users **can not get the role "can manage" spaces**~ postponed for now

docs/adr/0003-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+  other users?)
+- should this be really tied just to creating shares? Or should we have
```
tbsbdr: guest users should get access to the system via a **share** or **space membership**.
could you give some examples what you mean with "other contexts"?

docs/adr/0004-guest-users.md
tbsbdr: Ideally yes.

docs/adr/0004-guest-users.md
tbsbdr: yep. I consider the internal IDP to become a legacy system in the next years, as long as it does not support MFA. Having no MFA or other strong authentication like passkeys will become a nogo in the next years i guess.

docs/adr/0004-guest-users.md
```diff
@@ -0,0 +1,192 @@
...
+    specific LDAP service)
+  - everything in-between and outside of the above
+- Each of these options have different ways for user-provisioning and in
+  the way userids are generated and managed
+
+### How do we keep track of invitations?
+
+- Completely rely on external system?
+- Track creation and acceptance of invitations somehow?
+- Do invitation expire at some point?
```
tbsbdr, line 103: By default invitations expire after 30 days. can be configured

docs/adr/0003-guest-users.md
tbsbdr: there must be an option on the keycloak selfregistration form to accept the "terms of service"
acceptance date per user must be logged

docs/adr/0004-guest-users.md
rhafer: :heavy_check_mark:

docs/adr/0004-guest-users.md
rhafer: > default: guest user accounts expire after 120 days of inactivity (=no login), can be configured

This is where the whole thing gets murky again. I don't think we can reliably provide such a feature and keep the promise of working with all kinds external IDPs. Locking inactive user accounts is usually a feature of the IDP, we're not an IDP.

> the invitation expires after 30 days per default. can be configured.

docs/adr/0004-guest-users.md
rhafer: > * deactivate/activate login (triggers infomail to guestuser)
> 
> * delete invite/user (triggers infomail to guestuser)

See my other comment above. In a "normal" setup with an external IDP the OpenCloud admin is not able to delete users.  As "guest" users are supposed to come from the same IDP, we will also not be able to delete those.

Somehow I think we need to get a common understand of the whole "lifecycle of an OpenCloud user". I have the feeling everybody has a different understanding of what OpenCloud really should provide in that area.

docs/adr/0003-guest-users.md
rhafer: > guest users should get access to the system via a **share** or **space membership**. could you give some examples what you mean with "other contexts"?

Mainly I found it strange initially that the process of inviting a guest is supposed to be triggered only from the `create share` dialog. But meanwhile I think that is totally fine.

docs/adr/0003-guest-users.md
rhafer: > * guestusers **can not share** or invite other users to a space or create public links. (primary focus of the feature is to provide a simple way to grant external, authorized access. anything else like resharing would undermine regular user accounts).
> 
>     * ~the default should be, that guestusers **can not see who else has access**.~ postponed for now
> 
>     * guestusers **can use the desktop and mobile** client to access their shares or spaces
> 
>     * ~guest users **can not get the role "can manage" spaces**~ postponed for now

I've added those. What about looking up or search for other users? Currently we don't have any restriction on the respective graph API endpoints regarding that.

docs/adr/0004-guest-users.md
pbleser-oc: I would strongly assume that that means the content and the branding of the email.

rhafer:
We've agreed to postpone this for now in favour of a more lightweight approach (#2513). I've added a note to the document accordingly so that we can still merge this PR (even if only for historic reasons).

PR #2194, opened 2026-01-21
show tracing link

ScharfViktor:
image

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -1748,7 +1748,6 @@ def uploadTracingResult(ctx):
     status = ["failure"]
     if "with-tracing" in ctx.build.title.lower():
         status = ["failure", "success"]
-    S3_HOST = s3.ci.opencloud.eu
 
     return [{
         "name": "upload-tracing-result",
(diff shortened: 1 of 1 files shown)


.woodpecker.star
```diff
@@ -434,6 +434,7 @@ MINIO_MC_ENV = {
         "from_secret": "cache_s3_secret_key",
     },
     "PUBLIC_BUCKET": "public",
+    "MC_PUBLIC_HOST": "https://s3.ci.opencloud.eu",
```
saw-jan: is it working? I would expect the same result because if the url is in the secret, the ci will try to hide it.

v-scharf: hm, not sure
okay, second way in the next commit

.woodpecker.star
```diff
@@ -1758,7 +1758,7 @@ def uploadTracingResult(ctx):
             "mc cp -a %s/reports/e2e/playwright/tracing/* s3/$PUBLIC_BUCKET/web/tracing/$CI_REPO_NAME/$CI_PIPELINE_NUMBER/" % dirs["web"],
             "cd %s/reports/e2e/playwright/tracing/" % dirs["web"],
             'echo "To see the trace, please open the following link in the console"',
-            'for f in *.zip; do echo "npx playwright show-trace $MC_HOST/$PUBLIC_BUCKET/web/tracing/$CI_REPO_NAME/$CI_PIPELINE_NUMBER/$f \n"; done',
+            'for f in *.zip; do echo "npx playwright show-trace $CACHE_S3_SERVER/$PUBLIC_BUCKET/web/tracing/$CI_REPO_NAME/$CI_PIPELINE_NUMBER/$f \n"; done',
```
saw-jan: you can try running these commands in one of the workflow step to check if it works or not
```
"echo $MC_HOST",
"echo $CACHE_S3_SERVER",
"echo 'https://s3.ci.opencloud.eu'",
```

v-scharf: you're right:
image

saw-jan: is the server_url secret repo level or org level?. if the secret is not used then we can try removing it

saw-jan: env, variable and plain text didn't work:
```
+ echo $MC_HOST
https://********
+ echo https://********
https://********
+ echo 'https://********'
https://********
+ buildifier --mode=check .woodpecker.star
```

v-scharf: If the URL matches the secret 1-to-1, it cannot be displayed in the logs in plain text.

The only option is to remove the URL from the secret. @flimmy could you remove s3.ci.opencloud.eu from secret?

PR #2215, opened 2026-01-26
reva-bump-2.42.0

ScharfViktor:
bump reva for upcoming rolling release


v-scharf:
hm `tests/e2e/cucumber/features/spaces/participantManagement.feature:3` is failed third time

public downloads file and file hangs on postrocessing state. If test users try to do something with file -> they get error

see here
https://github.com/opencloud-eu/reva/issues/505#issuecomment-3798512860

I couldn't reproduce it localy

Issue #1933, opened 2025-11-27
dev docs list environment variabls as introduced in version 'next'.


jnweiger:
## Describe the bug
https://docs.opencloud.eu/docs/dev/server -> Services
has a section 'Environment Variables' in each of the listed services.

These services have one or more environment variables that say 'next' for their introduction version:
- ActivityLog
- Collaboration
- Graph
- Postprocessing
- Search
- Storage-users
- Users
- Web

## Expected behavior
Pleas review all occurrences of next, and substitute with a concrete version number.

## Actual behavior
It is not clear, if next means the current 4.0.0 release or something later or earlier.

micbar:
Correct: that depends on #1926

micbar:
Then we need https://github.com/opencloud-eu/markdown-docs-generator/pull/6

micbar:
And a new run of the docs generator.

@Svanvith @schweigisito

dragonchaser:
@micbar the markdown generator does not replace the `%%NEXT%%` and `%%NEXT_PROD%%` placeholders, this needs to be done BEFORE manually in the opencloud code there is NO automation for that. After that the docs pipeline needs to be run, this will resolve the issue. This is not a bug.

dragonchaser:
We need a pipeline step that greps for `%%NEXT%%` and `%%NEXT_PROD%%` that fails if such tokens are still present in the opencloud code.

@individual-it I think we had this before?

individual-it:
@dragonchaser I cannot recall having that ever in CI, but should be relatively easy to implement

individual-it:
@micbar @dragonchaser is there anything left to do in this issue, or should we close it?

PR #2236, opened 2026-01-29
ci: check for version placeholders

individual-it:
## Description
check for placeholders `%%NEXT%%` and `%%NEXT_PRODUCTION_VERSION%%`  and fail CI if there are still in the code

## Related Issue
part of https://github.com/opencloud-eu/opencloud/issues/1933#issuecomment-3816466374

## Motivation and Context
make sure the placeholders are replaced with correct versions before the release

## How Has This Been Tested?
:robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -523,6 +523,7 @@ def main(ctx):
         testPipelines(ctx)
 
     build_release_pipelines = \
+        checkVersionPlaceholder() + \
         dockerReleases(ctx) + \
         binaryReleases(ctx)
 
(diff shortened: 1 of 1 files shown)


.woodpecker.star
```diff
@@ -1817,6 +1818,27 @@ def dockerReleases(ctx):
...
+                    "grep -r -e '%%NEXT%%' -e '%%NEXT_PRODUCTION_VERSION%%' %s/services %s/pkg > next_version.txt" % (
+                        dirs["base"],
+                        dirs["base"]
+                    ),
+                    'if [ -s next_version.txt ]; then echo "replace version placeholders"; cat next_version.txt; exit 1; fi',
+                ],
+            }
+        ],
+        "when": [
+            event["pull_request"],
```
individual-it: when all works this should be changed to `event["tag"]`

micbar:
@dragonchaser @individual-it 
Do we want to drop that completely?

I don't understand the current discussion. Maybe i missed something.

Wasn't it the idea to use these placeholders because we do not know the next version string ATM?

The replacement check should be done closely before the release, when we already know the next version.

PR #2253, opened 2026-02-02
adr: Document OIDC client parameter discovery

rhafer:
This basically documents what as discussed in https://github.com/opencloud-eu/opencloud/pull/2072 and https://github.com/opencloud-eu/internal/issues/172

I am mainly adding to this to keep a record of why we're doing it and how.

@kulmann To make all clients use the same mechanism for discovering the oidc parameter it would IMO be good if web could switch to as well. However, looking at the other OIDC related parameters that can be configured in web. I have some questions.

- `WEB_OIDC_METADATA_URL`: This allows to set the URL for the `.well-known/openid-configuration` endpoint. Do we really need that? This should always be the issuer-url as returned in the `http://openid.net/specs/connect/1.0/issuer` relation + `.well-known/openid-configuration`.  The OIDC spec even define is as a `MUST` (https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig). 
- `ResponseType`: We default to `code` here, have we ever encountered the need to set something else? To we need to have this configurable?
- `PostLogoutRedirectURI`: I guess this is really specifc to web? So we should probably keep it in the config.json

### DIFF
--- a/docs/adr/0003-oidc-client-config-discovery.md
+++ b/docs/adr/0003-oidc-client-config-discovery.md
@@ -2,7 +2,7 @@
 title: "Discover OIDC Client configuration via WebFinger"
 ---
 
-* Status: pending
+* Status: accepted
 * Deciders: [@TheOneRing @kulmann @rhafer @dragotin]
 * Date: 2026-02-02
 


TheOneRing:
lgtm

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+## Context and Problem Statement
+
+Up to now our client applications used hard-coded OIDC client configurations.
+So it is not possible to change the client id that a client should use or the
+list of scopes that a client needs to request. This makes it hard to integrate
+OpenCloud with various existing identity providers. For example:
+
+- Authentik basically creates a different issuer URL for each client. As OpenCloud
+  can only work with a single issuer URL, all OpenCloud clients need to use the
+  same client id to work with Authetnik.
```
kulmann:
```suggestion
  same client id to work with Authentik.
```

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+- Some IDPs (kanidm) are not able to work with user-supplied client ids. They generate
+  client ids automatically and do not allow to specify them manually.
+- To make features like automatic role assignment work, clients need to request
+  specific scopes, depending on which exact IDP is used.
+
+## Decision Drivers
+
+* Support broader set of IDPs
+* Do required the user got configure anything additional on the client side
```
kulmann: I don't understand this sentence 🙈

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Decision
+
+Enhance the WebFinger service in OpenCloud to provide platform-specific OIDC
+discovery, enabling clients to query for the correct OIDC `client_id` and
+`scopes` based on their application type (e.g., web, desktop, android, ios).
+
+This is achieved by allowing and additional `platform` query parameter to be used
```
kulmann:
```suggestion
This is achieved by allowing an additional `platform` query parameter to be used
```

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Example
+
+### Client Request
+
+```
+GET /.well-known/webfinger?resource=https://cloud.opencloud.test&rel=http://openid.net/specs/connect/1.0/issuer&platform=desktop
+```
+
+### Example Response
```
kulmann:
```suggestion
### Server Response
```

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+    "rel": "http://openid.net/specs/connect/1.0/issuer",
+    "href": "https://idp.example.com"
+  }],
+  "properties": {
+    "http://opencloud.eu/ns/oidc/client_id": "desktop-client-id",
+    "http://opencloud.eu/ns/oidc/scopes": ["openid", "profile", "email", "offline_access"]
+  }
+}
+```
+
```
kulmann, line 67: How would the server configuration for this look like? Especially since every platform might differ slightly.

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,67 @@
...
+Up to now our client applications used hard-coded OIDC client configurations.
+So it is not possible to change the client id that a client should use or the
+list of scopes that a client needs to request. This makes it hard to integrate
+OpenCloud with various existing identity providers. For example:
+
+- Authentik basically creates a different issuer URL for each client. As OpenCloud
+  can only work with a single issuer URL, all OpenCloud clients need to use the
+  same client id to work with Authetnik.
+- Some IDPs (kanidm) are not able to work with user-supplied client ids. They generate
+  client ids automatically and do not allow to specify them manually.
```
kulmann, line 22: So how does OpenCloud get the information which client id to propagate to e.g. the Android app then?

micbar: This will be a Server configuration variable

docs/adr/0003-oidc-client-config-discovery.md
rhafer: Me neither 😄

rhafer: fixed.

docs/adr/0003-oidc-client-config-discovery.md
rhafer: I've added the new config settings here. Will also adapt the implementation to allow the fallbacks.

kulmann: Ah, so they're generated but then fixed. Ok :-)

docs/adr/0003-oidc-client-config-discovery.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Server configuration (suggestion)
+
+To configure the OpenCloud server a couple of new config settings need to be introduced. This would
+be two new settings per client, e.g.:
+
+
+```
+WEBFINGER_ANDROID_OIDC_CLIENT_ID
+WEBFINGER_ANDROID_OIDC_CLIENT_SCOPE
```
kulmann, line 76: Why isn't it pluralized (`SCOPES`)? 🤔

rhafer: It actually is in the implementation. See #2072. Forgot to change it back here. :man_facepalming:

rhafer: Fixed in #2353

Issue #2252, opened 2026-02-02
Unable to increase OIDC userinfo cache TTL


mjnaderi:
I believe the default OIDC userinfo cache TTL (10 seconds) is too short, and I’d like to increase it.

According to the documentation, the `PROXY_OIDC_USERINFO_CACHE_TTL` environment variable is only applied when the access token does **not** have an expiration time. What is the reasoning behind this restriction? In our setup, the access token *does* have an expiration, which means we cannot increase the userinfo cache TTL at all.

We have two concrete reasons for wanting to increase the TTL:

**1. Random user logouts**
We are experiencing the same issue described in #1493, where users get logged out seemingly at random. One suggested workaround to mitigate this problem is increasing the OIDC userinfo cache TTL via `PROXY_OIDC_USERINFO_CACHE_TTL`, but this is not possible in our case due to the restriction mentioned above.

**2. Excessive requests to the IdP userinfo endpoint**
With a TTL of only 10 seconds, OpenCloud sends a large number of requests to the IdP’s userinfo endpoint. Our external IdP enforces rate limiting, and when many users are active at the same time, the userinfo endpoint starts returning `429 Too Many Requests`, which negatively impacts user sessions.

Given these issues, it would be very helpful to understand the logic behind tying the userinfo cache TTL to the absence of an access token expiration, and whether this behavior could be reconsidered or made configurable.

Here is the configuration related to our external IdP:

```yaml
---
services:
  opencloud:
    environment:
      FRONTEND_READONLY_USER_ATTRIBUTES: "user.onPremisesSamAccountName,user.displayName,user.mail,user.passwordProfile,user.accountEnabled" # noqa yaml[line-length]
      PROXY_OIDC_REWRITE_WELLKNOWN: "true"
      WEB_OIDC_CLIENT_ID: ${OC_OIDC_CLIENT_ID:-web}
      PROXY_ROLE_ASSIGNMENT_DRIVER: "default"
      OC_OIDC_ISSUER: ${IDP_ISSUER_URL}
      # Start all services except idp. It is replaced by external service.
      OC_EXCLUDE_RUN_SERVICES: idp
      # IdP specific configuration for auto-provisioning
      OC_LDAP_SERVER_WRITE_ENABLED: "true"
      PROXY_AUTOPROVISION_ACCOUNTS: "true"
      # Use the `sub` claim from the IdP for the user ID
      # It must be a unique and stable identifier that doesn’t change over time.
[...]

PR #2256, opened 2026-02-03
docs(proxy): Clarify PROXY_OIDC_USERINFO_CACHE_TTL value

rhafer:
Try to make it more precise when that value is actually relevant.

Closes: #2252


rhafer:
There seems to be some heavy flakyness in the `test-Core-API-3` suite. Every run is failing with a different error. But it can't be related to this PR. Merging anyways.

PR #2285, opened 2026-02-05
docs(search): add note about re-indexing when enabling Tika

### LINKED ISSUE #217: How to verify tika is indexing files
I added some files (PDFs, ZIP files and many others) to my opencloud instance via the windows sync client (mainly open source modular synthesizer schematics and documentation)

I wanted to search for a text "TL072" to find a PDF that mentions that chip

No results were returned.

Then I saw that tika was not set up

So I edited .env and added `:search/tika.yml` and ran `docker compose up -d` again 
Tika server started successfully. 

I waited for about 10 minutes but I could neither see a substantial CPU load (Intel G5420 CPU - should be quite busy when indexing, right)  nor find the document

michaelstingl:
## Description

Add notes to the search service README clarifying that:

1. Enabling Tika does not automatically re-extract content from already indexed files
2. The `opencloud search index --all-spaces` command skips files with unchanged modification time
3. Workaround: delete the Bleve search index before re-indexing to force full content extraction

## Related Issue

- Fixes opencloud-eu/opencloud-compose#217

## Motivation and Context

When users enable Tika on an existing instance, they expect full-text search to work for all files. However, `opencloud search index --all-spaces` skips files already in the index (mtime-based check in `services/search/pkg/search/service.go`), so the Tika extractor is never called for previously indexed files. This is undocumented and confusing.

## How Has This Been Tested?

- test environment: Read the source code in `services/search/pkg/search/service.go` (IndexSpace method, mtime skip logic at line ~495)
- test case 1: Verified the skip behavior by tracing the code path: IndexSpace → Walk → mtime check → skip if already indexed
- test case 2: Confirmed no `--force` flag exists in the CLI or protobuf definition (`IndexSpaceRequest`)

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation added

### DIFF
--- a/services/search/README.md
+++ b/services/search/README.md
@@ -104,6 +104,8 @@ Additionally, the following optional settings can be set:
 
 *   `SEARCH_EXTRACTOR_TIKA_CLEAN_STOP_WORDS=true` (default: `true`): ignore stop words like `I`, `you`, `the` during content extraction.
 
+> **Note:** Enabling Tika does not automatically re-extract content from already indexed files. You need to delete the existing search index and trigger a full re-index. See [Manually Trigger Re-Indexing a Space](#manually-trigger-re-indexing-a-space) for details.
+
 ## Manually Trigger Re-Indexing a Space
 
 The service includes a command-line interface to trigger re-indexing a space:
@@ -117,6 +119,13 @@ It can also be used to re-index all spaces:
 ```shell
 opencloud search index --all-spaces
 ```
+
+> **Note:** The re-index command skips files whose modification time has not changed since they were last indexed. If you changed the extractor type (e.g., from `basic` to `tika`), you need to delete the existing search index first to force a full content re-extraction:
+>
+> ```shell
+> rm -rf $OC_BASE_DATA_PATH/search   # default: /var/lib/opencloud/search
+> opencloud search index --all-spaces
+> ```
 
 ## Metrics
 


services/search/README.md
```diff
@@ -118,6 +120,13 @@ It can also be used to re-index all spaces:
 opencloud search index --all-spaces
 ```
 
+> **Note:** The re-index command skips files whose modification time has not changed since they were last indexed. If you changed the extractor type (e.g., from `basic` to `tika`), you need to delete the existing search index first to force a full content re-extraction:
```
micbar, line 123: @ScharfViktor Is that true? I think we need to verify this first.

michaelstingl:
I can do an exact reproducer the next few days

services/search/README.md
v-scharf: yes, that is true. I can reproduce it

services/search/README.md
```diff
@@ -118,6 +120,13 @@ It can also be used to re-index all spaces:
...
+>
+> ```shell
+> rm -rf $OC_BASE_DATA_PATH/search   # default: /var/lib/opencloud/search
```
v-scharf, line 126: @micbar maybe it is bug? I expect re-index without deleting `/search`

micbar: @aduffeck can you clarify? You know the implementation

aduffeck: That's the current behavior, yes. I consider that a bug.

A re-index should unconditionally rebuild the index for the space/all space in my opinion. Maybe it would be helpful to also have a command or flag for just "syncing" the index, i.e. picking up changes that haven't been indexed yet (the current behavior), but that shouldn't be the default behavior of the index command.

michaelstingl: Any progress on fixing the re-index behavior? Happy to close this draft once the code fix lands.

Issue #350, opened 2025-03-11
Fix outdated documentation URLs in Makefile


michaelstingl:
## Description

The Makefile contains several references to documentation URLs that point to `https://opencloud.dev/opencloud/development/...`, which appears to be outdated or non-existent. These references were added during the debranding process from owncloud/ocis to opencloud in January 2025 (commit 6a84a45c0).

Current documentation is now hosted at `https://docs.opencloud.eu/` as confirmed in issue #265, and the OpenAPI documentation is available at `https://docs.opencloud.eu/swagger/libre-graph-api/` according to issue #232.

## Current situation

In `Makefile`, lines 88, 97, 102, and 107 contain references to outdated URLs:

Line 88:
```
@echo -e "${PURPLE}\tdocs: https://opencloud.dev/opencloud/development/testing/#testing-with-test-suite-natively-installed${RESET}\n"
```

Line 97:
```
@echo -e "${PURPLE}\tdocs: https://opencloud.dev/opencloud/development/build/${RESET}\n"
```

Line 102:
```
@echo -e "${PURPLE}\tdocs: https://opencloud.dev/opencloud/development/build-docs/${RESET}\n"
```

Line 107:
```
@echo -e "${PURPLE}\tdocs: https://opencloud.dev/opencloud/development/testing/#testing-with-test-suite-in-docker${RESET}\n"
```

These links are shown to users in the Makefile help output and can cause confusion since they point to non-existing resources.

## Proposed solution

Update these references to point to the correct documentation URLs at `https://docs.opencloud.eu/` or remove specific URLs if the exact path structure is not yet established.

For example:
```
@echo -e "${PURPLE}\tdocs: https://docs.opencloud.eu/development/testing/#testing-with-test-suite-natively-installed${RESET}\n"
```

Or a more generic approach if paths aren't finalized:
```
@echo -e "${PURPLE}\tdocs: See https://docs.opencloud.eu/ for documentation${RESET}\n"
```

## Additional context

A documentation CI/CD pipeline was set up in February 2025 (PR #233) that deploys to `docs.opencloud.eu`, but the Makefile references weren't updated at that time.

michaelstingl:
Here are the code references that should be included in the issue:

## Current problematic URLs in Makefile:

Line 88:


Line 97:


Line 102:


Line 107:


## Proposed replacement examples:

Option 1 - Update to new domain with similar paths:


Option 2 - Generic reference:

michaelstingl:
[...]

PR #2193, opened 2026-01-21
fix url typo

dragonchaser:
fix url typo

refs #350

PR #2304, opened 2026-02-09
fix: make file urls

ScharfViktor:
fixed https://github.com/opencloud-eu/opencloud/issues/350

image


v-scharf:
@dragonchaser I didn't correct yours PR #2193 and created new. I checked -> all links open and lead to the right place

PR #2328, opened 2026-02-12
resolve logout token subject:sessions for the idp backchannel logout

dragonchaser:
## Description

the backchannel logout only considers the logout token sessionId, as mentioned by the spec,
the sessionId is optional if the subject exists.

The current implementation ignores the subject and fails if the sessionId is not part of the logoutToken.
A more detailed description of the problem and how to reproduce it can be found [here](https://github.com/opencloud-eu/internal/issues/246).

## Related Issue

- part of https://github.com/opencloud-eu/internal/issues/246

## Motivation and Context

from the spec:

> A Logout Token MUST contain either a sub or a sid Claim, and MAY contain both. If a sid Claim is not present, the intent is that all sessions at the RP for the End-User identified by the iss and sub Claims be logged out.

## Known side-effects

- keyCloak "Sign out all active sessions" does not send a backchannel logout request,
  as the devs mention, this may lead to thousands of backchannel logout requests,
  therefore, they recommend a short token lifetime.
  https://github.com/keycloak/keycloak/issues/27342#issuecomment-2408461913

- keyCloak user self-service portal, "Sign out all devices" may not send a backchannel
  logout request for each session, it's not mentionex explicitly,
  but maybe the reason for that is the same as for "Sign out all active sessions"
  to prevent a flood of backchannel logout requests.

- if the keyCloak setting "Backchannel logout session required" is disabled (or the token has no session id),
  we resolve the session by the subject which can lead to multiple session records (subject.*),
  we then send a logout event (sse) to each connected client and delete our stored cache record (subject.session & claim).
  all sessions besides the one that triggered the backchannel logout continue to exist in the identity provider,
  so the user will not be fully logged out until all sessions are logged out or expired.
  this leads to the situation that web renders the logout view even if the instance is not fully logged out yet.

## Things to pay special attention to during the review

- `services/proxy/pkg/middleware/oidc_auth.go`
  - [x] format of the cache key
    - [x] `{key: ".sessionId"}` => used for `*.sessionId`
    - [x] `{key: "subject.sessionId"}`  => used for `subject.*`
- `services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go`
  - [x] suffix key matching and lookup
  - [x] prefix key matching and lookup
  - [x] suspicious query detection
- `services/proxy/pkg/staticroutes/backchannellogout.go`
  - [x] error handling (all cases and return values are ok?)

## ToDos:

- the spec mentions `by the iss and sub Claims` ..., we don't do that?
- we use the memory cache store by default, should we switch it to nats?

## Questions

- why don't we just store the claim records in the cache and use it for everything?

## How Has This Been Tested?

- unit tests
- local keyCloak installation & multiple clients & multiple cache stores (memory, nats-kv)

## Types of changes

- [x] Technical debt

## Checklist:

- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
+++ b/services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
@@ -68,21 +68,18 @@ var ErrInvalidSubjectOrSession = errors.New("invalid subject or session")
 // NewSuSe parses the subject and session id from the given key and returns a SuSe struct
 func NewSuSe(key string) (SuSe, error) {
 	suse := SuSe{}
-	switch keys := strings.Split(strings.Join(strings.Fields(key), ""), "."); {
-	// key: '.session'
-	case len(keys) == 2 && keys[0] == "" && keys[1] != "":
-		suse.encodedSession = keys[1]
-	// key: 'subject.'
-	case len(keys) == 2 && keys[0] != "" && keys[1] == "":
-		suse.encodedSubject = keys[0]
-	// key: 'subject.session'
-	case len(keys) == 2 && keys[0] != "" && keys[1] != "":
-		suse.encodedSubject = keys[0]
-		suse.encodedSession = keys[1]
-	// key: 'session'
-	case len(keys) == 1 && keys[0] != "":
+	keys := strings.Split(key, ".")
+	switch len(keys) {
+	case 1:
 		suse.encodedSession = keys[0]
+	case 2:
+		suse.encodedSubject = keys[0]
+		suse.encodedSession = keys[1]
 	default:
+		return suse, ErrInvalidSubjectOrSession
+	}
+
+	if suse.encodedSubject == "" && suse.encodedSession == "" {
 		return suse, ErrInvalidSubjectOrSession
 	}
 
@@ -134,13 +131,13 @@ func GetLogoutRecords(suse SuSe, store microstore.Store) ([]*microstore.Record, 
 
 	var key string
 	var opts []microstore.ReadOption
-	switch mode {
-	case logoutModeSubject:
+	switch {
+	case mode == logoutModeSubject && suse.encodedSubject != "":
 		// the dot at the end prevents prefix exploration in the cache,
 		// so only keys that start with 'subject.*' will be returned, but not 'sub*'.
 		key = suse.encodedSubject + "."
 		opts = append(opts, microstore.ReadPrefix())
-	case logoutModeSession:
+	case mode == logoutModeSession && suse.encodedSession != "":
 		// the dot at the beginning prevents sufix exploration in the cache,
 		// so only keys that end with '*.session' will be returned, but not '*sion'.
 		key = "." + suse.encodedSession


services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -25,6 +26,12 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	if r.PostFormValue("logout_token") == "" {
+		logger.Warn().Msg("logout_token is missing")
+		render.Status(r, http.StatusBadRequest)
+		render.JSON(w, r, jse{Error: "invalid_request", ErrorDescription: "logout_token is missing"})
```
fschade:
```go
...
return
```

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -33,6 +40,12 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	if strings.TrimSpace(logoutToken.SessionId) == "" && strings.TrimSpace(logoutToken.Subject) == "" {
+		logger.Warn().Msg("invalid logout token")
+		render.Status(r, http.StatusBadRequest)
+		render.JSON(w, r, jse{Error: "invalid_request", ErrorDescription: "invalid logout token"})
```
fschade:
```go
...
return
```

fschade:
toDo:

- using nats kv by default is ok?
- what should we do if the token contains a session id and a subject
- check comment toDos

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -33,51 +42,139 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
...
+			return nil, false
+		}
+
+		return records, true
 	}
-	if err != nil {
-		logger.Error().Err(err).Msg("Error reading userinfo cache")
+
+	// the map will contain the record references that are needed for the logout.
+	// - nil value means: subject or claim record
```
micbar: i understand what a `subject` and `sid` record is. But what is a `claim` record?

fschade: > i understand what a `subject` and `sid` record is. But what is a `claim` record?

the claim record contains the claim as []byte, e.g.:
https://github.com/dragonchaser/opencloud/blob/1e53cc92969ca56b3071211e189405615d398ced/services/proxy/pkg/middleware/oidc_auth.go#L110

the value is needed for the backchannel logout event for GetUserByClaims, e.g.:
https://github.com/dragonchaser/opencloud/blob/1e53cc92969ca56b3071211e189405615d398ced/services/proxy/pkg/staticroutes/backchannellogout.go#L160

the existing token cant be used because one subject could contain multiple sessions with one token each....

fschade:
A few minor things here and there, but it should be fine for now.


https://github.com/user-attachments/assets/69d78035-1db9-41d8-a780-3e0e11624a0d

fschade:
i updated the issue description, hope it makes sense

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -114,16 +115,26 @@ func (m *OIDCAuthenticator) getClaims(token string, req *http.Request) (map[stri
...
-				// reuse user cache for session id lookup
-				err = m.userInfoCache.Write(&store.Record{
-					Key:    sid,
-					Value:  []byte(encodedHash),
-					Expiry: time.Until(expiration),
-				})
-				if err != nil {
-					m.Logger.Error().Err(err).Msg("failed to write session lookup cache")
-				}
+			subject, sessionId := strings.Join(strings.Fields(aClaims.Subject), ""), strings.Join(strings.Fields(aClaims.SessionID), "")
```
rhafer: What is this supposed to achieve? Remove whitespaces in the Subject and SessionID? Why?

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -114,16 +115,26 @@ func (m *OIDCAuthenticator) getClaims(token string, req *http.Request) (map[stri
...
+				return
+			}
+
+			// if the claim has no subject, we can leave it empty,
+			// it's important to keep the dot in the key to prevent
+			// sufix and prefix exploration in the cache.
+			//
+			// ok: {key: ".sessionId"}
+			// ok: {key: "subject.sessionId"}
+			key := strings.Join([]string{subject, sessionId}, ".")
```
rhafer: Hm, what happens if the subject and/or sessionid contain a '.' already?

services/proxy/pkg/staticroutes/backchannellogout.go
```diff
@@ -33,8 +57,21 @@ func (s *StaticRouteHandler) backchannelLogout(w http.ResponseWriter, r *http.Re
...
+		logger.Warn().Msg(jseErr.ErrorDescription)
+		render.Status(r, http.StatusBadRequest)
+		render.JSON(w, r, jseErr)
+		return
+	}
+
+	requestSubjectAndSession := bcl.SuSe{Session: session, Subject: subject}
+	// find out which mode of backchannel logout we are in
+	// by checking if the session or subject is present in the token
+	logoutMode := bcl.GetLogoutMode(requestSubjectAndSession)
```
rhafer: Is this call really needed here? Shouldn't it just be done inside `GetLogoutRecords` , AFAICS that's the only place where we need it.

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -114,16 +115,26 @@ func (m *OIDCAuthenticator) getClaims(token string, req *http.Request) (map[stri
...
-					Key:    sid,
-					Value:  []byte(encodedHash),
-					Expiry: time.Until(expiration),
-				})
-				if err != nil {
-					m.Logger.Error().Err(err).Msg("failed to write session lookup cache")
-				}
+			subject, sessionId := strings.Join(strings.Fields(aClaims.Subject), ""), strings.Join(strings.Fields(aClaims.SessionID), "")
+			// if no session id is present, we can't do a session lookup,
+			// so we can skip the cache entry for that.
```
rhafer: Why. If we don't add the cache entry here. How will a backchannel logout work if the logout token just contains a `sub` but no `sid` claim?

Isn't that the actual problem this PR is supposed to solve?

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
```diff
@@ -0,0 +1,134 @@
...
+	LogoutModeSubject
+)
+
+// GetLogoutMode determines the backchannel logout mode based on the presence of subject and session in the SuSe struct
+func GetLogoutMode(suse SuSe) LogoutMode {
+	switch {
+	case suse.Session != "":
+		return LogoutModeSession
+	case suse.Subject != "":
+		return LogoutModeSubject
```
rhafer: What is the behavior if the logout token contains both, a session and a subject? 

Currently we just seem to account for the sesssion and basically ignore the subject? Maybe that's fine the spec isn't really clear about that.

services/proxy/pkg/middleware/oidc_auth.go
fschade: good point... gone

services/proxy/pkg/middleware/oidc_auth.go
fschade: good point, i thought its guaranteed that there is no point... specially since we use nats which uses a dot as separator too... the subject and string is now base64 encoded

services/proxy/pkg/middleware/oidc_auth.go
fschade: good point, its fixed too and now this makes sense: https://github.com/opencloud-eu/web/blob/6680a9b159354b27db79b1bde59f59f95e639119/packages/web-runtime/src/container/sse/common.ts#L4

services/proxy/pkg/staticroutes/backchannellogout.go
fschade: gone

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
fschade: if the session is given, we use only the session and ignore the subject... wich results in 1 logout... any better idea?

fschade:
> I am not quite sure this PR is actually fixing the bug. If we receive a logout with just a `sub` claim. We still need to invalidate all cached userinfo claims for that user, don't we?
> 
> (We might to be able to trigger the Logout Event though)

thanks for your detailed review, you found some good flaws, yes if only a subkect is given, we invalidate all records that belong to that subject.

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
```diff
@@ -0,0 +1,188 @@
...
+	return string(subject), nil
+}
+
+// ErrInvalidSubjectOrSession is returned when the provided key does not match the expected key format
+var ErrInvalidSubjectOrSession = errors.New("invalid subject or session")
+
+// NewSuSe parses the subject and session id from the given key and returns a SuSe struct
+func NewSuSe(key string) (SuSe, error) {
+	suse := SuSe{}
+	switch keys := strings.Split(strings.Join(strings.Fields(key), ""), "."); {
```
rhafer: Why are we doing this Split(Join(Fields())) dance here? An simple `Split(key,".")` should be enough.

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
```diff
@@ -0,0 +1,188 @@
...
+	// key: 'subject.session'
+	case len(keys) == 2 && keys[0] != "" && keys[1] != "":
+		suse.encodedSubject = keys[0]
+		suse.encodedSession = keys[1]
+	// key: 'session'
+	case len(keys) == 1 && keys[0] != "":
+		suse.encodedSession = keys[0]
+	default:
+		return suse, ErrInvalidSubjectOrSession
+	}
```
rhafer, line 80: What do we need all those `case`s for? Is this just enough:
```
keys := strings.Split(key, ".")
switch len(keys) {
case 1:
	suse.encodedSession = keys[0]
case 2:
	suse.encodedSubject = keys[0]
	suse.encodedSession = keys[1]
default:
	return suse, ErrInvalidSubjectOrSession
}
```

Or am I overlooking something?

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
rhafer: I think  that's fine.

fschade:
@dragonchaser im at the team days, can you check @rhafer comment and bring the pr over the finishing line? Thanks 🙏

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
dragonchaser: Your example does not filter out empty values that could be present in the request.

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
dragonchaser: This will also remove invalid whitespaces from the strings.

dragonchaser:
@fschade IMHO it is fine, but somebody who was not involved should add the final thumbs-up

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
fschade: @dragonchaser thats ok keys are base64 now and not uuid anymore

services/proxy/pkg/staticroutes/internal/backchannellogout/backchannellogout.go
fschade: i pushed a change which simplifies that and filters out invalid requests (whitespaces, etc)

PR #2359, opened 2026-02-18
Integrate ocdav into the webdav service

aduffeck:
With this PR the ocdav services moves from reva to the opencloud webdav service, streamlining webdav handling into one service.
This will also allow for persisting favorites without duplication the favorite manager moving forward.


dragonchaser:
When `OC_INSECURE` is not set to true it will fail with 500 errors in dev env.

aduffeck:
Put this on hold for now, it became less important with the changed favorites approach.

PR #2404, opened 2026-02-25
fix(ocm): allow insecure tls for wayf discovery

MahdiBaghbani:
This is related to the https://github.com/opencloud-eu/web/issues/1774 and would help this specific issue https://github.com/opencloud-eu/web/pull/1486#pullrequestreview-3670391220

### DIFF
--- a/services/ocm/pkg/config/config.go
+++ b/services/ocm/pkg/config/config.go
@@ -86,7 +86,7 @@ type ScienceMesh struct {
 	MeshDirectoryURL     string `yaml:"science_mesh_directory_url" env:"OCM_MESH_DIRECTORY_URL" desc:"URL of the mesh directory service." introductionVersion:"1.0.0"`
 	DirectoryServiceURLs string `yaml:"directory_service_urls" env:"OCM_DIRECTORY_SERVICE_URLS" desc:"Space delimited URLs of the directory services." introductionVersion:"3.5.0"`
 	InviteAcceptDialog   string `yaml:"invite_accept_dialog" env:"OCM_INVITE_ACCEPT_DIALOG" desc:"/open-cloud-mesh/accept-invite;The frontend URL where to land when receiving an invitation" introductionVersion:"3.5.0"`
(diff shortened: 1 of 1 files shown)


services/ocm/pkg/config/config.go
```diff
@@ -86,6 +86,7 @@ type ScienceMesh struct {
 	MeshDirectoryURL     string `yaml:"science_mesh_directory_url" env:"OCM_MESH_DIRECTORY_URL" desc:"URL of the mesh directory service." introductionVersion:"1.0.0"`
 	DirectoryServiceURLs string `yaml:"directory_service_urls" env:"OCM_DIRECTORY_SERVICE_URLS" desc:"Space delimited URLs of the directory services." introductionVersion:"3.5.0"`
 	InviteAcceptDialog   string `yaml:"invite_accept_dialog" env:"OCM_INVITE_ACCEPT_DIALOG" desc:"/open-cloud-mesh/accept-invite;The frontend URL where to land when receiving an invitation" introductionVersion:"3.5.0"`
+	OCMClientInsecure    bool   `yaml:"ocm_client_insecure" env:"OCM_SCIENCEMESH_OCM_CLIENT_INSECURE" desc:"Dev-only. Disable TLS certificate verification for ScienceMesh WAYF discovery requests. Do not set this in production environments." introductionVersion:"unreleased"`
```
MahdiBaghbani: `introductionVersion:"unreleased` I'd need your help here, since I'm not sure which upcoming version to use here :smile:

kulmann: Please use `introductionVersion:"%%NEXT%%"`. That will be replaced with the correct version on the next release. :-)

kulmann: Could you please rename the env var to `OCM_CLIENT_INSECURE` and also add `OC_INSECURE` to it? The resulting env var string should be: `OC_INSECURE;OCM_CLIENT_INSECURE`. That way it will be sufficient to set `OC_INSECURE=true` but can still be overwritten with the more specific variable if needed.

PR #2411, opened 2026-02-27
update links and references in CONTRIBUTING.md

ScharfViktor:
Fix broken links and outdated references in [CONTRIBUTING.md](https://github.com/opencloud-eu/opencloud?tab=contributing-ov-file#opencloud-contribution-guidelines)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,8 +152,7 @@ Unsure where to begin contributing to OpenCloud? You can start by looking throug
 It is fine to pick one of the lists following personal preference.
 While not perfect, the number of comments is a reasonable proxy for the impact a given change will have.
 
-To find out how to set up OpenCloud for local development, please refer to the [Developer Documentation](https://docs.opencloud.eu/docs/dev/web/getting-started).
-It contains a lot of information that will come in handy when starting to work on the project.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
v-scharf: that is correct link but it's more about web. Do we have setup guide for backend side?

dragotin: Probably the best bet is to link to the general README for the server, and keep both links for Web and Backend.

CONTRIBUTING.md
```diff
@@ -257,7 +256,7 @@ Categorizes the issue to also indicate the type of the issue.
 
 #### Status
 
-The status in the ticket life cycle. Keep an eye on that one, especially for the `Waiting-for-Feedback` tag which might indicate that the reporter is asked for feedback.
+The status in the ticket life cycle. Keep an eye on that one, especially for the `Status:Needs-Review` tag which might indicate that the reporter is asked for feedback.
```
dragotin, line 258: Hm, we're not using this label a lot. Maybe we should review the labels again. But that is for a different PR I guess.

CONTRIBUTING.md
v-scharf: Thanks 👍 fixed

PR #2484, opened 2026-03-16
feat(init): support environment variables for all generated secrets

### LINKED ISSUE #2483: opencloud init: support environment variables for all generated secrets
## Problem

`opencloud init` generates random passwords for all internal secrets (IDM service passwords, JWT, transfer secrets, etc.) and provides no mechanism to inject pre-defined values via environment variables.

In containerized deployments (Kubernetes, Docker), this means:

- Every pod restart with an empty config dir generates new random secrets
- IDM (LDAP) service passwords written during init no longer match what services use at runtime
- This causes `LDAP bind` failures and broken authentication after any restart without config volume persistence

Only `--admin-password` is exposed as a CLI flag. The 12 other secrets have no override mechanism.

## Affected secrets

| Secret | Config field | Proposed ENV var |
|--------|-------------|------------------|
| IDM service password | `idm.service_user_config.password` | `IDM_SVC_PASSWORD` |
| IDM Reva service password | `idm.service_user_config.reva_password` | `IDM_REVASVC_PASSWORD` |
| IDM IDP service password | `idm.service_user_config.idp_password` | `IDM_IDPSVC_PASSWORD` |
| Admin password | `admin_password` | `OC_ADMIN_PASSWORD` |
| JWT secret | `token_manager.jwt_secret` | `OC_JWT_SECRET` |
| Machine auth API key | `machine_auth_api_key` | `OC_MACHINE_AUTH_API_KEY` |
| Transfer secret | `transfer_secret` | `OC_TRANSFER_SECRET` |
| Service account secret | `service_account.service_account_secret` | `OC_SERVICE_ACCOUNT_SECRET` |
| WOPI secret | `collaboration.wopi.wopi_src_secret` | `COLLABORATION_WOPI_SECRET` |
| System user API key | `system_user_api_key` | `SYSTEM_USER_API_KEY` |
| URL signing secret | n/a | `OC_URL_SIGNING_SECRET` |
| Thumbnails transfer secret | n/a | `THUMBNAILS_TRANSFER_SECRET` |

## Proposed fix

Add a helper in `opencloud/pkg/init/` that checks for the corresponding environment variable before falling back to random generation:

```go
func secretFromEnvOrRandom(envVar string, length int) (string, error) {
    if v := os.Getenv(envVar); v != "" {
        return v, nil
    }
    return generators.GenerateRandomPassword(length)
}
```

This is fully backward-compatible: when no ENV var is set, behavior is identical to today.

## Use case

Helm charts and container orchestrators can inject stable, pre-defined secrets via Kubernetes Secrets → ENV vars, eliminating the need for config volume persistence and ensuring LDAP credentials remain consistent across restarts.

bernardgut:
## Summary

`opencloud init` now checks for environment variables before generating random passwords for internal secrets. This is fully backward-compatible — when no ENV var is set, behavior is identical to today.

Closes #2483

## Changes

- `opencloud/pkg/init/functions.go`: add `secretFromEnvOrRandom` helper
- `opencloud/pkg/init/init.go`: use helper for all 12 generated secrets

## Motivation

In containerized deployments, pod restarts with an empty config directory generate new random secrets. The IDM service passwords written during init then no longer match what LDAP expects at runtime, causing bind failures and broken authentication.

This change lets orchestrators inject stable secrets via ENV vars, eliminating the need for config volume persistence.

## Supported ENV vars

`IDM_SVC_PASSWORD`, `IDM_REVASVC_PASSWORD`, `IDM_IDPSVC_PASSWORD`, `OC_ADMIN_PASSWORD`, `OC_JWT_SECRET`, `OC_MACHINE_AUTH_API_KEY`, `OC_TRANSFER_SECRET`, `OC_SERVICE_ACCOUNT_SECRET`, `COLLABORATION_WOPI_SECRET`, `SYSTEM_USER_API_KEY`, `OC_URL_SIGNING_SECRET`, `THUMBNAILS_TRANSFER_SECRET`

## Testing

- Built custom Docker image with this patch
- Deployed to Kubernetes with all 12 secrets injected via K8s Secrets → ENV vars
- Verified OIDC login, Graph API, WebDAV upload/download, PROPFIND — all pass
- **Restart resilience**: deleted pod, confirmed all auth and operations work after restart without config persistence

### DIFF
--- a/opencloud/pkg/init/functions.go
+++ b/opencloud/pkg/init/functions.go
@@ -8,6 +8,8 @@ import (
 	"os/exec"
 	"path"
 	"time"
+
+	"github.com/opencloud-eu/opencloud/pkg/generators"
 )
 
 func checkConfigPath(configPath string) error {
@@ -105,4 +107,11 @@ func writePatch(configPath string, yamlOutput []byte) error {
 	}
 	fmt.Printf("diff written to %s\n", patchPath)
 	return nil
+}
+
+func secretFromEnvOrRandom(envVar string, length int) (string, error) {
+	if v := os.Getenv(envVar); v != "" {
+		return v, nil
+	}
+	return generators.GenerateRandomPassword(length)
 }
--- a/opencloud/pkg/init/init.go
+++ b/opencloud/pkg/init/init.go
@@ -109,55 +109,55 @@ func CreateConfig(insecure, forceOverwrite, diff bool, configPath, adminPassword
 		storageUsersMountID = uuid.NewString()
 		serviceAccountID = uuid.NewString()
 
-		idmServicePassword, err = generators.GenerateRandomPassword(passwordLength)
+		idmServicePassword, err = secretFromEnvOrRandom("IDM_SVC_PASSWORD", passwordLength)
 		if err != nil {
 			return fmt.Errorf("could not generate random password for idm: %s", err)
 		}
-		idpServicePassword, err = generators.GenerateRandomPassword(passwordLength)
+		idpServicePassword, err = secretFromEnvOrRandom("IDM_IDPSVC_PASSWORD", passwordLength)
 		if err != nil {
 			return fmt.Errorf("could not generate random password for idp: %s", err)
 		}
 		ocAdminServicePassword = adminPassword
 		if ocAdminServicePassword == "" {
-			ocAdminServicePassword, err = generators.GenerateRandomPassword(passwordLength)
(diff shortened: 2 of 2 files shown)


bernardgut:
This is ready for review. Thanks.

micbar:
See my comment in the ticket.

That is the wrong approach. Just use the ENV variables in kubernetes and completely skip the init script, which is just a helper for docker compose.

bernardgut:
After testing extensively: Not actually needed. MB
See https://github.com/opencloud-eu/opencloud/issues/2483#issuecomment-4066587250

Issue #2319, opened 2026-02-11
[research/poc] Use share-manager for managing space memberships


rhafer:
- howto rework the space listing, especially with regards to personal spaces
- try to avoid different mechanisms per space type 
- what about existing setups (migrations?)
- possible options for caching

# Acceptance Criteria:

- [ ] ADR 

related to: https://github.com/opencloud-eu/opencloud/pull/2183/

rhafer:
> * howto rework the space listing, especially with regards to personal spaces

The current way to list spaces (e.g. `/me/drives`) via the  CS3 ListStorageSpaces() call can probably stay as it is. That is probably not going away anytime soon. We could however also utilize `GetSharedWithMe()` in the future, which might have
some (performance benefits)

>  * try to avoid different mechanisms per space type

Ideally Personal Spaces would also (upon creation) add an entry to the share manager

>  * what about existing setups (migrations?)

Existing space grants needs to be imported into the share manager once. Would be extra nice if this could happen without interruptive to the running cloud.
 
>  * possible options for caching

Not sure what was meant with this, but AFAIK the (jsoncs3) share-manager already implements caching. However we need to add more filtering options e.g. to be able to efficiently  to query like:
- give me only spaces
- give me everything that is not a space

rhafer:
Just for completeness: The gateway already has a config switch for using the shareprovider for spaces: https://github.com/opencloud-eu/reva/blob/main/internal/grpc/services/gateway/usershareprovider.go#L43 (I guess the eos backend used that)

rhafer:
Also: With this change we'd might be able to get rid of that hacky permission ids on the graph api for space permissions (https://github.com/opencloud-eu/opencloud/blob/main/services/graph/pkg/service/v0/base.go#L582).

We could just use the `shareid` like for "normal" shares instead of that `u:<userid>` or `g:<groupid>` hack.

rhafer:
Also²: The share-manager is able to track who added a member to the space and when. This info is not available on the space grants in the storage provider.

Update: This is not quite true the `grant` attribute at least have the creator id as a property.

rhafer:
I just created https://github.com/opencloud-eu/opencloud/pull/2492 to illustrate the required changes.

PR #2492, opened 2026-03-17
poc: Use share manager for space membership

rhafer:
This is a proof of concept implementation for #2319. (Persisting Space-Membership in the share-manager)

Requires an adjustment to the CS3 API. Basically just an addtional sharefilter-type: https://github.com/rhafer/cs3apis/commit/6333ce58573cbcba3a17685cf15c2a04d5b05b57

The required reva changes are here: https://github.com/rhafer/reva/tree/spaces-in-sharemanager. The rest is mainly accomplished by settting `use_common_space_root_share_logic` to `true`. If we decide to move forward with this I'd suggest to remove that option and make the behavior the default. Should get us rid of a bit more code in reva.

As you can see, the changes simplifies the sharing code in the graph service quite a bit as it removes quite a few of the "is this a space root" special cases.


rhafer:
cc @aduffeck @butonic @micbar

services/gateway/pkg/revaconfig/config.go
```diff
@@ -51,7 +51,8 @@ func GatewayConfigFromStruct(cfg *config.Config, logger log.Logger) map[string]i
 					"ocminvitemanagersvc":           cfg.OCMEndpoint,
 					"ocmproviderauthorizersvc":      cfg.OCMEndpoint,
 					"ocmcoresvc":                    cfg.OCMEndpoint,
-					"commit_share_to_storage_grant": cfg.CommitShareToStorageGrant,
+					"use_common_space_root_share_logic": true,
```
aduffeck, line 54: Is this actually being used? I didn't find a reference to it in reva.

aduffeck (commented):
This looks really nice as a first step.

I suppose in the future we might want to retrieve the list of shared spaces from the space manager too instead of relying on the grants using `ListStorageSpaces` (analogous to what we do with "regular" shares), but that's something we can discuss later.

services/gateway/pkg/revaconfig/config.go
rhafer: It's used in the gateway:

https://github.com/opencloud-eu/reva/blob/main/internal/grpc/services/gateway/gateway.go#L79

We should probably remove that config switch, if we decide to switch to the share-manager.

aduffeck: Ah, I thought it was a new flag being introduced with your change. Thanks.

rhafer:
Update, I've addressed many of the failing tests. However there is a few things I can't really fix myself:

1. Some of the test still fail because they assume a specific format of the `permissionId` on the SpaceRoots. Up to now for space roots that format was something like `u:<grantee-userid>` or `g:<grantee-groupid>` that was however just a crutch and has change now. The permissionId should always be treated as opaque Id. I already fixed that in a lot of places, but the remaining places exceed my php skills

2. The more difficult part is probably all the failing tests that still use the `ocs` endpoints. Before diving into fixing ocs we should probably finally decide how to move forward with `ocs` for sharing (especially for managing space membership). Maybe it is about time to just let is rest in peace. (@micbar @aduffeck @dragotin opinions?) AFAIK we don't have any  client that is still using `ocs` for managing space memberships.

rhafer:
>     2. The more difficult part is probably all the failing tests that still use the `ocs` endpoints.

The `ocs` tests are green now. The `ocs` API is now able to handle the space memberships when they are stored in the ShareProvider.

rhafer:
The tests that are still failing are:

- apiSharingNg1/removeAccessToDrive.feature:132
- apiSharingNg1/removeAccessToDrive.feature:133
- apiSharingNg1/removeAccessToDrive.feature:134
- apiSharingNg1/removeAccessToDrive.feature:137
- apiSharingNg1/removeAccessToDrive.feature:145
- apiSharingNg1/removeAccessToDriveItem.feature:123
- apiSharingNg1/removeAccessToDriveItem.feature:124
- apiSharingNg1/removeAccessToDriveItem.feature:125
- apiSharingNg1/removeAccessToDriveItem.feature:143
- apiSharingNg1/removeAccessToDriveItem.feature:144
- apiSharingNg1/removeAccessToDriveItem.feature:145

As already mentioned, all of them are failing because the test suite is making wrong assumptions about the format of the `permissionid` for permissions on Spaces. With this change the `permissionId` for a space membership does not longer have the format `u:<userid>` or `g:<groupid>`. The IDs should just be treated as opaque ids (similar to permissions on folder and files). 

So to remove the access of a specific user from a space. One needs to first list the permissions on that space and then pick the permission id of the permission that belongs to the user.

@ScharfViktor would be cool if you could take a look into this. This is a bit above my PHP level. For many of the other failures I've worked around this by just recording the id using `shareNgAddToCreatedUserGroupShares` maybe that could work here as well, I am unsure.

ScharfViktor:
```
apiSharingNg1/removeAccessToDrive.feature:137
apiSharingNg1/removeAccessToDrive.feature:145
```
still failed because user can remove own access even he is the last space manager
```
Scenario: user cannot remove himself from the project space if he is the last manager using root endpoint # /Users/v.scharf/Work/opencloud/opencloud/tests/acceptance/features/apiSharingNg1/removeAccessToDrive.feature:137
    Given the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API         # GraphContext::theAdministratorHasGivenTheRoleUsingTheGraphApi()
    And user "Alice" has created a space "NewSpace" with the default quota using the Graph API              # SpacesContext::theUserHasCreatedASpaceByDefaultUsingTheGraphApi()
    When user "Alice" tries to remove own access from space "NewSpace" using root endpoint of the Graph API # SharingNgContext::userRemovesOwnAccessFromSpaceUsingGraphAPI()
    Then the HTTP status code should be "403"                                                               # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 204 is not the expected value 403
      Failed asserting that 204 matches expected '403'.
    And the user "Alice" should have a space called "NewSpace"   
```

tests/acceptance/bootstrap/SharingNgContext.php
```diff
@@ -1040,12 +1028,18 @@ public function removeAccessToSpace(
...
 		$spaceId = ($this->spacesContext->getSpaceByName($sharer, $space))["id"];
 
-		$permissionID = match ($shareType) {
-			'link' => $this->featureContext->shareNgGetLastCreatedLinkShareID(),
-			'user' => 'u:' . $this->featureContext->getAttributeOfCreatedUser($recipient, 'id'),
-			'group' => 'g:' . $this->featureContext->getAttributeOfCreatedGroup($recipient, 'id'),
-			default => throw new Exception("shareType '$shareType' does not match user|group|link "),
-		};
+		// if recipient is not provided, it means user tries to remove own access, then we need to get the user permission id
+		if (!isset($recipient)) {
```
rhafer: @ScharfViktor This breaks `userRemovesLinkFromSpaceUsingRootEndpointOfGraphAPI()` I think. 
It calls `$this->removeAccessToSpace($sharer, 'link', $space)` (so `$recipient` is unset). But it would need `$permissionID` be set to `$this->featureContext->shareNgGetLastCreatedLinkShareID()` because of `$shareType='link'`

rhafer:
> ```
> apiSharingNg1/removeAccessToDrive.feature:137
> apiSharingNg1/removeAccessToDrive.feature:145
> ```
> 
> still failed because user can remove own access even he is the last space manager
> 
> ```
> Scenario: user cannot remove himself from the project space if he is the last manager using root endpoint # /Users/v.scharf/Work/opencloud/opencloud/tests/acceptance/features/apiSharingNg1/removeAccessToDrive.feature:137
>     Given the administrator has assigned the role "Space Admin" to user "Alice" using the Graph API         # GraphContext::theAdministratorHasGivenTheRoleUsingTheGraphApi()
>     And user "Alice" has created a space "NewSpace" with the default quota using the Graph API              # SpacesContext::theUserHasCreatedASpaceByDefaultUsingTheGraphApi()
>     When user "Alice" tries to remove own access from space "NewSpace" using root endpoint of the Graph API # SharingNgContext::userRemovesOwnAccessFromSpaceUsingGraphAPI()
>     Then the HTTP status code should be "403"                                                               # FeatureContext::thenTheHTTPStatusCodeShouldBe()
>       HTTP status code 204 is not the expected value 403
>       Failed asserting that 204 matches expected '403'.
>     And the user "Alice" should have a space called "NewSpace"   
> ```

I've bumped  reva to https://github.com/opencloud-eu/reva/pull/587/changes/90717df04562369b91fe2748abb5babcf1e52453 that should have fixed it.

ScharfViktor:
> I've bumped reva to [opencloud-eu/reva@90717df](https://github.com/opencloud-eu/reva/commit/90717df04562369b91fe2748abb5babcf1e52453) that should have fixed it.

I can confirm, that it fixed

rhafer:
Closing this to reopen it against the feature/guest-links branch

PR #2501, opened 2026-03-20
add final step for branch protection

ScharfViktor:
related: https://github.com/opencloud-eu/qa/pull/81

image

Now the `all-check-finished` returns `exit(0)` if all tests pass, or `exit(1)` if any tests fail—we made this step required to protect the branch

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -498,7 +498,8 @@ def main(ctx):
     is_release_pr = (ctx.build.event == "pull_request" and ctx.build.sender == "openclouders" and "🎉 release" in ctx.build.title.lower())
     if is_release_pr:
         return checkVersionPlaceholder() + \
-               licenseCheck(ctx)
+               licenseCheck(ctx) + \
+               notifyMatrixCheckSteps(ctx, getPipelineNames(licenseCheck(ctx) + checkVersionPlaceholder()))
 
     build_release_helpers = \
         readyReleaseGo()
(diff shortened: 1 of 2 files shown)


v-scharf:
> This would mean that the chat notification step will also show as "fail" when there are other failed steps, but that should be OK, because in that case CI is already red

yes, It will failed, but sends notification

.woodpecker.star
```diff
@@ -569,7 +570,7 @@ def main(ctx):
         ),
     )
 
-    pipelines = test_pipelines + build_release_pipelines + genDocsPr(ctx) + notifyMatrix(ctx)
+    pipelines = test_pipelines + build_release_pipelines + genDocsPr(ctx) + notifyMatrixCheckSteps(ctx, getPipelineNames(testPipelines(ctx)))
```
micbar, line 573: This needs to be discussed. Let us look at it next week. We need to also wait for all non test pipelines.

v-scharf: The `matrix-notification.go` file checks that all pipelines have completed successfully. However, it is important that it runs at the very end.

Currently, it runs after:
image

I haven't found any pipelines that run after the tests. Do you know of any?

kulmann:
Screenshot 2026-03-23 at 06 43 53

I don't know how to make github to like it 🙈  the status doesn't seem to be reported... edit: ah lol, my screenshot wasn't even from this PR. Is the status a push-thing? this PR here is green... nevermind then I guess.

.woodpecker.star
micbar: Fine. Lets merge it.

PR #2514, opened 2026-03-24
Improve opensearch highlighting, fix favorites

aduffeck:
This PR introduces a new opensearch index which
* fixes persisting and querying favorite information
* enables the Fast Vector Highlighter which performs much faster and doesn't run into the character limits the previous default highlighter had

**Note** This is an incompatible change. Users of the opensearch backend will have to drop their old index and rebuild it using

```
opencloud search index --all-spaces
```


micbar:
Good work!

There is an API test related to stop words failing.
Are we planning a behavior change?

aduffeck:
> Good work!
> 
> There is an API test related to stop words failing. Are we planning a behavior change?

Yes, we currently enable stop word filtering by default which leads to phrase accuracy issues during search and chopped up highlighting results. I would like to change the default to keep stop words.

Here's an example showing both versions:

Content:

image

Highlights:

image


Indexing stop words slightly increases the index size, but I would argue that the current highlighting results are very confusing and not helpful at all.

Issue #2282, opened 2026-02-05
Double check nats is shut down gracefully in docker


butonic:
I saw a corrupted KV_ids-storage-users stream. opencloud would consume all available CPU and memory. AFAICT this only happens when nats is killed while writing to disk. This is unrelated to the Jepson report and more of our own doing I fear. My guess is that we are still not gracefully shutting down nats when the opencloud process receives the SIGTERM from docker. In kubernetes nats is deployed as a dedicated service so this is not as much of a problem there.

We did look into the [graceful shutdown](https://github.com/opencloud-eu/opencloud/pull/1478) but there are still [corner cases](https://github.com/opencloud-eu/opencloud/issues/1494). 

Maybe there is still an `exit(0)` somewhere that kills nats while it is writing the stream.

👀 

well it is not an exit(0) but the suture runtime does not seem to wait for all services. I have patched the main.go to panic on exit and wait 15 seconds when the command has finished. I also wanted to know when nats is being shut down gracefully:
```diff
diff --git a/opencloud/cmd/opencloud/main.go b/opencloud/cmd/opencloud/main.go
index 7968efc1d87..9c2022982fc 100644
--- a/opencloud/cmd/opencloud/main.go
+++ b/opencloud/cmd/opencloud/main.go
@@ -3,13 +3,22 @@ package main
 import (
        "fmt"
        "os"
+       "time"
 
        "github.com/opencloud-eu/opencloud/opencloud/pkg/command"
 )
 
+var osExit = os.Exit
+
 func main() {
+       osExit = func(code int) {
+               panic(fmt.Sprintf("os.Exit(%d) called", code))
+       }
        if err := command.Execute(); err != nil {
                fmt.Fprintln(os.Stderr, err.Error())
                os.Exit(1)
        }
+       fmt.Println("waiting 15 sec")
+       time.Sleep(15 * time.Second)
+       fmt.Println("program exit")
 }
diff --git a/services/nats/pkg/command/server.go b/services/nats/pkg/command/server.go
index 342036167e8..9deec13b3af 100644
--- a/services/nats/pkg/command/server.go
+++ b/services/nats/pkg/command/server.go
@@ -5,6 +5,7 @@ import (
        "crypto/tls"
        "fmt"
        "os/signal"
+       "time"
 
        "github.com/opencloud-eu/opencloud/pkg/config/configlog"
        pkgcrypto "github.com/opencloud-eu/opencloud/pkg/crypto"
[...]

PR #2518, opened 2026-03-24
revisit checks

butonic:
I saw that pods were marked as not ready in k8s when running load tests. So long that kubernetes decides to kill them. That is a problem if the shutdown procedure does not persist all data. See https://github.com/opencloud-eu/opencloud/issues/2282

It might be that the share persisting in https://github.com/opencloud-eu/reva/pull/567 eats too much cpu cyles? Other services are running in the same process, they also eat cpu cycles. Maybe that starves the check handlers? Some http or grpc queue might be full?

health checks should only verify the process is alive. a simple 200 OK response is enough for that.

the readyness check can check if the process is ready to serve traffick.

under high load kubernetes will stop sendig it traffick if the readyness probe fails. but it will only kill the pod if the healthy probe fails.

This needs more investigation.

### DIFF
--- a/pkg/checks/checkgrpc.go
+++ b/pkg/checks/checkgrpc.go
@@ -6,22 +6,31 @@ import (
 
 	"github.com/opencloud-eu/opencloud/pkg/handlers"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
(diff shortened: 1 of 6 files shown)


butonic:
Currently, all our health checks are running against localhost. Which is why the FailSaveAddress now always returns localhost in case `0.0.0.0` or `::` was configured. That should be good enough for all deployment scenarios. But needs to be double checked.

PR #2524, opened 2026-03-25
Fix bleve batches

aduffeck:
This PR fixes the batch size limit with the bleve backend whichwasn't taken into account properly and sets a more conservative default batch size.


aduffeck:
@micbar I would suggest to backport this to the stable branch. wdyt?

PR #2531, opened 2026-03-26
[full-ci] api-test: check that report contains favorites

ScharfViktor:
related https://github.com/opencloud-eu/opencloud/pull/2510

step: `And as user "Alice" folder/file "name" should be favorited` checks that Propfind has `<oc:favorite>1</oc:favorite>`

this PR adds checking that Report response contains`<oc:favorite>1</oc:favorite>` if file/folder marked as favorite or `<oc:favorite>0</oc:favorite>` if it's not favorite

### DIFF
--- a/tests/acceptance/features/apiContract/propfind.feature
+++ b/tests/acceptance/features/apiContract/propfind.feature
@@ -124,6 +124,7 @@ Feature: Propfind test
   @issue-1523
   Scenario: propfind response contains a restored folder with correct name
     Given user "Alice" has created a folder "folderMain" in space "Personal"
+    And user "Alice" has marked folder "folderMain" as favorite from space "Personal"
     And user "Alice" has deleted folder "folderMain"
     And user "Alice" has created a folder "folderMain" in space "Personal"
     When user "Alice" restores the folder with original path "/folderMain" to "/folderMain (1)" using the trashbin API
@@ -136,6 +137,7 @@ Feature: Propfind test
       | oc:name        | folderMain        |
       | oc:permissions | RDNVCKZP          |
       | oc:size        | 0                 |
+      | oc:favorite    | 0                 |
     And as user "Alice" the PROPFIND response should contain a resource "folderMain (1)" with these key and value pairs:
       | key            | value             |
       | oc:fileid      | %file_id_pattern% |
@@ -143,3 +145,4 @@ Feature: Propfind test
       | oc:name        | folderMain (1)    |
       | oc:permissions | RDNVCKZP          |
       | oc:size        | 0                 |
+      | oc:favorite    | 1                 |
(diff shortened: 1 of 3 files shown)


v-scharf:
@aduffeck here are some failed tests because `REPORT` doesn't return `<oc:favorite>0</oc:favorite>`

I expect similar behavior with PROPFIND response

Steps: 
- create folders `main/sub` and `main/sub-1`
- mark `main/sub` favorites
- request PROPFING: both contain `oc:favorite`
Screenshot 2026-03-26 at 16 34 52

- request REPORT (search by `sub`): only `main/sub` contains `oc:favorite`

but `main/sub-1` doesn't contains `oc:favorite` - ❌️
Screenshot 2026-03-26 at 16 39 06

aduffeck:
Hm, o we really have to include the `<oc:favorite>0</oc:favorite>` in the response for each item that isn't a favorite (which is gonna be almost every result)?
If clients don't require it I would rather omit the negative case. @AlexAndBear @JammingBen do you have an opinion on that?

AlexAndBear:
@aduffeck web doesn't require it 🤷‍♀️

v-scharf:
> If clients don't require it I would rather omit the negative case.

I think we specify what we want to get through the `PROPFIND` request payload: 
```
<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
  <d:prop>
    <oc:permissions/>
    <oc:favorite/>
    <oc:fileid/>
    <oc:file-parent/>
    <oc:name/>
    <d:lockdiscovery/>
    <d:activelock/>
    <oc:owner-id/>
    <oc:owner-display-name/>
    <oc:remote-item-id/>
    <oc:shareroot/>
    <oc:share-types/>
    <oc:privatelink/>
    <d:getcontentlength/>
    <oc:size/>
    <d:getlastmodified/>
    <d:getetag/>
    <d:getcontenttype/>
    <d:resourcetype/>
    <oc:tags/>
    <oc:audio/>
    <oc:location/>
    <oc:image/>
    <oc:photo/>
    <oc:has-preview/>
  </d:prop>
</d:propfind>
```

if we exclude `<oc:favorite/>` -> response will be without favorite


But it doesn't work for `REPORT`:
- if we exclude `<oc:favorite/>` -> we still get `<oc:favorite>1</oc:favorite>`
- if we send request with `<oc:favorite/>` -> we don't get `<oc:favorite>0</oc:favorite>`

aduffeck:
I don't think we have to support all features for `REPORT` which we support for `PROPFIND`. I'd rather not add complexity to the code which isn't really needed by the server or the clients and remove those tests again to be honest.
We can always implement that behavior when it's actually needed.

v-scharf:
> I don't think we have to support all features for `REPORT` which we support for `PROPFIND`. I'd rather not add complexity to the code which isn't really needed by the server or the clients and remove those tests again to be honest. We can always implement that behavior when it's actually needed.

Yeah, agree. I deleted tests which expected `<oc:favorite>0</oc:favorite>` in the report response

v-scharf:
image

https://ci.opencloud.rocks/repos/3/pipeline/2447/config


we forgot that some depended pipelines can be skipped

Example here: 
my Pr changes only `tests/acceptance/` -> `test-litmus` doesn't start because` exclude": skipIfUnchanged(ctx, "litmus")`

`all-checks-finished` doesn't start because no `test-litmus` in the `depends_on`

PR #2553, opened 2026-03-31
adapt pipeline for prod releases

dragonchaser:
This PR splits the check_version pipeline into two steps, the step for detecting leftover `%%NEXT_PRODUCTION_VERSION%%`  placeholders only is run when the label `production_release` is set for the PR.

/cc @ScharfViktor && @individual-it

### DIFF
--- a/.woodpecker.star
+++ b/.woodpecker.star
@@ -1838,7 +1838,7 @@ def checkVersionPlaceholder():
         "when": [
             event["pull_request"],
         ],
-    },{
+    }, {
         "name": "check-version-placeholder-next-production-release",
         "steps": [
             {
@@ -1850,14 +1850,14 @@ def checkVersionPlaceholder():
                         dirs["base"],
                     ),
                     'if [ -s next_production_version.txt ]; then echo "replace version placeholders"; cat next_production_version.txt; exit 1; fi',
-                    ],
+                ],
             },
         ],
         "when": [
             {
-               "event": "pull_request",
-               "evaluate": 'CI_COMMIT_PULL_REQUEST_LABELS contains "production_release"',
-            }
+                "event": "pull_request",
+                "evaluate": 'CI_COMMIT_PULL_REQUEST_LABELS contains "production_release"',
+            },
         ],
     }]
 


.woodpecker.star
```diff
@@ -1838,6 +1838,24 @@ def checkVersionPlaceholder():
         "when": [
             event["pull_request"],
         ],
+        "name": "check-version-placeholder-next-production-release",
```
v-scharf, line 1842: now `check-version-placeholder-next-production-release` replace `check-version-placeholder-next`. I think we need do `[{}, {}]`

dragonchaser:
Please merge as is, this will never get green because of set test labels.

v-scharf:
> Please merge as is, this will never get green because of set test labels.

we can't merge it until getting green`all-checks-finished`

v-scharf:
I see your last running 
Screenshot 2026-03-31 at 16 14 08


to get green you need run CI without `🎉 Release` in the title and label `production_release`
try to force push

dragonchaser:
@ScharfViktor I did, otherwise I would not be able to test the functionality.

.woodpecker.star
```diff
@@ -1838,6 +1838,27 @@ def checkVersionPlaceholder():
...
+                    ),
+                    'if [ -s next_production_version.txt ]; then echo "replace version placeholders"; cat next_production_version.txt; exit 1; fi',
+                    ],
+            },
+        ],
+        "when": [
+            event["pull_request"],
+            {
+                "evaluate": 'CI_COMMIT_PULL_REQUEST_LABELS contains "production_release"',
+            },
```
individual-it: This does not work, because ` If at least one of the conditions in the when block evaluate to true the step is executed, otherwise it is skipped.` see: https://woodpecker-ci.org/docs/usage/workflow-syntax#when---conditional-execution

Also, the linter shows a warning:
Image

`A condition is evaluated to true if all sub-conditions are true.`

```suggestion
            {
               "event": "pull_request",
               "evaluate": 'CI_COMMIT_PULL_REQUEST_LABELS contains "production_release"',
            }
```

or

```suggestion
            {
                "evaluate": CI_PIPELINE_EVENT == "pull_request" && CI_COMMIT_PULL_REQUEST_LABELS contains "production_release"',
            },
```

individual-it (changes requested):

PR #2560, opened 2026-04-01
feat(collaboration): add UserExtraInfo with avatar and mail to CheckFileInfo

pedropintosilva:
feat(collaboration): add UserExtraInfo with avatar and mail to CheckFileInfo

Add UserExtraInfo (avatar + mail) to the WOPI CheckFileInfo response for
authenticated, non-public-share users.

UserExtraInfo format (per Collabora SDK):
https://sdk.collaboraonline.com/docs/advanced_integration.html#userextrainfo

```json
{
  "avatar": "http://url/to/user/avatar",
  "mail": "user@server.com"
}
```

After this change, CheckFileInfo returns:

```json
{
  "BaseFileName": "Pedro-filled-hazcom.docx",
  "UserFriendlyName": "Admin",
  "UserId": "346364...39323030",
  "UserCanWrite": true,
  "UserCanRename": true,
  "IsAdminUser": true,
  "EnableInsertRemoteImage": true,
  "EnableInsertRemoteFile": true,
  "EnableOwnerTermination": true,
  "UserExtraInfo": {
    "avatar": "https://host:9300/wopi/avatars/{userID}?access_token={wopiToken}",
    "mail": "admin@example.org"
  },
  "PostMessageOrigin": "https://localhost:9200",
  "message": "CheckFileInfo: success"
}
```

Avatars are served via a new /wopi/avatars/{userID} endpoint on the
collaboration service, authenticated by the WOPI token. The endpoint
calls the Graph service directly (bypassing the proxy) using the reva
access token via x-access-token header.

All tests pass:
go test ./services/collaboration/... ./services/graph/... ./services/proxy/...

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>

### DIFF
--- a/services/collaboration/pkg/config/cs3api.go
+++ b/services/collaboration/pkg/config/cs3api.go
@@ -9,10 +9,9 @@ import (
 // CS3Api defines the available configuration in order to access to the CS3 gateway.
 type CS3Api struct {
 	Gateway                 Gateway               `yaml:"gateway"`
-	DataGateway             DataGateway            `yaml:"datagateway"`
-	GraphEndpoint           string                 `yaml:"graph_endpoint" env:"COLLABORATION_CS3API_GRAPH_ENDPOINT" desc:"The internal HTTP endpoint of the Graph service, used to fetch user profile photos for avatars." introductionVersion:"4.0.0"`
-	GRPCClientTLS           *shared.GRPCClientTLS  `yaml:"grpc_client_tls"`
-	APPRegistrationInterval time.Duration          `yaml:"app_registration_interval" env:"COLLABORATION_CS3API_APP_REGISTRATION_INTERVAL" desc:"The interval at which the app provider registers itself." introductionVersion:"4.0.0"`
+	DataGateway             DataGateway           `yaml:"datagateway"`
+	GRPCClientTLS           *shared.GRPCClientTLS `yaml:"grpc_client_tls"`
+	APPRegistrationInterval time.Duration         `yaml:"app_registration_interval" env:"COLLABORATION_CS3API_APP_REGISTRATION_INTERVAL" desc:"The interval at which the app provider registers itself." introductionVersion:"4.0.0"`
 }
 
 // Gateway defines the available configuration for the CS3 API gateway
--- a/services/collaboration/pkg/config/defaults/defaultconfig.go
+++ b/services/collaboration/pkg/config/defaults/defaultconfig.go
@@ -65,7 +65,6 @@ func DefaultConfig() *config.Config {
 			DataGateway: config.DataGateway{
 				Insecure: false,
 			},
-			GraphEndpoint:           "http://127.0.0.1:9120/graph",
 			APPRegistrationInterval: 30 * time.Second,
 		},
 	}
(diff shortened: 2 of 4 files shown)


pedropintosilva:
image

pedropintosilva:
I cannot see what are the specific complains from SonarCloud , I'm not part of the org so, I get some error when attempting to open it in my SonarCloud account.

rhafer:
> I cannot see what are the specific complains from SonarCloud , I'm not part of the org so, I get some error when attempting to open it in my SonarCloud account.

Yeah, that is really inconvenient. I took a quick look into the complaints and they all where about to much "Cognitive Complexity" (i.e. too deeply nested if/switch statements). SonarCloud is probably right about that, but I've "accepted" those warning for now, since it isn't your changes that introduced the complexity.

Btw, I think you should see the warnings in the annotations here: https://github.com/opencloud-eu/opencloud/pull/2560/changes (I can see them there at least)

Thanks for you contributions, I like it.  We still need to take a deeper look into the changes before we can merge them.

micbar:
@pedropintosilva 
First of all, really nice feature! That creates more adoption and "feel good vibes".

I think we need to change the implementation a bit.

You are making the avatar GET Api call an unprotected == public route. That is not possible.

We should implement a dedicated endpoint in the Collaboration service which is authenticated by the WOPI token.
I think we did something similar with the template fetching.

Does Collabora have the possibility to send the WOPI token to the download endpoint?

pedropintosilva:
> @pedropintosilva First of all, really nice feature! That creates more adoption and "feel good vibes".

Agreed :)

> 
> I think we need to change the implementation a bit.
> 
> You are making the avatar GET Api call an unprotected == public route. That is not possible.

You are correct, I was a bit unsure if that's ok (that's why I left that note in the commit message) but at the end I thought would be ok, as anyhow even guests benefit from seeing who they are collaborating with.

> 
> We should implement a dedicated endpoint in the Collaboration service which is authenticated by the WOPI token. I think we did something similar with the template fetching.
> 
> Does Collabora have the possibility to send the WOPI token to the download endpoint?

Hm I don't think we have we just set the img.src directly: https://github.com/CollaboraOnline/online/blob/main/browser/src/app/LOUtil.ts#L351

but you are right we already have that /wopi/templates/ case so, it would be elegant to have also /wopi/avatars/ and we would still receive the url that we need.

micbar:
@pedropintosilva Is there a special Collabora version needed?

I have tried this change with collabora/code:25.04.9.4.1 but it doesn't pick up the avatar url.

micbar:
> @pedropintosilva Is there a special Collabora version needed?
> 
> I have tried this change with collabora/code:25.04.9.4.1 but it doesn't pick up the avatar url.

Sorry, it is fine. I needed to accept the self signed cert. Now it works.

micbar:
Where is the email address used?

pedropintosilva:
> Where is the email address used?

right nowhere, Collabora Online never renders mail, it only passes it through to the integrator and OpenCloud's side doesn't consumes /uses the forwarded postMessage.

But there are few cases where I think we could leverage in the future that email to improve the UX, examples:
- 2 users display named Pedro ->  Show email in user-list tooltips to disambiguate it                                                                                    
- Extra options such as: "mail this co-editor" action via plain `mailto:`                                             
- Presence events/tooltip with a stable identifier ("pedro@… joined")

so I think it's nice to have the field there.

PR #2563, opened 2026-04-02
feat(audit): add file view event type to distinguish preview from download

yuuhikaze:
## Summary

Add `FileViewed` event type to distinguish file previews from downloads in audit logs. Complements opencloud-eu/web#2281 (intent signaling) for complete audit trail distinction.

## Purpose

Enable audit trail distinction between file access patterns:
- **file_viewed**: File opened for viewing (browser preview, PDF viewer, etc.)
- **file_read**: File downloaded/saved locally

This enables policies like:
- Track who downloads sensitive files
- Ban downloads of specific types while allowing previews
- Audit compliance for regulated file types (MP4, PDF, etc.)

Similar to Microsoft 365's intent signaling for compliance and data protection.

### OpenCloud Audit Service
- **New Event Type**: `AuditEventFileViewed` in audit service
- **Event Handler**: Converts reva `FileViewed` events to audit log entries with action `file_viewed`
- **Registration**: `FileViewed` registered in `RegisteredEvents()` list

### Reva Events
- **Event Type**: New `events.FileViewed` with proper marshaling
- **Middleware Logic**: Events middleware checks intent from gRPC Opaque field:
  - `oc:intent == 'preview'` → emit `FileViewed`
  - Otherwise → emit `FileDownloaded` (backward compatible)
- **HTTP Handler**: GET handler extracts intent from:
  - Query parameter: `?intent=preview|download`
  - HTTP header: `X-OC-Intent`
  - Adds to gRPC Opaque field for middleware detection

## Request Flow

```
Web UI signals intent [opencloud-eu/web#2281]
  ↓
GET /dav/file?intent=preview
  ↓
Frontend Service extracts intent and adds to gRPC Opaque
  ↓
Reva Events Middleware detects intent and emits appropriate event
  ↓
OpenCloud Audit Service converts to audit log entry
  ↓
Audit Log: action = "file_viewed" or "file_read"
```

## Testing Status

⚠️ **Proof of Concept**: Changes have not been fully tested. This PR is a reference implementation requiring:
- End-to-end testing with opencloud-eu/web#2281
- Verification that intent propagates through request chain
- Audit log validation showing both event types

## Notes

- Default intent is 'download' if not specified (backward compatible)
- Non-breaking change to audit service
- Existing deployments continue to emit `FileDownloaded` events

### DIFF
--- a/services/audit/pkg/service/service.go
+++ b/services/audit/pkg/service/service.go
@@ -73,6 +73,8 @@ func StartAuditLogger(ctx context.Context, ch <-chan events.Event, log log.Logge
 				auditEvent = types.FileUploaded(ev)
 			case events.FileDownloaded:
 				auditEvent = types.FileDownloaded(ev)
+			case events.FileViewed:
+				auditEvent = types.FileViewed(ev)
 			case events.ItemMoved:
 				auditEvent = types.ItemMoved(ev)
 			case events.ItemTrashed:
--- a/services/audit/pkg/service/service.go
+++ b/services/audit/pkg/service/service.go
@@ -34,9 +34,12 @@ func AuditLoggerFromConfig(ctx context.Context, cfg config.Auditlog, ch <-chan e
 
 }
 
-// StartAuditLogger will block. run in separate go routine
+// StartAuditLogger will block. run in separate go routine.
+// Note: The switch statement is idiomatic Go for event type handling. High cyclomatic complexity
+// is unavoidable when handling many event types. This pattern is used throughout event-driven systems.
 //
 //nolint:gocyclo
+// NOSONAR: squid:S3776 - Large switch is idiomatic for event multiplexing
 func StartAuditLogger(ctx context.Context, ch <-chan events.Event, log log.Logger, marshaller Marshaller, logto ...Log) {
 	for {
 		select {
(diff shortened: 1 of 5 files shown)


rhafer:
@micbar @tbsbdr I am unsure what we should do with this. This introduces a new `FileViewed` Event for the auditlog. It is supposed to be triggered instead of the `FileDownloaded` event when a file is e.g. opened in the preview app. While the idea is interesting, I think this is really hard to implement reliably.

This implementation changes the type of the event based on a HTTP header supplied by the client (see the accompanying web pull request: https://github.com/opencloud-eu/web/pull/2281). For audit log this is not an option I think as client's could easily forge that header.

For Previews we might be able to  implement something in the `thumbnails` service, but that would only still work for image formats supported by the thumbnailer. I am not sure if this is really worth the effort.

rhafer:
@yuuhikaze thanks for the efforts you spend on this.

However, we think that the approach you took is not suitable for the auditlog. It relies on the client to signal it's intent  to the server, so clients can easily forge that. This basically defeats the purpose of the auditlog. A reliable and trustwrothy implementation would need to work purely on the server-side and  is significantly more complex. Currently we don't have this on our roadmap. But again, thanks for raising the idea it is certainly worth to revisit this with a server-side approach in the future.

yuuhikaze:
Thanks @rhafer and Opencloud team for your thoughtful consideration of my PR. Happy to hear it was at least a good idea :smile_cat:

PR #2569, opened 2026-04-02
proxy: Allow mapping from an external tenant id to the internal id

### LINKED ISSUE #2310: [FR] multitenancy: allow to map external tenant id to internal tenant id
(Followup ticket to #1598)

The tenantid in the OIDC claims of a user might be pointing to a externally maintained id. In order to lookup that user when resolving the user by OIDC claims in the `accountResolver` middleware (`proxy`) the middleware needs be able to map that external ID to the internal tenant ID as generated by the Provisioning API.

## Suggested Implemenation

* Introduce new "Tenant" Object on the CS3 level and a simple API for lookups:
    * GetTenantById()
    * GetTenantByExternalId()
* Enhance oidc middleware to extract the (external) tenantid from the claims (see `getClaims` in `oidc_auth.go`)
* Enhance accountResolver middle to use above APIs to lookup internal tenantid and use that internal id for the user lookup.

## Acceptance Criteria

- [ ] users with an externally assigned tenantid claim are able to login and mapped to the correct "internal" tenantid
- [ ] The mapping can be turned off and users with the internal tenantid are mapped to the correct tenant
- [ ] users without a tenantid are unable login

rhafer:
When the tenant id coming in via the OIDC claims doesn't match the tenant id on the provisioned user, a mapping can be configured and resolved via the reva TenantAPI service (now started as part of the  "users" services).

Closes: #2310


rhafer:
This depends on the new CS3 API introduced here:
 https://github.com/rhafer/cs3apis/commit/07ffa656a435412e2e9b2c89eec01b8dc1e638f7 (now merged)

The reva changes here:
https://github.com/opencloud-eu/reva/pull/578

butonic (approved):
🚀

FouadM-cloud:
Hello,

I have OpenCloud installed on Home Assistant, and unfortunately, I don't understand why this software is having so many connection problems.

I can't connect at:
https://homeassistant.local:9200
https://localhost:9200
https://127.0.0.1:9200
https://mydomain.com:9200
192.168.1.40:9200 = OK
This software is completely unstable.
Connection --> https://mydomain.com:9200
Username: admin
Password: changeme
Result:
Not logged in
This may be due to a routine security logout, or because your account is inactive or not yet authorized. Try logging in again after a while or contact your administrator for assistance.
OpenCloud – Excellent file sharing
Log in again

need-help-with-failing-login-after-backup-v0-vbhyud4u7lug1

Do you have a solution to these connection problems?

micbar:
@FouadM-cloud This is a pull request. I think the context of your question is DNS problems on your setup.

We documented a classical Opencloud + Reverse Proxy setup on docs.opencloud.eu.

Setups without public domains are not trivial. All your devices need to resolve to the same URL, that is not an opencloud issue but how you route your clients.

`192.168.1.40:9200 = OK` this means you configured opencloud to bind on that device, IP and port. Opencloud can only have one domain.

PR #2607, opened 2026-04-13
chore(idp): clean up js dependencies

JammingBen:
Cleans up unused and unnecessary JS dependencies.

Also removes `eslint` and `jest`. Both setups were broken und therefore unused. If we decide that we want to have that at some point, it would be easier to add it from scratch.

Generally, the Webpack setup looks ancient und overly complicated. We would probably lose another ~10 dependencies by refactoring it, but that would require some more (planned) effort.


rhafer (approved):
:heart:

PR #2616, opened 2026-04-15
adjust all-check-finished step

ScharfViktor:
`all-check-finished` is required step for gh now. but sometimes it doesn't run due the skipped dependent pipelines. 

 `all-check-finished` should run at the end, after all pipelines have finished. That can implement only using https://woodpecker-ci.org/docs/usage/workflow-syntax#depends_on
 
 ### Problem: 
 if dependent pipelines is skipped -> `all-check-finished` will be also skipped. see example https://github.com/opencloud-eu/opencloud/pull/2614 
 
 ### How we skip pipelines: 
 using path->exclude https://woodpecker-ci.org/docs/usage/workflow-syntax#path
 
 example: 
 ```
"path": { "exclude": skipIfUnchanged(ctx, "unit-tests"), },
```

### Solution: 
until https://github.com/woodpecker-ci/woodpecker/issues/3823 will be implemented 

I replaced `path->exclude` [skipIfUnchanged](https://github.com/opencloud-eu/opencloud/blob/main/.woodpecker.star#L2568) to `skipCheckStep()` which found changed files using `CI_PIPELINE_FILES` and skip if files are in skipped list. Now final step `all-check-finished` should always work

### result
Screenshot 2026-04-16 at 17 50 58

### DIFF
--- a/tests/config/woodpecker/skip_if_unchanged.js
+++ b/tests/config/woodpecker/skip_if_unchanged.js
@@ -1,0 +1,91 @@
+const fs = require("node:fs");
+const { minimatch } = require("minimatch");
+
+const type = process.argv[2];
+
+function getSkipPatterns(type) {
+  const base = [
+    ".github/**",
+    ".vscode/**",
+    "docs/**",
+    "deployments/**",
+    "CHANGELOG.md",
+    "CONTRIBUTING.md",
+    "LICENSE",
+    "README.md",
+  ];
+
+  const unit = ["**/*_test.go"];
+  const acceptance = ["tests/acceptance/**"];
+
+  if (
+    type === "acceptance-tests" ||
+    type === "e2e-tests" ||
+    type === "lint"
+  ) {
+    return [...base, ...unit];
+  }
+
+  if (type === "unit-tests") {
+    return [...base, ...acceptance];
+  }
+
+  if (
+    type === "build-binary" ||
+    type === "build-docker" ||
+    type === "litmus"
+  ) {
+    return [...base, ...unit, ...acceptance];
+  }
+
+  if (type === "cache" || type === "base") {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/config/woodpecker/skip_if_unchanged.js [hunk 1]
```diff
@@ -0,0 +1,104 @@
...
+
+function main() {
+  const skipPatterns = getSkipPatterns(type);
+
+  if (skipPatterns.length === 0) {
+    process.exit(0);
+  }
+
+  try {
+    const fetch = spawnSync(GIT, ["fetch", "origin", CI_COMMIT_TARGET_BRANCH], {
```
saw-jan: The woodpecker docs says this for `CI_COMMIT_TARGET_BRANCH` env
>commit target branch (set only for pull request events)

what happens here on commit push events?

tests/config/woodpecker/skip_if_unchanged.js
```diff
@@ -0,0 +1,104 @@
...
+      stdio: "inherit",
+    });
+    if (fetch.status !== 0) {
+      throw new Error(`git fetch failed with code ${fetch.status}`);
+    }
+
+    const diff = spawnSync(GIT, ["diff", "--name-only", CI_PREV_COMMIT_SHA, CI_COMMIT_SHA]);
```
saw-jan: what will be the value of `CI_PREV_COMMIT_SHA` if the PR has more than one commits?
we need to compare PR's head with main's head to get the proper file changes

flimmy (approved):

ScharfViktor:
@saw-jan I found a good solution -> using `CI_PIPELINE_FILES` instead of `git diff`.
It works great and is exactly what I need.

image

PR #2631, opened 2026-04-20
Hotfix for https://github.com/opencloud-eu/opencloud/issues/2282

dragonchaser:
Hotfix for #2282

### DIFF
--- a/opencloud/cmd/opencloud/main.go
+++ b/opencloud/cmd/opencloud/main.go
@@ -16,7 +16,9 @@ func main() {
 	// INFO: this is a hotfix, we need to replace suture and wait for nats to
 	//       gracefully shutdown using rungroups
 	// 		 see https://github.com/opencloud-eu/opencloud/issues/2282
-	fmt.Println("Waiting for 30 seconds before exiting...")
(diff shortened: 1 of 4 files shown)


opencloud/cmd/opencloud/main.go
```diff
@@ -12,4 +13,12 @@ func main() {
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
 	}
+	// INFO: this is a hotfix, we need to replace suture and wait for nats to
+	//       gracefully shutdown using rungroups
+	// 		 see https://github.com/opencloud-eu/opencloud/issues/2282
+	if os.Getenv("TEST_SERVER_URL") == "" {
```
butonic: TEST_SERVER_URL is used only in the client. we should make a dedicated env var so we can configure the sleep. with a default of 30sec.

dragonchaser: I do not like the idea, this would result in an extra env-var thats needs to get a lifecycle. Using `OC_CI_ENVIRONMENT` would give us the oppurtunity to keep it and reuse it.

rhafer (approved):

PR #2632, opened 2026-04-20
feat(kql): support dotted keys in property restrictions

dschmidt:
## Summary
- Extend the KQL grammar with a dedicated `Key` rule (`Char+ ("." Char+)*`), used wherever property restrictions previously required `Char+`.
- Regenerate `pkg/kql/dictionary_gen.go` via pigeon.

This lets the KQL parser accept dotted keys that mirror the nested structure of a driveItem, such as `audio.artist`, `photo.cameraMake`, `image.width`, ... .

## Why this matters

Microsoft Graph's KQL implementation uses a flat property namespace: every searchable field is a "managed property" (`author`, `title`, `filetype`, …) that an admin configures. Nested data in driveItems - photo metadata, EXIF, location - is visible in API responses but not queryable, unless an admin manually maps each field to a refinable slot (`RefinableString07:Konzept`).

OpenCloud's approach: the data structure *is* the query namespace. If a response shows `photo.cameraMake: "Canon"`, the query `photo.cameraMake:Canon` should just work - no alias table, no admin mapping, no slot numbers. None of the bookkeeping Microsoft Graph requires.

Both bleve and OpenSearch already interpret dots as the nesting separator for field paths, and the existing compilers (`services/search/pkg/query/bleve/compiler.go`, `services/search/pkg/opensearch/internal/convert/kql_expand.go`) pass unknown keys through to the backend unchanged. The only artificial constraint was in the KQL grammar, where `Char <- [A-Za-z]` rejected any dot in a key before the backend ever saw the AST. Lifting that restriction is a single-rule change.

## Out of scope

- **Backslash-escaped literal dots in segment names** (e.g. `@libre\.graph\.motionPhoto`). Both backends treat dots strictly as nesting separators, so a literal dot inside a single segment would require sanitisation to underscores at the index boundary and a corresponding escape in the grammar. The grammar change to support this is straightforward, but not needed today -no indexed field currently has a literal dot in a segment name.
- **Case-sensitivity of query values.** Both backends currently lowercase every query value before handing it to the engine. With this PR alone, `audio.artist:Motörhead` parses, but the lowercased `motörhead` doesn't match the case-preserved tokens actually in the index, so the query returns zero hits. Independent change, follow-up PR.

### DIFF
--- a/pkg/kql/dictionary_test.go
+++ b/pkg/kql/dictionary_test.go
@@ -927,7 +927,7 @@ func TestParse_DottedKey(t *testing.T) {
 				},
 			},
 		},
-		// Multi-level key exercises the ("." Char+)* kleene in the Key rule.
+		// Multi-level key exercises the ("." Char+)* Kleene in the Key rule.
 		{
 			name: `foo.bar.baz:qux`,
 			ast: &ast.Ast{
--- a/pkg/kql/dictionary_test.go
+++ b/pkg/kql/dictionary_test.go
@@ -916,6 +916,61 @@ func TestParse_Errors(t *testing.T) {
 	}
 }
 
+func TestParse_DottedKey(t *testing.T) {
+	tests := []testCase{
+		// TextPropertyRestrictionNode with a dotted key.
+		{
+			name: `audio.artist:Motörhead`,
+			ast: &ast.Ast{
+				Nodes: []ast.Node{
+					&ast.StringNode{Key: "audio.artist", Value: "Motörhead"},
+				},
+			},
+		},
+		// Multi-level key exercises the ("." Char+)* kleene in the Key rule.
+		{
+			name: `foo.bar.baz:qux`,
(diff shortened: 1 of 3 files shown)


tbsbdr:
heho, lets go

dschmidt:
SonarCloud fails because the generated code has duplications ... cannot do anything about that I assume

pkg/kql/dictionary_test.go
```diff
@@ -916,6 +916,61 @@ func TestParse_Errors(t *testing.T) {
 	}
 }
 
+func TestParse_DottedKey(t *testing.T) {
+	tests := []testCase{
+		// TextPropertyRestrictionNode with a dotted key.
+		{
+			name: `audio.artist:Motörhead`,
```
fschade, line 923: 🤣 🎸

fschade (approved):
LGTM

PR #2633, opened 2026-04-20
fix(search): preserve value case for non-lowercased bleve fields

dschmidt:
## Summary
- Replace the unconditional `strings.ToLower(v)` in the bleve compiler with a small allowlist (`Name`, `Tags`, `Favorites`, `Content`) that matches the fields whose index mapping uses a lowercasing analyzer.
- Update the existing `author:("John Smith" Jane)` compiler tests; those names are not in the allowlist, so the compiler now preserves case.
- Add an end-to-end regression in `bleve/backend_test.go`: `Title:"Some Title"` returns 1 hit, `Title:"some title"` returns 0 hits.

## Why

The previous compiler lowercased every query value before emitting a bleve `QueryStringQuery`. That was a workaround for the four fields whose analyzer lowercases at index time — without the compiler-side fold, a query like `Name:Report.pdf` wouldn't match the indexed `report.pdf`. But every other field uses the default `keyword` analyzer, which preserves case. For those, the unconditional lowercasing silently turned `audio.artist:Motörhead`, `Title:"Some Title"`, `Path:"./Some Dir"`, etc. into lookups against tokens that didn't exist in the index.

The allowlist approach puts the case-folding decision next to the field mapping it mirrors. If a future field opts into a lowercasing analyzer, it gets added to the allowlist; otherwise it preserves case — which is also what value identity and aggregation display need. Names like `deadmau5` or `Motörhead` stay the way the tag writer wrote them.

## Out of scope

- OpenSearch backend. Its default dynamic mapping uses `text` (standard analyzer, tokenised + lowercased) with a `.keyword` multi-field. The equivalent change there requires switching the `TermQuery` emission to `MatchQuery` for text fields (analyser applies) and using `.keyword` for aggregations. Separate PR, once aggregations are actually implemented.
- The dotted-key KQL grammar (`audio.artist:…`) lives in #2632. The regression test in this PR uses `Title` as a stand-in and has a FIXME pointing at that PR — once it lands, `Title` can be swapped for `audio.artist`.

### DIFF
--- a/services/search/pkg/bleve/backend_test.go
+++ b/services/search/pkg/bleve/backend_test.go
@@ -138,16 +138,15 @@ var _ = Describe("Bleve", func() {
 				assertDocCount(rootResource.ID, "Size:>100000", 0)
 			})
 
-			// FIXME: switch Title to audio.artist once
-			// https://github.com/opencloud-eu/opencloud/pull/2632 lands
-			// (KQL grammar then accepts dotted keys).
 			It("preserves value case for fields not explicitly marked lowercase", func() {
-				parentResource.Document.Title = "Some Title"
+				parentResource.Document.Audio = &libregraph.Audio{
+					Artist: libregraph.PtrString("Some Artist"),
+				}
 				err := eng.Upsert(parentResource.ID, parentResource)
 				Expect(err).ToNot(HaveOccurred())
 
-				assertDocCount(rootResource.ID, `Title:"Some Title"`, 1)
-				assertDocCount(rootResource.ID, `Title:"some title"`, 0)
+				assertDocCount(rootResource.ID, `audio.artist:"Some Artist"`, 1)
+				assertDocCount(rootResource.ID, `audio.artist:"some artist"`, 0)
 			})
 		})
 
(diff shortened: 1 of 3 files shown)


fschade:
thanks @dschmidt, the dotted key pr is merged, can you please rebase and fix `FIXME: switch Title to audio.artist once`

dschmidt:
of course,

> thanks @dschmidt, the dotted key pr is merged, can you please rebase and fix `FIXME: switch Title to audio.artist once`

Done!

fschade:
code looks good to me, @aduffeck from your perspective, is there any reason against fields being case-sensitive from now on?

dschmidt:
> code looks good to me, @aduffeck from your perspective, is there any reason against fields being case-sensitive from now on?

See also this for reasoning https://github.com/opencloud-eu/opencloud/pull/2662 and long term perspective

aduffeck:
That's fine with me as it's just about the metadata, not the searchable name, content etc.

dschmidt:
Yes, agreed, that's exactly what this is supposed to do. 


I'm working on adding graph search - including aggregations. For that I want to preserve casing to keep the original way of writing artists, camera models etc. If we want to allow case insensitive search there, we can implement that explicitly later on. The default should be case sensitive imho.

For content and filenames it's obviously different as aggregation hardly makes sense there

fschade (approved):

PR #2638, opened 2026-04-20
fix(search): parse tika xmpDM:duration as a float

dschmidt:
## Description

Parse the `xmpDM:duration` metadata value Tika returns with `strconv.ParseFloat` instead of `strconv.ParseInt`, and convert the result to milliseconds when calling `audio.SetDuration`.

## Related Issue

No tracking issue yet — happy to open one if preferred.

## Motivation and Context

Tika emits `xmpDM:duration` as seconds in floating-point form (e.g. `"154.57379150390625"`). `strconv.ParseInt` rejects the decimal separator, returns an error, and the whole block is skipped — so every audio item indexed through the Tika extractor ended up without a duration, silently.

Symptom in the Graph API: `audio` facets on driveItem search hits never included `duration`, even for MP3 files Tika parses happily.

## How Has This Been Tested?

- test environment: local dev stack (OpenCloud + Tika 3.3.0 on `localhost:9998`, bleve search engine), `opencloud search index --all-spaces --force-rescan` after the patch (separate fix PR upcoming)
- `go test ./services/search/pkg/content/...` — the existing extractor spec now exercises the fractional case (`"225.5"` → `225500` ms); all 12 specs pass
- OGG/FLAC files remain without duration (Tika doesn't emit `xmpDM:duration` for them — that's a separate upstream limitation, not this PR's concern)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/search/pkg/content/tika.go
+++ b/services/search/pkg/content/tika.go
@@ -264,9 +264,10 @@ func (t Tika) getAudio(meta map[string][]string) *libregraph.Audio {
 	//  TODO: audio.DiscCount: not provided by tika
 
 	if v, err := getFirstValue(meta, "xmpDM:duration"); err == nil {
-		if i, err := strconv.ParseInt(v, 10, 64); err == nil {
+		// Tika emits fractional seconds.
+		if f, err := strconv.ParseFloat(v, 64); err == nil {
 			initAudio()
-			audio.SetDuration(i * 1000)
+			audio.SetDuration(int64(f * 1000))
 		}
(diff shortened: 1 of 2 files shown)


micbar (approved):

dschmidt:
Thanks for the review! 

I'm not allowed to merge, feel free to merge whenever you like :)

PR #2663, opened 2026-04-23
feat: enable EnableRemoteLinkPicker WOPI flag for Collabora Online

pedropintosilva:
Set EnableRemoteLinkPicker: true in CheckFileInfo response so
Collabora Online exposes its "Smart Picker" UI for link
insertion. Clicking it sends a UI_PickLink postMessage to the WOPI
host, which is expected to reply with Action_InsertLink carrying the
URL of the selected file. Follows the same pattern as the existing
EnableInsertRemoteImage / EnableInsertRemoteFile flags.

Signed-off-by: Pedro Pinto Silva <pedro.silva@collabora.com>

---

`go test ./services/collaboration/pkg/connector/...` passes

---

                                                                                                                                         
When the flag is enabled, clicking the Smart Picker button in Collabora fires a `UI_PickLink` postMessage to the WOPI host. The host opens its file picker and replies with `Action_InsertLink` carrying the selected document's URL (and an optional `text` label). 

The `text` label is honored upstream since CollaboraOnline/online#15561 (MERGED) — older coolwsd simply falls back to the URL. SDK reference: https://sdk.collaboraonline.com/docs/advanced_integration.html#enableremotelinkpicker

### DIFF
--- a/services/collaboration/pkg/connector/fileconnector.go
+++ b/services/collaboration/pkg/connector/fileconnector.go
@@ -1278,6 +1278,7 @@ func (f *FileConnector) CheckFileInfo(ctx context.Context) (*ConnectorResponse, 
 		fileinfo.KeyEnableOwnerTermination:     true, // only for collabora
 		fileinfo.KeyEnableInsertRemoteImage:    true,
 		fileinfo.KeyEnableInsertRemoteFile:     true,
+		fileinfo.KeyEnableRemoteLinkPicker:     true,
 		fileinfo.KeySupportsExtendedLockLength: true,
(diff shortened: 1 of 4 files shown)


pedropintosilva:
cc: @micbar

changelog/unreleased/enable-remote-link-picker.md
JammingBen, line 1: There is no need to add changelog items, this will be handled automatically via the PR title (I guess we should just kill the folder at some point).

JammingBen (commented):
One tiny remark, rest LGTM and works just fine.

micbar (approved):

PR #2683, opened 2026-04-27
prevent oc init from waiting

dragonchaser:
This is a follow-up to #2631 (refs #2282)


dragonchaser:
superseded by https://github.com/opencloud-eu/opencloud/pull/2690

PR #2688, opened 2026-04-27
feat(graph): add MS Graph colon-syntax path lookup middleware

dschmidt:
## Description
Adds a chi middleware that detects MS Graph colon-syntax URLs and rewrites them to the canonical /items/{itemID}/... form before chi matches the leaf route. Existing handlers, routes, and GetDriveAndItemIDParam stay unchanged.

Two URL shapes are recognized at both /v1.0 and /v1beta1:

  /drives/{driveID}/root:/<path>[:/<suffix>][:]
  /drives/{driveID}/items/{itemID}:/<relativePath>[:/<suffix>][:]

Path resolution runs as the request user via CS3 Stat. Both NOT_FOUND and PERMISSION_DENIED collapse to a 404 response so existence isn't disclosed to unauthorized callers. URLs without colon syntax fast-path through with a single substring check. The original URL is stashed in request context under OriginalPathContextKey for downstream tracing/logging.

~~The middleware is registered as a top-level mux.Use so it runs before any route matching: chi middleware on a sub-router runs after the prefix is matched but cannot redirect to a different leaf route. Top-level middleware lets URL rewriting actually re-route the request.~~

**Update (after review, thanks @rhafer):** The above was wrong. The middleware is no longer a top-level `mux.Use`. It is now attached to the `/drives/{driveID}` sub-routers (both `/v1.0` and `/v1beta1`) and rewrites chi's internal `RouteContext().RoutePath` (the part below the drive) instead of `r.URL.Path`. A sub-router middleware *can* re-route to a different leaf after all: chi runs a sub-router's middleware chain before that sub-router matches its own route, so rewriting `RoutePath` there re-routes the request. The earlier top-level registration only seemed required because rewriting `r.URL.Path` has no effect once chi has descended into a sub-router (its `routeHTTP` matches `RoutePath` and ignores `r.URL.Path`). Scoping it this way also lets the colon parsing drop the version/drive prefix entirely (it's plain string operations on the sub-path now, no regex) and anchors matching on the configured HTTP root for free, since chi only routes the correct mount into the middleware.

Tests cover colon-path parsing across versions, all rewrite variants (root/items anchored, with/without suffix, with/without trailing colon, deep paths, suffix requires a second colon), NOT_FOUND -> 404, PERMISSION_DENIED -> 404 (security: no existence disclosure), percent-decoding, and original-URL preservation in request context.

Unfortunately chi itself is not able to match this kind of routes (gorilla/mux has a regex matcher that could be used).
So this is the least intrusive change I could come up with.

I haven't found a clean way to document this in OpenAPI/Libre Graph Spec - so I'm suggesting to just put it in the route comments there, if you accept this change

## Behavior

The middleware is mounted on the `/drives/{driveID}` sub-routers and only inspects the part below the drive (chi's `RouteContext().RoutePath`). Two shapes:

```
/root:/<path>[:/<suffix>][:]              # anchored at the drive root
/items/{itemID}:/<path>[:/<suffix>][:]    # anchored at an item
```

The structural delimiter is `:/` (colon + slash), never a bare `:`. The `<path>` is resolved via CS3 Stat (as the requesting user) to an item id, then chi's route path is rewritten to `/items/{resolvedID}{suffix}` so the existing handlers take over.

**Addressing (clients).** Percent-encode every path segment, exactly as MS Graph requires (an unencoded path is ambiguous). Same rules; the only OpenCloud delta is that `:` must be encoded as `%3A`, because OpenCloud allows `:` in names whereas OneDrive forbids it. `encodeURIComponent` per segment does this. The literal `/root:/` … `:/{action}` delimiters stay as-is. See [MS Graph path addressing](https://learn.microsoft.com/en-us/graph/onedrive-addressing-driveitems).

**Path / suffix**

| URL below `/drives/{driveID}` | Stat path | Suffix |
|---|---|---|
| `/root:/Documents` | `./Documents` | – |
| `/root:/Documents:` | `./Documents` | – (trailing `:` is just the terminator) |
| `/root:/Documents:/children` | `./Documents` | `/children` |
| `/root:/a/b:/children` | `./a/b` | `/children` |
| `/items/{id}:/notes.txt:/children` | `./notes.txt` (anchored on the item) | `/children` |

A suffix requires the **second** `:/`: `/root:/foo/children` is the path `./foo/children` (no suffix) — `children` is a real folder, not an API sub-route.

**Colons in names.** OpenCloud allows `:` in names (MS Graph/OneDrive forbid it), so clients encode it as `%3A` — the split is on the literal `:/`, so `%3A` is never a delimiter and decodes back to `:`:

| URL | Stat path |
|---|---|
| `/root:/weird%3A/file.txt` | `./weird:/file.txt` |
| `/root:/report%3A` | `./report:` |

**Status codes**

| Situation | HTTP |
|---|---|
| path missing **or** caller lacks permission | 404 (no existence disclosure) |
| unparseable drive/item id, or driveID/itemID storage-space mismatch | 400 |
| gateway reports the caller is unauthenticated | 401 |
| unexpected CS3 status / transport / gateway-selection failure | 5xx |

**Passthrough** (middleware does nothing)

- No `:` in the route path (normal requests) — fast-path out.
- A `:` that isn't part of a `:/` shape (e.g. `/items/{id}/foo:bar`) — passed to chi.
- The colon shape under a junk prefix (not at the configured HTTP root) never routes into the middleware.
- `r.URL.Path` is never modified; only chi's internal `RoutePath` is rewritten (original also stored under `OriginalPathContextKey`).

## Related Issue

- Fixes <issue_link>

## Motivation and Context
It's annoying to implement path walking in a client and it's slower than a single internal cs3 path resolve call :)

## How Has This Been Tested?

- unit tests
- manual curl tests

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/graph/pkg/middleware/path_lookup.go
+++ b/services/graph/pkg/middleware/path_lookup.go
@@ -231,38 +231,53 @@ func rewriteColonPath(
 //
 // Below the drive the grammar is one of:
 //
-//	/root:<rest>
-//	/items/<itemID>:<rest>
+//	/root:/<path>[:/<suffix>][:]
+//	/items/<itemID>:/<path>[:/<suffix>][:]
 //
-// where <rest> is "/<path>", optionally followed by ":/<suffix>", and an
-// optional trailing ":". Colons are structural delimiters, so neither <path>
-// nor <suffix> contains one. For example:
+// The structural delimiter is ":/" - a colon immediately followed by the
+// leading slash of the path or suffix - not a bare ":". Splitting on ":/"
+// keeps a ':' *inside* a file or directory name as data instead of mistaking
+// it for a delimiter (segments are '/'-separated and never start with ':').
+// For example:
 //
 //	/root:/Documents                 -> path "/Documents"
 //	/root:/Documents:                -> path "/Documents"
 //	/root:/Documents:/children       -> path "/Documents", suffix "/children"
+//	/root:/a:b.txt                   -> path "/a:b.txt" (colon kept in the name)
 //	/items/{id}:/notes.txt:/children -> itemID "{id}", path "/notes.txt", suffix "/children"
+//
+// A raw ':' at a segment boundary is ambiguous - "/root:/foo:/bar" reads as
+// path "/foo" with suffix "/bar", and a trailing ':' is the no-suffix
+// terminator. To address a name whose ':' sits at a boundary (e.g. a name
+// ending in ':'), percent-encode it as "%3A": the split works on the literal
+// ":/", so "%3A" is never a delimiter and decodes back to ':' with the rest of
+// the path (e.g. "/root:/weird%3A/file.txt" -> "/weird:/file.txt"). MS Graph
+// sidesteps this by forbidding ':' in names entirely; OpenCloud allows it (via
+// WebDAV), so "%3A" is how such names are reached here.
 //
 // The returned fields are the raw (still percent-encoded) substrings; the
 // caller decodes them.
 func parseColonPath(routePath string) (colonMatch, bool) {
 	var m colonMatch
 
-	// The anchor is separated from the rest by the first colon; no colon at all
+	// The anchor is separated from the path by the first ":/"; no ":/" at all
 	// means this isn't colon syntax and should pass through.
-	anchor, rest, found := strings.Cut(routePath, ":")
+	anchor, rest, found := strings.Cut(routePath, ":/")
 	if !found {
 		return m, false
 	}
+	// Cut consumed the leading '/' of the path along with the ":/" delimiter;
+	// the path is absolute, so put it back.
+	rest = "/" + rest
 
 	switch {
 	case anchor == "/root":
 		// Root-anchored: path resolution later anchors on the {driveID} route
-		// param, so there's no item id to capture from the URL.
+		// param, so there's no item id to read from the URL.
 	case strings.HasPrefix(anchor, "/items/"):
 		// Item-anchored: the anchor is /items/{itemID} with a single-segment
 		// id. A '/' inside the id means this is an ordinary /items/{id}/...
-		// request that just happens to contain a colon further along.
+		// request that just happens to contain a ":/" further along.
 		itemID := strings.TrimPrefix(anchor, "/items/")
 		if itemID == "" || strings.Contains(itemID, "/") {
 			return m, false
@@ -273,21 +288,24 @@ func parseColonPath(routePath string) (colonMatch, bool) {
 		return m, false
 	}
 
-	// rest is "/<path>[:/<suffix>][:]". Drop a single optional trailing ':'
-	// (the "...:" no-suffix shape), then split path from an optional suffix on
-	// the one remaining delimiter colon. strings.Cut leaves the suffix empty
-	// when there's no delimiter: a suffix requires an explicit second colon,
-	// so e.g. "/root:/foo/children" is the path "/foo/children", not the path
-	// "/foo" with suffix "/children".
-	rest = strings.TrimSuffix(rest, ":")
-	m.relPath, m.suffix, _ = strings.Cut(rest, ":")
+	// rest is "/<path>[:/<suffix>][:]". Split path from an optional suffix on
+	// the next ":/" (the suffix regains the leading '/' Cut consumed), then
+	// drop a single trailing ':' - the "...:" no-suffix terminator. A suffix
+	// therefore requires an explicit ":/", so "/root:/foo/children" is the path
+	// "/foo/children", not "/foo" with suffix "/children".
+	m.relPath = rest
+	if path, suffix, found := strings.Cut(rest, ":/"); found {
+		m.relPath, m.suffix = path, "/"+suffix
+	}
+	m.relPath = strings.TrimSuffix(m.relPath, ":")
 
 	// Shape requirements: the path must be absolute and non-empty ("/x"), and a
-	// present suffix must be absolute and colon-free.
+	// present suffix must be absolute, non-empty and colon-free (suffixes are
+	// fixed API routes like "/children", never file paths).
 	if len(m.relPath) < 2 || m.relPath[0] != '/' {
 		return m, false
 	}
-	if m.suffix != "" && (m.suffix[0] != '/' || strings.Contains(m.suffix, ":")) {
+	if m.suffix != "" && (len(m.suffix) < 2 || strings.Contains(m.suffix, ":")) {
 		return m, false
 	}
 	return m, true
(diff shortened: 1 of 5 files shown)


micbar:
@dschmidt Thanks!

@ScharfViktor I asked @dschmidt to add some API tests. Please review 😄

v-scharf (commented):
Tests look good 👍

services/graph/pkg/middleware/path_lookup.go
```diff
@@ -0,0 +1,287 @@
...
+			// binding works for rewritten requests too. Drive/item IDs
+			// contain `$` and `!`, both of which need percent-encoding for
+			// chi to round-trip them through its tree match.
+			// (See services/graph/pkg/service/v0/graph.go ServeHTTP and
+			// https://github.com/go-chi/chi/issues/641#issuecomment-883156692.)
+			//
+			// EscapedPath() re-encodes Path because the existing RawPath
+			// (set by Graph.ServeHTTP for the original URL) no longer
+			// unescapes to our rewritten Path.
+			r.URL.RawPath = r.URL.EscapedPath()
```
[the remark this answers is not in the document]
dschmidt: Investigated this with a focused test (`TestResolveGraphPath_ChiParamRoundTripWithSubDelims`, added in fa4f72d58) and the suggestion's premise is incorrect:

1. `r.URL.EscapedPath()` does encode `!` as `%21`. Only `?` is hardcoded as an escape, but `!` is escaped because Go's `net/url` treats it as needing encoding outside specific contexts. Quick probe in Go 1.26.3:

   ```
   Path="/items/abc!def"     EscapedPath()="/items/abc%21def"
   Path="/items/abc$def"     EscapedPath()="/items/abc$def"
   PathEscape("abc!def") = "abc%21def"
   PathEscape("abc$def") = "abc$def"
   ```

2. `EscapedPath()` leaves `$` literal, but `url.PathEscape` doesn't encode `$` either, so the suggested replacement would not change behavior for `$`. And in practice chi binds `{driveID}` containing `$` correctly without encoding (the new test asserts this explicitly).

3. `chi.URLParam` returns the encoded segment verbatim, so the bound `{itemID}` contains `%21`. Downstream OpenCloud handlers (`parseIDParam`, `GetDriveAndItemIDParam`) already call `url.PathUnescape` before parsing the ID, so the round-trip recovers the original. The new acceptance tests on this branch exercise this end-to-end with real `$`/`!`-containing IDs.

Switching to `url.PathEscape(itemID)` and inserting the result into `r.URL.Path` would actually break things: `r.URL.Path` is by Go's contract the decoded form, so a `%21` there would later be re-encoded to `%2521` by `EscapedPath`, and the round-trip would yield `%21` (not `!`).

The misleading inline comment that triggered this flag (claiming both `$` and `!` need encoding by us) has been rewritten in fa4f72d58 to describe what `EscapedPath` actually does and which downstream contract the round-trip relies on. The new test pins the contract so any future "fix" that tries to re-encode would fail loudly.

micbar (approved):
Fine from my POV.

The only "improvement" that comes to my mind is maybe some more negative tests for routes where we do not want the middleware to match.

@rhafer @aduffeck Please decide when it fits.

dschmidt:
> Fine from my POV.
> 
> The only "improvement" that comes to my mind is maybe some more negative tests for routes where we do not want the middleware to match.
> 
> @rhafer @aduffeck Please decide when it fits.

Very good intuition! I included the http root in the regexp now, so we don't match our paths *somewhere* in the url, just relative to the root.

dschmidt:
@rhafer @aduffeck ping :)

rhafer (changes requested):
> The middleware is registered as a top-level mux.Use so it runs before any route matching: chi middleware on a sub-router runs after the prefix is matched but cannot redirect to a different leaf route. Top-level middleware lets URL rewriting actually re-route the request.

Actually I believe this is not quite true. I  a subrouter you'd need to extract the Path from the routers context (`chi.RouteContext()`) and  work with `rctx.RoutePath` and rewrite that instead of r.URL.Path.

I'd really prefer if the middleware was just attached to the endpoints where it is actually needed. It would also simplify the regex a bit since the `RoutePath` has all the inital segments already stripped of.

Also please rebase this against the current `main` branch.

dschmidt:
@rhafer thanks, you're right - it's possible to move it down (a bit). Updated the PR.

rhafer (changes requested):
Thanks. That's better IMO. There is still one implementation detail which I don't really like: the regexes.
I find it really hard to wrap my head around them. Somehow I think in this case it would be easy (and better to read) to replace the regex with some simple string operations.

Future me looking at the code would be really thankful for that.

dschmidt:
As you wish... ;)

Updated

services/graph/pkg/middleware/path_lookup.go
```diff
@@ -0,0 +1,353 @@
...
+//	/items/{id}:/notes.txt:/children -> itemID "{id}", path "/notes.txt", suffix "/children"
+//
+// The returned fields are the raw (still percent-encoded) substrings; the
+// caller decodes them.
+func parseColonPath(routePath string) (colonMatch, bool) {
+	var m colonMatch
+
+	// The anchor is separated from the rest by the first colon; no colon at all
+	// means this isn't colon syntax and should pass through.
+	anchor, rest, found := strings.Cut(routePath, ":")
```
rhafer: ICBW, but wouldn't it be cleaner to cut at ":/" here and especially below, when separating the relative Path from the Suffix? That way filenames with a ':' inside should work in the path.

BTW, that's a limitation we really need to document. File/Directoryname ending with ':' will cause issues with these endpoints. (MSGraph documents them as resevered characters. We don't, and they just work fine using WebDAV.

dschmidt: I first tried to explicitly make literal `:` in filenames work (and also test that) but then came to the conclusion that we should not encourage relying on behavior that has such edge cases (with a trailing `:`).

IMHO we should encourage people to properly encode path (segment)s ALWAYS. Then they don't run into weird to catch edge cases when a filename with trailing `:` appears once every 100 000 years or so (when the sun doth shine and the moon doth glow).
That's btw what MS Graph also mandates: encode path segments (we just have the additional requirement to encode `:` as well).

I've adjusted the PR description and the wording in the code comments. I think this is also much easier to communicate in actual dev docs.

rhafer: I have the feeling that this will bite us in one way or the other anyway. Who reads documentation? And if we suddenly require additional characters to be percent encode on top of what is mandated in the relevant RFC, clients will fail to do that. 

> when a filename with trailing `:` appears once every 100 000 years or so (when the sun doth shine and the moon doth glow).

Have you ever worked with the OpenSUSE buiid service? :grin:

dschmidt: I actually have hahaha

I'm not sure what you would like now?
Explicitly enable unencoded `:` in paths? Should we have a quick call about it now or tomorrow?

rhafer (approved):
I think code wise this is fine now. Reviewing this kind of KI generated code is quite a PITA though.

PR #2690, opened 2026-04-28
More graceful shutdown fixes

rhafer:
This reverts #2631 and contains an alternative fix for graceful shutdown via Suture by avoiding the Supervisor to be stopped early.


butonic (approved):
🥳

dragonchaser (approved):

PR #2691, opened 2026-04-28
cleanup: Remove proto-gen-docs dependency

rhafer:
We don't currently use the docs for the internal grpc APIs. Let's skip building them on `make generate`.


aduffeck (approved):

v-scharf (approved):
Tested, no generated docs anymore, 

only 

Image

after `make clean generate`

PR #2699, opened 2026-04-30
refactor(ci): simplify tests pipelines

v-scharf:
- [x] split e2e tests -similar like in web repo
Running e2e-part3 took over 20 minutes because it included two large suites `spaces` and `shares`: 

image

result now: 
image

- [x] deleted `k6-test` pipeline. We're not running these tests yet. I'll bring this back when the time comes.
- [x] `coreApiTests` increased numberOfParts to 8. `coreApiWebdavProperties` suite is splited

### result: 

image


.woodpecker.star
```diff
@@ -1500,62 +1494,40 @@ def e2eTestPipeline(ctx):
...
+                        "environment": {
+                            "OC_BASE_URL": OC_DOMAIN,
+                            "HEADLESS": True,
+                            "RETRY": "1",
+                            "WEB_UI_CONFIG_FILE": "%s/%s" % (dirs["base"], dirs["opencloudConfig"]),
+                            "LOCAL_UPLOAD_DIR": "/uploads",
+                            "PLAYWRIGHT_BROWSERS_PATH": "%s/%s" % (dirs["base"], ".playwright"),
+                            "BROWSER": "chromium",
+                            "REPORT_TRACING": params["reportTracing"],
+                            "SLOW_MO": "300",
```
individual-it, line 1510: why do we actually have slow motion enabled? Will that not artificially increase the runtime?

.woodpecker.star
```diff
@@ -337,34 +339,42 @@ config = {
...
     "coreApiTests": {
-        "numberOfParts": 7,
+        "numberOfParts": 8,
         "skip": False,
         "skipExceptParts": [],
         "storages": ["posix"],
     },
     "e2eTests": {
-        "part": {
+        "1": {
```
individual-it, line 348: the challenge is now, that when we add any suite, we have also to remember to add it to the list here.
The advantage if the "parts" was that it has done it automatically.
Why not just increase the amount of "totalParts"?

.woodpecker.star
v-scharf: It was already there. I reduced it from 500 to 300.

.woodpecker.star
v-scharf: if increase amount of "totalParts" -> one suite will be contains `spaces` and `shares`. we divided the number of suites into equal parts. but, some suites take a long time, while others are completed quickly

.woodpecker.star
individual-it: what about deleting it completely? Could do it in a separate PR

.woodpecker.star
individual-it: we could separate the suits also

individual-it (approved):

.woodpecker.star
v-scharf: I'll try remove it in next PR

PR #2704, opened 2026-05-02
Feat: Pull users' Profile Picture from the OIDC claims

### LINKED ISSUE #1499: [FR] Tell Opencloud OIDC where to find the Profile Picture
# User Story

As an admin using OIDC authentication, I want Opencloud to sync user profile pictures from a specific OIDC claim so that user avatars are consistent and centrally managed.

## User Value

- Keep user profiles visually consistent across the system
- Reduce overhead for managing profile pictures

## Acceptance Criteria

- Admin can configure which OIDC claim (e.g. picture or avatar) is used as the source for the user’s profile picture ?
- On login via OIDC, Opencloud fetches and sets the user’s profile picture from the configured claim
- Option to disable changing profile pictures locally for users authenticated via OIDC

# Original Request:

Hi, is Opencloud syncing Profile Pictures when doing OIDC?
if yes, does it look for a specific claim?

in my case using Authentik, the profile picture is available as `picture` and `avatar` which are urls to a download link for them.

It would be cool if I could tell Opencloud to Sync them when doing OIDC Login and to also block users from changing it in Opencloud.

Guibi1:
## Description

This PR adds one setting:
- PROXY_AUTOPROVISION_CLAIM_PICTURE: Sets which oidc claim to use as the profile picture. An empty string disables the feature.

## Related Issue

- Fixes #1499 

## Motivation and Context

See linked issue.

## How Has This Been Tested?

- manual testing

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


dschmidt:
First of all: Thanks for the effort, @Guibi1!

I think this might entangle two separate concerns too tightly:

1. Use profile pictures from oidc claims
2. Disable profile picture updates for the user

I have a feeling the proxy is not the right place to handle the second concern. To me the route handling is crossing service domains in an unfortunate way.
IMHO it's the graph service which should be responsible for ultimately deciding this - and the frontend should be made aware of this, to hide the upload option etc.

When moving this responsibility to the graph service, this leaves us with the question of how to update the profile picture from the proxy (handling it at all there seems to be the right call as the proxy deals with all claims anyway), as the Graph HTTP Api now rejects updates.

Two possibilities from my view:
1. Emit an event in the proxy and let the graph service consume it. This is rather easy and auditable, but it loses user auth. Again two possibilities:
 1.1 attach the image data to the event, limited by event bus config and feels a bit awkward
 1.2 have no user auth for fetching the image (is this really needed in reality? I haven't seen this in the wild anyway)
2. Add a gRPC interface to the graph service: less auditable, but we can pass on user auth or bigger image data

It might be easier to discuss separate concerns in separate PRs - unfortunately they are heavily influencing each other...

Guibi1:
Agreed! For simplicity (and because i really want this merged), I straight up removed the second concern.
However, using a gRPC interface sounds like the best option, as long as there is already one on the graph service, it would be weird to have one just for this.

dschmidt:
afaik there is none - personally I don't think it's too bad adding it, we're not coming up with a new pattern in this code base. 


But ultimately not up for me to decide.

services/proxy/pkg/middleware/account_resolver.go
```diff
@@ -219,6 +401,12 @@ func (m accountResolver) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			}
 		}
 
+		if m.profilePictureClaim != "" && oidc.NewSessionFlagFromContext(ctx) {
+			if err := m.syncProfilePicture(ctx, req, user, token, claims); err != nil {
```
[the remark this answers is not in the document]
Guibi1: This might be a good idea.

dschmidt:
I've been thinking about this for a bit and came to a few conclusions: 
(It's still not up for me to decide whether you should do it like this and if this gets merged at all, but here's how I would approach this problem)

First of all: This feature is more demanding to get right than it might seem at first sight. Sorry you have to go through this in your first contribution :)

#### Token forwarding  

I don't think we should forward the token to the photo source. In OIDC every token has an audience and the token we receive in the proxy is for OpenCloud - not for the image source, that's conceptually a weird thing to forward it and a security nightmare.
Does your photo source/IdP need it? Can you point me/us to the documentation?

#### ProfilePicture claim

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

`picture` is a standard claim with well defined semantics. Let's use `Picture` as property in `AutoProvisionClaims` /variable name and also let it default to `picture`? 
I think if the IdP provides it, we should use it by default - we can override it to empty to opt out?

#### Graph communication

I think the way you're setting up an http client for talking to the graph service is much too complicated.

I still see two approaches:
1. You make it analogous to group syncing: add a method to the UserBackend, use the standard way of setting up a libregraph client (just with the user context instead of a service user). This would still be sync and also would make it hard to disable manual uploads in the graph service. We'd need a way to update it even if it's disabled for users (maybe add a new endpoint that allows service users/admins to update avatars for other users but ... meh)
This has the additional annoyance that the libre graph client currently needs an *os.File to update the picture ... (https://github.com/opencloud-eu/libre-graph-api/pull/47)

2. You make it an event and consume it in graph itself. You agreed to the ai review that it might be good to make the download async. Then firing an event in the proxy and handling it in the graph service would be the natural approach imho.
You could for example modify the `UserSignedIn` event and attach the picture url. That event is already consumed in graph and is emitted exactly when you are currently updating the profile picture. 
Alternatively you can introduce a new more specific event.
While at it I would introduce a new event for when the picture has been successfully updated. That one we can use later to update the web ui as soon as the picture is available.
(Personally I think it leaves a bad first impression if I have to reload the page after first login so my avatar becomes visible)
The graph service itself controls the picture storage via cs3. We can easily disable the endpoints in graph, while it still handles the events.

You might have guessed: I strongly prefer the second approach.

#### Http Client configuration

If you agree / the team agrees with me that graph should handle it through an event, we can get rid of the http client setup in the proxy - but the image fetcher needs the same kind of config for insecure mode etc. Not sure we have a useful pattern/example to copy.
I'm a bit torn how we want to handle allowed urls. Default to only the IdP host? allow all urls because the IdP MUST provide us with safe urls (convenient, but :-\ )?

Guibi1:
Thanks for all that, i'm really not familiar enough with the codebase to have thought about it that way.
Being opt-out and using events sounds good! 
Of the urls, we could default to idP host only and have an unsafe option to allow all? I feel like someone could have a setup where their idP acts as a proxy in front of Active Directory for exemple, and where the picture url is just relayed.

dschmidt:
Maybe allow a glob pattern like `https://*.cdn.mydomain.com`? 
That would also allow for `*`for all hosts (although that's only a good idea if the IdP takes care of managing secure urls only)
Default to IdP url without path sounds fine to me :)

services/graph/pkg/config/config.go
```diff
@@ -116,6 +117,11 @@ type Identity struct {
 	LDAP    LDAP   `yaml:"ldap"`
 }
 
+type OIDCProfilePicture struct {
+	OIDCIssuer   string   `yaml:"oidc_issuer" env:"OC_URL;OC_OIDC_ISSUER;GRAPH_OIDC_ISSUER" desc:"URL of the OIDC issuer used to derive the default profile-picture URL allowlist when no explicit allowlist is configured." introductionVersion:"6.3.0"`
+	URLAllowlist []string `yaml:"url_allowlist" env:"GRAPH_OIDC_PROFILE_PICTURE_URL_ALLOWLIST" desc:"A comma separated allowlist of URL patterns accepted for profile-picture sync events. Patterns can be full URLs with glob support in the host (for example 'https://*.example.com') or '*' to allow all URLs. If empty, only the OIDC issuer host is allowed by default." introductionVersion:"6.3.0"`
```
dschmidt: It should be clear that `'*'` is dangerious - and it should only be used if the IdP is trusted to provide only valid urls

services/graph/pkg/config/config.go
```diff
@@ -116,6 +117,11 @@ type Identity struct {
 	LDAP    LDAP   `yaml:"ldap"`
 }
 
+type OIDCProfilePicture struct {
+	OIDCIssuer   string   `yaml:"oidc_issuer" env:"OC_URL;OC_OIDC_ISSUER;GRAPH_OIDC_ISSUER" desc:"URL of the OIDC issuer used to derive the default profile-picture URL allowlist when no explicit allowlist is configured." introductionVersion:"6.3.0"`
```
dschmidt: It feels a bit weird to have an explicit setting just to make it the default for another setting - not sure...

services/graph/pkg/service/v0/service.go
```diff
@@ -39,8 +45,9 @@ import (
 
 const (
 	// HeaderPurge defines the header name for the purge header.
-	HeaderPurge     = "Purge"
-	displayNameAttr = "displayName"
+	HeaderPurge          = "Purge"
+	displayNameAttr      = "displayName"
+	maxProfilePhotoBytes = 10 << 20
```
dschmidt: `   const maxProfilePhotoBytes = 10 << 20 // 10 MiB`
maybe?

services/graph/pkg/service/v0/service.go
```diff
@@ -588,6 +605,125 @@ func (g *Graph) StartListenForLogonEvents(ctx context.Context, l log.Logger) err
...
+		return path.Clean(parsedPattern.Path) == path.Clean(target.Path)
+	}
+
+	hostMatcher, err := glob.Compile(strings.ToLower(pattern))
+	if err != nil {
+		return false
+	}
+	return hostMatcher.Match(strings.ToLower(target.Host))
+}
+
```
dschmidt: My feeling is that `syncProfilePictureFromURL` should be a function on the userProfilePhotoService and the rest of this block also belongs there. Yes, it needs a bit of wiring to pass configs down - but I feel it's worth it to have a clean separation of concerns

services/graph/pkg/service/v0/option.go
```diff
@@ -47,6 +48,13 @@ type Options struct {
 	NatsKeyValue             jetstream.KeyValue
 }
 
+// ProfilePictureHTTPClient provides a function to set the HTTP client used for downloading OIDC profile pictures.
+func ProfilePictureHTTPClient(val HTTPClient) Option {
+	return func(o *Options) {
+		o.ProfilePictureHTTPClient = val
+	}
+}
+
```
dschmidt: Nitpick but I think this belongs further down in the file

services/graph/pkg/service/v0/service.go
```diff
@@ -200,6 +208,10 @@ func NewService(opts ...Option) (Graph, error) { //nolint:maintidx
 		traceProvider:            options.TraceProvider,
 		valueService:             options.ValueService,
 		natskv:                   options.NatsKeyValue,
+		profilePictureHTTPClient: options.ProfilePictureHTTPClient,
```
dschmidt: Why not keep this as a member in the photo service?

services/proxy/pkg/config/defaults/defaultconfig.go
```diff
@@ -90,6 +90,7 @@ func DefaultConfig() *config.Config {
 			Username:    "preferred_username",
 			Email:       "email",
 			DisplayName: "name",
+			ProfilePicture: "",
```
dschmidt: If I understand the spec correctly, the default is picture (some IdPs seem to use avatar additionally?)

Let's use "picture" here - if someone doesnt want to use it although their IdP provides it, then they can still opt out by setting the value to empty explicitly?

vendor/github.com/opencloud-eu/reva/v2/pkg/events/users.go
```diff
@@ -122,8 +122,9 @@ func (BackchannelLogout) Unmarshal(v []byte) (interface{}, error) {
...
 // UserSignedIn is emitted when a user signs in
 type UserSignedIn struct {
-	Executant *user.UserId
-	Timestamp *types.Timestamp
+	Executant  *user.UserId
+	PictureURL string `json:",omitempty"`
+	Timestamp  *types.Timestamp
 }
 
 // Unmarshal to fulfill umarshaller interface
```
dschmidt, line 129: Unfortunately we can't do this - it would be lost on the next reva update. 

We need to upstream it to the reva repo first. Maybe there's also the right place to have a discussion whether we really want to reuse this event or introduce a new one (or new ones: for picture updated e.g.)

services/proxy/pkg/middleware/account_resolver.go
```diff
@@ -126,6 +128,18 @@ func readStringClaim(path string, claims map[string]any) (string, error) {
 	return value, fmt.Errorf("claim path '%s' not set or empty", path)
 }
 
+func (m accountResolver) readProfilePictureURL(claims map[string]any) string {
+	if m.autoProvisionClaims.ProfilePicture == "" {
+		return ""
+	}
+	pictureURL, err := readStringClaim(m.autoProvisionClaims.ProfilePicture, claims)
+	if err != nil {
+		m.logger.Debug().Err(err).Str("claim", m.autoProvisionClaims.ProfilePicture).Msg("profile picture claim missing")
```
dschmidt: Maybe the wording should be a more neutral?

"missing" sounds like a misconfiguration, maybe "no profile picture claim found" would be more neutral.
And maaaybe we should add some user information to the log? like this it's really hard to debug or learn anything useful from the log message

services/proxy/pkg/command/server.go
```diff
@@ -367,6 +367,7 @@ func loadMiddlewares(logger log.Logger, cfg *config.Config,
 			middleware.UserOIDCClaim(cfg.UserOIDCClaim),
 			middleware.UserCS3Claim(cfg.UserCS3Claim),
 			middleware.TenantOIDCClaim(cfg.TenantOIDCClaim),
+			middleware.AutoProvisionClaims(cfg.AutoProvisionClaims),
```
dschmidt, line 378: Why do you have to add it here?

It was already added in another place (conditionally)

dschmidt (changes requested):
Did you forget to commit the actual photo service? I don't see it.
Also I don't see how you instantiate it and if you always just use the default http client. 

We probably need to setup a configurable client like in the proxy for talking to the IdP:

```golang
	oidcHTTPClient := &http.Client{
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{
				MinVersion:         tls.VersionTLS12,
				InsecureSkipVerify: cfg.OIDC.Insecure, //nolint:gosec
			},
			DisableKeepAlives: true,
		},
		Timeout: time.Second * 10,
	}
```

Anyhow I really like where this is going! Good job

services/graph/pkg/config/config.go
dschmidt: not only valid: but *trusted* urls - otherwise it's an SSRF attack vector

services/proxy/pkg/command/server.go
Guibi1: The middleware needs access to the picture claim to emit the sync event.

pkg/events/events.go
```diff
@@ -20,3 +21,27 @@ func (ResourceMention) Unmarshal(v []byte) (interface{}, error) {
...
+type UserProfilePictureUpdated struct {
+	Executant *user.UserId
+	Timestamp *types.Timestamp
+}
+
+func (UserProfilePictureUpdated) Unmarshal(v []byte) (interface{}, error) {
+	e := UserProfilePictureUpdated{}
+	err := json.Unmarshal(v, &e)
+	return e, err
+}
```
dschmidt, line 47: minor nitpick: one event has User prefix, one has not - feels a bit inconsistent

changelog/unreleased/oidc-profile-picture-sync.md
dschmidt, line 1: Looking at the changelog folder, it seems we're not doing changelog items there anymore - but I'm not sure about that

dschmidt (commented):
Great, I'm pretty happy with this now :)

unfortunately it's ultimately not up to me to decide

services/graph/pkg/config/config.go
```diff
@@ -25,13 +25,14 @@ type Config struct {
...
-	Events            Events       `yaml:"events"`
-	UnifiedRoles      UnifiedRoles `yaml:"unified_roles"`
-	MaxConcurrency    int          `yaml:"max_concurrency" env:"OC_MAX_CONCURRENCY;GRAPH_MAX_CONCURRENCY" desc:"The maximum number of concurrent requests the service will handle." introductionVersion:"1.0.0"`
+	Application                Application  `yaml:"application"`
+	Spaces                     Spaces       `yaml:"spaces"`
+	Identity                   Identity     `yaml:"identity"`
+	ProfilePictureURLAllowlist []string     `yaml:"profile_picture_url_allowlist" env:"GRAPH_PROFILE_PICTURE_URL_ALLOWLIST" desc:"A comma separated allowlist of URL patterns accepted for profile-picture sync events. Patterns can be full URLs with glob support in the host (for example 'https://*.example.com') or '*' to allow all URLs (dangerous: only use if the IdP is fully trusted to provide safe URLs, otherwise this is an SSRF attack vector). If empty, the OpenCloud URL host is allowed by default." introductionVersion:"6.3.0"`
```
rhafer, line 31: introductionVersion needs to be `%%NEXT%%` for all newly introduced config settings. It will be replaced with the correct version during the release process.

services/graph/pkg/server/http/server.go
```diff
@@ -155,7 +157,36 @@ func Server(opts ...Option) (http.Service, error) {
...
 
-		userProfilePhotoService, err = svc.NewUsersUserProfilePhotoService(photoStorage)
+		photoOpts := []svc.UserProfilePhotoOption{}
+		// The profile picture sync is always enabled; the proxy decides
+		// whether to emit sync events based on PROXY_AUTOPROVISION_CLAIM_PICTURE.
+		// The default allowlist uses the OpenCloud URL host.
+		oidcIssuer := ""
+		if options.Config.Commons != nil {
+			oidcIssuer = options.Config.Commons.OpenCloudURL
+		}
```
rhafer, line 167: The variable name `oidcIssuer` is misleading here when you're defaulting the URL of the OpenCloud instance. `oidcIssuer` usually refers to the IDP.

services/graph/pkg/server/http/server.go
```diff
@@ -155,7 +157,36 @@ func Server(opts ...Option) (http.Service, error) {
...
+		// Respects OC_INSECURE for TLS certificate validation and
+		// honours HTTP_PROXY/HTTPS_PROXY environment variables.
+		profilePictureHTTPClient := &stdhttp.Client{
+			Timeout: 10 * time.Second,
+			Transport: &stdhttp.Transport{
+				Proxy: stdhttp.ProxyFromEnvironment,
+				//nolint:gosec // We need the ability to run with "insecure" (dev/testing)
+				TLSClientConfig: &tls.Config{
+					MinVersion:         tls.VersionTLS12,
+					InsecureSkipVerify: options.Config.Events.TLSInsecure,
```
rhafer, line 178: I am pretty sure you're using the wrong source  for the Insecure flag here. `options.Config.Events.TLSInsecure` is only meant for the event listener and nothing else. You might need to introduce a separate insecure flag for the picture download client.

pkg/events/events.go
rhafer: Agreed.

services/graph/pkg/service/v0/api_users_user_profile_photo.go
```diff
@@ -85,6 +116,128 @@ func (s UsersUserProfilePhotoService) UpdatePhoto(ctx context.Context, id string
...
+			return true
+		}
+		if strings.HasSuffix(parsedPattern.Path, "*") {
+			prefix := strings.TrimSuffix(parsedPattern.Path, "*")
+			return strings.HasPrefix(target.Path, prefix)
+		}
+		return path.Clean(parsedPattern.Path) == path.Clean(target.Path)
+	}
+
+	hostMatcher, err := glob.Compile(strings.ToLower(pattern))
```
rhafer, line 234: What is this good for? We only get here if url.Parse returns an error, or if the host part is empty.

I guess this is for the `pattern == *` case? If yes, I'd suggest to make that explicit. And do an `if pattern == *` with an early return right before trying to parse the pattern as an URL.

services/graph/pkg/service/v0/api_users_user_profile_photo.go
```diff
@@ -85,6 +116,128 @@ func (s UsersUserProfilePhotoService) UpdatePhoto(ctx context.Context, id string
...
+			return false
+		}
+		if parsedPattern.Path == "" || parsedPattern.Path == "/" {
+			return true
+		}
+		if strings.HasSuffix(parsedPattern.Path, "*") {
+			prefix := strings.TrimSuffix(parsedPattern.Path, "*")
+			return strings.HasPrefix(target.Path, prefix)
+		}
+		return path.Clean(parsedPattern.Path) == path.Clean(target.Path)
```
rhafer, line 231: Reading the code and documentation in config.go it is a bit unclear to me what kind of patterns you do want to support. E.g. the code seems to accept patterns with out a Scheme. The annotation in config.go only refers to globs in hostname part. The code also accepts a trailiing * on the path ...

services/graph/pkg/service/v0/service.go
```diff
@@ -585,6 +587,17 @@ func (g *Graph) StartListenForLogonEvents(ctx context.Context, l log.Logger) err
 					if err := g.identityBackend.UpdateLastSignInDate(ctx, ev.Executant.OpaqueId, utils.TSToTime(ev.Timestamp)); err != nil {
 						l.Error().Err(err).Str("userid", ev.Executant.OpaqueId).Msg("Error updating last sign in date")
 					}
+				case ocevents.ProfilePictureSyncRequested:
+					if err := g.userProfilePhotoService.SyncPhotoFromURL(ctx, ev.Executant.GetOpaqueId(), ev.PictureURL); err != nil {
```
rhafer, line 591: Somehow I think we should trigger the photo download asynchronously in a go routine and not block the event loop here.
If for some reason the download + upload the metadata storage takes longer this can have ill effects on the event consumer.

I know, that the `UploadLastSignInDate()` doesn't do it correct either, but that shouldn't stop us from making it better here.

services/proxy/pkg/config/config.go
```diff
@@ -168,6 +168,7 @@ type AutoProvisionClaims struct {
 	Username    string `yaml:"username" env:"PROXY_AUTOPROVISION_CLAIM_USERNAME" desc:"The name of the OIDC claim that holds the username." introductionVersion:"1.0.0"`
 	Email       string `yaml:"email" env:"PROXY_AUTOPROVISION_CLAIM_EMAIL" desc:"The name of the OIDC claim that holds the email." introductionVersion:"1.0.0"`
 	DisplayName string `yaml:"display_name" env:"PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME" desc:"The name of the OIDC claim that holds the display name." introductionVersion:"1.0.0"`
+	Picture     string `yaml:"picture" env:"PROXY_AUTOPROVISION_CLAIM_PICTURE" desc:"The name of the OIDC claim that holds the profile picture URL. Defaults to 'picture' (the standard OIDC claim). Set to an empty string to disable profile picture sync." introductionVersion:"6.3.0"`
```
rhafer, line 171: same as above "introductionVersion" should be `%%NEXT%%` 

Also I am a bit unhappy with the `PROXY_AUTOPROVISION_CLAIM_PICTURE` name. As your feature is somewhat orthogonal with the rest of the `PROXY_AUTOPROVISION_*` settings which all are focussed around the creation and udpate of the user accounts and are only relevant when `PROXY_AUTOPROVISION_ACCOUNTS` is `true`. The new value is different. Should it even live in the `AutoProvisionClaims` struct?

rhafer (changes requested):
@Guibi1 Thanks a lot for you effort on this pull request.

I haven't looked through all of the code yet. Especially I still need to get through the changes in the proxy. But in general I like it. I've added some comments and questions on the parts that I already reviewed.

changelog/unreleased/oidc-profile-picture-sync.md
kulmann: that's correct. Changelog is being built from PR titles (linking to the PRs) by the `ready-release-go`. So having a meaningful PR title and description is all you need. This changelog item can be deleted.

PR #2724, opened 2026-05-06
allow configuring initial scan of the posix fs

butonic:
In kubernetes deployments additional pods may be started under load. We do not want to scan the fs at that time. We added a new `STORAGE_USERS_POSIX_SCAN_FS="true"` env var to toggle that. The default remains enabled.


fschade (approved):

butonic:
closing in favor of https://github.com/opencloud-eu/opencloud/pull/2726

PR #2744, opened 2026-05-12
feat(graph): populate driveItem.webUrl per Libre Graph spec

dschmidt:
The Libre Graph spec declares `webUrl` on `driveItem` but `cs3ResourceToDriveItem` never set it, only `drive.webUrl` was populated.

Populate it by parsing the configured public base URL once at service construction and threading it through the shared driveItem builder. Every endpoint that returns drive items now carries the field.


micbar (approved):
Thanks!

IMHO this is a no brainer.

PR #2770, opened 2026-05-15
refactor(graph): use BaseGraphService.publicBaseURL consistently

dschmidt:
## Summary

Follow-up to #2744. `drive.WebUrl`, `driveItem.WebUrl` and the public share
link all derive from `graph.spaces.webdav_base`, but only `driveItem.WebUrl`
used the pre-parsed `BaseGraphService.publicBaseURL`; the other two re-parsed
the config per call. This routes all three through it.

Pure refactor, no behavior change.

## Test plan

- [x] build + `go test ./services/graph/pkg/service/v0/...` (incl. `TestCS3ResourceToDriveItemPopulatesWebUrl`)
- [ ] CI green


dschmidt:
Superseded by #3248, which integrates this refactor as its first commit (the signed download URL builds on the same pre-parsed `publicBaseURL`). Closing here to keep the change in a single stack.

Issue #1363, opened 2025-08-17
Error messages logged at wrong level


kaivol:
## Describe the bug
While configuring OpenCloud to work with an external LDAP server, I encountered errors like this:
````json
{
"level":"debug",
"service":"graph",
"request-id":"***",
"error":"itemNotFound: not found",
"time":"***",
"line":"github.com/opencloud-eu/opencloud/services/graph/pkg/service/v0/groups.go:253",
"message":"could not get group: backend error"
}
````
(Source [groups.go](https://github.com/opencloud-eu/opencloud/blob/main/services/graph/pkg/service/v0/groups.go#L253))

or 
````json
{
"level":"debug",
"service":"graph",
"request-id":"***",
"error":"itemNotFound: not found",
"entry":"***",
"time":"***",
"line":"github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap.go:1117",
"message":"error reading attribute member entry"
}
````
(Source [ldap.go](https://github.com/opencloud-eu/opencloud/blob/main/services/graph/pkg/identity/ldap.go#L1117))

If I'm not missing anything, these should be logged at the appropriate level `error`, not at `debug`.

PR #2780, opened 2026-05-18
change logger level

dragonchaser:
refs #1363

### DIFF
--- a/services/graph/pkg/identity/ldap.go
+++ b/services/graph/pkg/identity/ldap.go
@@ -1137,7 +1137,7 @@ func (i *LDAP) expandLDAPAttributeEntries(ctx context.Context, e *ldap.Entry, at
 		ue, err := i.getUserByDN(entryDN, searchTerm)
 		if err != nil {
 			// Ignore errors when reading a specific entry fails, just log them and continue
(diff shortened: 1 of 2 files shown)


services/graph/pkg/identity/ldap.go
```diff
@@ -1137,7 +1137,7 @@ func (i *LDAP) expandLDAPAttributeEntries(ctx context.Context, e *ldap.Entry, at
 		ue, err := i.getUserByDN(entryDN, searchTerm)
 		if err != nil {
 			// Ignore errors when reading a specific entry fails, just log them and continue
-			logger.Debug().Err(err).Str("entry", entryDN).Msg("error reading attribute member entry")
+			logger.Error().Err(err).Str("entry", entryDN).Msg("error reading attribute member entry")
```
rhafer: Nit: `Error` might be a bit too much here. As technically this is not a real error. LDAP doesn't provide any referential integrity so there can be situations where the user object does not exist any more, even if it is still referenced as a `member` in a group. `Warn` or `Info` might be more appropriate.

rhafer (approved):
Small nitpick. Feel free to merge anyway if you disagree.

rhafer (approved):

PR #2789, opened 2026-05-19
Add a "posixfs scan" command for scanning the posix tree for new files

aduffeck:
/cc @butonic


dschmidt:
Maybe worth cross linking documentation between `search index --force-rescan` and this? 
Explaining which is good for what and why and when I might need one or the other or both?

aduffeck:
> Maybe worth cross linking documentation between `search index --force-rescan` and this? Explaining which is good for what and why and when I might need one or the other or both?

That's a good point. While the `search index --force-rescan` is documented in the service README.me and thus ends up in the documentation the maintenance commands are currently missing completely as far as I know. @Svanvith Is that something we could add?

dragonchaser (approved):

PR #2810, opened 2026-05-21
chore(idp): bump to pnpm-11 and migrate config

rhafer:


JammingBen (approved):
Change looks good! Since the `minimumReleaseAge` now defaults to 24 hours, we should maybe set the same in the renovate config, at least for the JS deps.

rhafer:
> Since the `minimumReleaseAge` now defaults to 24 hours, we should maybe set the same in the renovate config, at least for the JS deps.

Thanks for the hint. We're still using dependabot. I adjusted the config accordingly.

Issue #2824, opened 2026-05-22
Public cannot download file using /dav/public-files/


v-scharf:
related: https://github.com/opencloud-eu/web/pull/2577

## Steps to reproduce
1. create a public link
2. public tries to download file (without remote.php)
`curl 'https://cloud.opencloud.test/dav/public-files/tEeBcqGhSHOTMvo/1.MP4' -u public:"password" -vk`

## Expected behavior
can download

## Actual behavior
401

## Setup
Please describe how you started the server and provide a list of relevant environment variables or configuration files.

<details>
<p>

```console
OC_XXX=somevalue
OC_YYY=somevalue
PROXY_XXX=somevalue
```

</p>
</details>

PR #2826, opened 2026-05-22
fix-2824. run tests without remote.php

v-scharf:
- [x] fixing #2824 
- [x] run tests without remote.php by default


services/proxy/pkg/config/defaults/defaultconfig.go
```diff
@@ -233,6 +233,10 @@ func DefaultPolicies() []config.Policy {
 					Endpoint: "/webdav/",
 					Service:  "eu.opencloud.web.frontend",
 				},
+				{
+					Endpoint: "/webdav",
```
aduffeck: Why do we need this change? `/webdav/` is already wired to the frontend service so wiring `/webdav` elsewhere sounds strange to me.

aduffeck (approved):

PR #2880, opened 2026-06-02
use ldap instead of ldaps internally

dragonchaser:
This PR removes the internal ssl of the idm, since this is never used externally but only on loopback.

### DIFF
--- a/services/idm/pkg/command/server.go
+++ b/services/idm/pkg/command/server.go
@@ -59,6 +59,8 @@ func Server(cfg *config.Config) *cobra.Command {
 
 				if cfg.IDM.LDAPSAddr != "" {
 					servercfg.LDAPSListenAddr = cfg.IDM.LDAPSAddr
+					servercfg.TLSCertFile = cfg.IDM.Cert
+					servercfg.TLSKeyFile = cfg.IDM.Key
 				}
 
 				if err := os.MkdirAll(path.Join(defaults.BaseDataPath(), "idm"), 0700); err != nil {
--- a/services/idm/pkg/command/server.go
+++ b/services/idm/pkg/command/server.go
@@ -8,10 +8,11 @@ import (
 	"html/template"
 	"os"
 	"os/signal"
+	"path"
 	"strings"
 
 	"github.com/opencloud-eu/opencloud/pkg/config/configlog"
-	pkgcrypto "github.com/opencloud-eu/opencloud/pkg/crypto"
+	"github.com/opencloud-eu/opencloud/pkg/config/defaults"
 	"github.com/opencloud-eu/opencloud/pkg/log"
 	"github.com/opencloud-eu/opencloud/pkg/runner"
 	"github.com/opencloud-eu/opencloud/services/idm"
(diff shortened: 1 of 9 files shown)


services/idm/pkg/command/server.go
```diff
@@ -47,23 +48,19 @@ func Server(cfg *config.Config) *cobra.Command {
...
-					Logger:          log.LogrusWrap(logger.Logger),
-					LDAPHandler:     "boltdb",
-					LDAPSListenAddr: cfg.IDM.LDAPSAddr,
-					TLSCertFile:     cfg.IDM.Cert,
-					TLSKeyFile:      cfg.IDM.Key,
-					LDAPBaseDN:      "o=libregraph-idm",
-					LDAPAdminDN:     "uid=libregraph,ou=sysusers,o=libregraph-idm",
+					Logger:         log.LogrusWrap(logger.Logger),
+					LDAPHandler:    "boltdb",
+					LDAPListenAddr: cfg.IDM.LDAPAddr,
```
rhafer, line 53: Hm, if we really want to only allow `ldap://` only for the embedded LDAP server. I think we need to ensure that `cfg.IDM.LDAPAddr` is set to a loopback address. (using nets.ResolveTCPAddr and net.IP.IsLoopback()). Or just make it not configurable.

Also I believe we already have users that expose the IDM on the network (e.g. for integrating with Authelia or other 3rd party IDPs). So we should probably not complete prevent exposing and `ldaps://` listener on the network.

Howabout this:

1. Always expose an unencrypted `ldap://` on the loopback interface on port 9236
2. Configure our internal client to connect on `ldap://127.0.0.1:9236`
3. Default `cfg.IDM.LDAPSAddr` to `""`. But allow it to be set for providing an SSL listener.
4. To workaround issues with expiring certificates:
    when the user did not set `cfg.IDM.Cert` and `cfg.IDM.Key`. generate a new self-signed certificate at every startup. (AFAIK this is also what the proxy does).

rhafer (changes requested):

services/idm/pkg/config/defaults/defaultconfig.go
```diff
@@ -51,5 +50,10 @@ func EnsureDefaults(cfg *config.Config) {
 
 // Sanitize sanitizes the configuration
 func Sanitize(cfg *config.Config) {
-	// nothing to sanitize here
+	if cfg.IDM.LDAPSAddr == "" &&
+		cfg.IDM.LDAPAddr != "" &&
+		(!strings.Contains(cfg.IDM.LDAPAddr, "127.0.0.1") &&
+			!strings.Contains(cfg.IDM.LDAPAddr, "localhost")) {
+		panic("Invalid configuration: 'ldap_addr' is set but 'ldaps_addr' is not set. For security reasons, the 'ldap_addr' setting is only allowed to be used with loopback addresses. Please set 'ldaps_addr' to a valid address and port to listen for LDAPS connections.")
+	}
```
rhafer: we we do not allow to configure `cfg.IDM.LDAPAddr` this check is not need, because it will always be `127.0.0.1`

services/idm/pkg/config/defaults/defaultconfig.go
```diff
@@ -30,9 +31,7 @@ func DefaultConfig() *config.Config {
 		CreateDemoUsers:    false,
 		DemoUsersIssuerUrl: "https://localhost:9200",
 		IDM: config.Settings{
-			LDAPSAddr:    "127.0.0.1:9235",
-			Cert:         path.Join(defaults.BaseDataPath(), "idm", "ldap.crt"),
-			Key:          path.Join(defaults.BaseDataPath(), "idm", "ldap.key"),
+			LDAPAddr:     "127.0.0.1:9235",
```
rhafer: We need to use a different port for LDAPAddr  otherwise the idm service fails to start when LDAPSAddr is set. Let use `9236`

services/idm/pkg/command/server.go
```diff
@@ -47,23 +48,23 @@ func Server(cfg *config.Config) *cobra.Command {
...
 
 					BoltDBFile: cfg.IDM.DatabasePath,
 				}
 
 				if cfg.IDM.LDAPSAddr != "" {
-					// Generate a self-signing cert if no certificate is present
-					if err := pkgcrypto.GenCert(cfg.IDM.Cert, cfg.IDM.Key, logger); err != nil {
-						logger.Fatal().Err(err).Msgf("Could not generate test-certificate")
-					}
+					servercfg.LDAPSListenAddr = cfg.IDM.LDAPSAddr
```
rhafer, line 61: When `cfg.IDM.LDAPSAddr` is set. We also need configure certificates (i.e. point `servercfg.TLSCertFile` and `servercfg.TLSKeyFile` ).

If `cfg.IDM.Cert` and `cfg.IDM.Key` are unset we could just generate new certificates at on every startup, at a fixed location. User could supply their own propper certificates by setting `cfg.IDM.Cert` and `cfg.IDM.Key`

rhafer (changes requested):

rhafer (approved):

rhafer:
@micbar how can we assure that we add something about the changed behavior to the release notes?

PR #2895, opened 2026-06-04
feat(proxy): add PROXY_USERS_DEFAULT_QUOTA for personal spaces

flash7777:
## Summary

- Adds a new `PROXY_USERS_DEFAULT_QUOTA` environment variable (config: `default_users_quota`) that sets a default quota in bytes for newly created personal spaces
- This quota acts as a fallback when no role-specific quota is configured via `role_quotas`
- A value of `0` (default) preserves the existing behavior (unlimited quota)

## Motivation

Currently, personal spaces are created without any quota unless a role-specific quota mapping is configured. This means that by default, every user gets unlimited storage on their personal space. On systems with limited storage (e.g. a 100GB volume), a single user can fill the entire disk -- which is what happened on our production system, causing a full outage.

The existing `GRAPH_SPACES_DEFAULT_QUOTA` only applies to project spaces created via the Graph API. There was no equivalent for personal spaces created during user login.

## Usage

```
PROXY_USERS_DEFAULT_QUOTA=1000000000  # 1 GB default for new personal spaces
```

Or in `proxy.yaml`:
```yaml
default_users_quota: 1000000000
```

## Changes

| File | Change |
|------|--------|
| `services/proxy/pkg/config/config.go` | New config field `DefaultUsersQuota` with env `PROXY_USERS_DEFAULT_QUOTA` |
| `services/proxy/pkg/middleware/options.go` | New middleware option + setter function |
| `services/proxy/pkg/middleware/create_home.go` | Fallback to `defaultUsersQuota` when no role quota matches |
| `services/proxy/pkg/command/server.go` | Wire config to middleware |

## Test plan

- [ ] Set `PROXY_USERS_DEFAULT_QUOTA=1000000000` and create a new user -- personal space should have 1GB quota
- [ ] Without the env var set (or set to 0) -- existing behavior, unlimited quota
- [ ] With both role quota and default quota configured -- role quota takes precedence
- [ ] Existing personal spaces are not affected (quota is only applied on creation)

With support of [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/services/proxy/pkg/command/server.go
+++ b/services/proxy/pkg/command/server.go
@@ -394,6 +394,7 @@ func loadMiddlewares(logger log.Logger, cfg *config.Config,
 			middleware.TraceProvider(traceProvider),
 			middleware.WithRevaGatewaySelector(gatewaySelector),
 			middleware.RoleQuotas(cfg.RoleQuotas),
+			middleware.DefaultUsersQuota(cfg.DefaultUsersQuota),
 		),
 	)
 }
--- a/services/proxy/pkg/config/config.go
+++ b/services/proxy/pkg/config/config.go
@@ -24,6 +24,7 @@ type Config struct {
 	GrpcClient    client.Client         `yaml:"-"`
 
 	RoleQuotas                    map[string]uint64   `yaml:"role_quotas"`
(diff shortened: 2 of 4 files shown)


rhafer (changes requested):
As you noted, there is already a way of configuring default quotas for users (based on the assigned role). I don't see a good reason for introducing yet another config setting (we already have far too many) as a fallback. What prevents you from using the existing mechanism?

rhafer:
I am closing this one. Feel free to re-open, when you've feedback for my review comments.

PR #2915, opened 2026-06-09
[DRAFT] Use inotifywait module from upstream

dragotin:
## Description
This PR drops the private inotifywait fork from the build. We built it privately to have a logger passed into the module. This patch now went upstream.

## Related PR

Same PR in Reva: https://github.com/opencloud-eu/reva/pull/668

## Motivation and Context
Reduce private forks of modules


rhafer:
I've added this to #2916

PR #2918, opened 2026-06-09
chore: reduce build context for docker

rhafer:
Always including the parent directory into the build context was a bad idea. This reverts the change and only add the opencloud tree to the build context by default.
To build a docker image with workspaces enabled a new make target "dev-docker-with-workspace" is introduced that adjusts the docker build call accordingly (by setting the right build context and the new SRCDIR arg to point to the correct directory for the opencloud tree).


rhafer:
alternative for #2913

AlexAndBear (approved):

micbar (approved):

PR #2935, opened 2026-06-11
fix(graph): translate sharing roles consistently

### LINKED ISSUE #2800: Sharing Roles will not be translated
### Initial Checklist

- [x] I understand this is a bug report and questions should be posted in the [Community Discussions](https://github.com/orgs/opencloud-eu/discussions/)
- [x] I searched [issues](https://github.com/opencloud-eu/web/issues?q=is%3Aissue) and couldn’t find anything (or linked relevant results below)

### Bug Description

The sharing roles "can view" "can edit" etc. will not be translated and stay on German even when you change the language:

Image

The reason is, the `DEFAULT_LANGUAGE=` variable is not added in the .env but in the docker-compose.yaml

### Reproduction Steps

latest opencloud-compose
Opencloud 6.2.0 rolling

Use an older .env or remove 'DEFAULT_LANGUAGE=' from the .env 
Make sure ` OC_DEFAULT_LANGUAGE: ${DEFAULT_LANGUAGE}` is set in the docker-compse.yaml

Rebuild the docker container
You will get a warning:
`WARN[0000] The "DEFAULT_LANGUAGE" variable is not set. Defaulting to a blank string.`

set the language to "German" and check the sharing roles.
Switch back to "English" and check them again
You can choose other languages too, they will stay German

### Expected Outcome

The sharing roles will will be translated to English again

### Actual Outcome

The sharing roles stay German

rhafer:
GetRoleDefinition/s does now handle l10n correctly. Previsouly it just returned the non-localized string. What made things worse was that ListPermissions() mutated global list of available roles and replaced some strings with translated values depending on the `accept-language` header. Which resulted in GetRoleDefinition returning results in mixed localization depending on who/what called ListPermissions before.

Fixes: #2800

### DIFF
--- a/services/graph/pkg/service/v0/api_driveitem_permissions.go
+++ b/services/graph/pkg/service/v0/api_driveitem_permissions.go
@@ -22,7 +22,6 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 	libregraph "github.com/opencloud-eu/libre-graph-api-go"
-
 	revactx "github.com/opencloud-eu/reva/v2/pkg/ctx"
 	"github.com/opencloud-eu/reva/v2/pkg/publicshare"
 	"github.com/opencloud-eu/reva/v2/pkg/rgrpc/todo/pool"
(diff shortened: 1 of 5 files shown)


micbar (commented):
Nice!

micbar (approved):
Nice!

PR #2960, opened 2026-06-16
feat(graph): add metadata API for custom xattrs (GET + PUT)

flash7777:
## Summary

- **GET** `/graph/v1beta1/drives/{driveID}/items/{itemID}/metadata` — returns all custom metadata (`user.oc.md.*`) as JSON
- **PUT** `/graph/v1beta1/drives/{driveID}/items/{itemID}/metadata` — sets metadata from JSON body, keys map to `user.oc.md.{key}` xattrs

This enables client-side features like notes, classifications, or any custom properties without requiring WebDAV PROPPATCH (which has namespace/key mapping issues for `oc:` properties).

## Test plan

- [ ] Upload a file, PUT `{"note": "test"}` to metadata endpoint, verify xattr `user.oc.md.note` is set
- [ ] GET metadata endpoint, verify `{"note": "test"}` is returned
- [ ] Test with special characters (umlauts, CJK, emoji)
- [ ] Verify permission check (only users with upload permission can set metadata)

### DIFF
--- a/services/graph/pkg/service/v0/api_drives_drive_item_metadata.go
+++ b/services/graph/pkg/service/v0/api_drives_drive_item_metadata.go
@@ -1,0 +1,131 @@
+package svc
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+
+	rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/opencloud-eu/opencloud/services/graph/pkg/errorcode"
+)
+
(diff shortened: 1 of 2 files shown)


dschmidt:
This should be spec'ed in https://github.com/opencloud-eu/libre-graph-api first before the endpoints are added here.

See my PR https://github.com/opencloud-eu/libre-graph-api/pull/35 for an attempt to achieve something similar (I believe :))
Especially see @butonic 's response for open extensions vs. fieldsets. I'm not sure which way we want to go.
Also by now I'm thinking we need some typing information, so it can be indexed better (e.g. geo data, date information etc. should possibly not be stored just as strings e.g.) so maybe we don't want to do it completely schemaless? But then how do we want to handle schema changes? It's a can of worms honestly.

Additional feedback on my PR would be greatly appreciated :)

PR #2963, opened 2026-06-17
Bump alpine base image

rhafer:
Also adjust the build image to use the opencloudeu/golang image. So we're in sync with how the binaries are built in CI. It already contains all the required dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,15 +21,14 @@ COPY ./ /opencloud/
 WORKDIR /opencloud/opencloud
 RUN make node-generate-prod
 
-FROM golang:1.24-alpine AS build
-RUN apk add bash make git curl gcc musl-dev libc-dev binutils-gold inotify-tools vips-dev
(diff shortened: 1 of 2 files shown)


dragonchaser:
fails with
 
```log
0.250 	-o '/dist/binaries/opencloud-linux-amd64' \
0.250 	./cmd/opencloud
0.253 go: ../go.work requires go >= 1.26.0 (running go 1.25.11; GOTOOLCHAIN=local)
0.254 make: *** [../.make/release.mk:23: release-linux-docker-amd64] Error 1
0.254 make: Leaving directory '/build/opencloud'

rhafer:
> 0.253 go: ../go.work requires go >= 1.26.0 (running go 1.25.11; GOTOOLCHAIN=local)

That's caused by your local `go.work` file. We can't do much about that.

dragonchaser:
ah nvm.... `go.work` caused the issue....

dragonchaser (approved):

PR #686, opened 2025-04-16
fix(decomposeds3): enable async-uploads by default

fschade:
## Description
decomposedfss3 has not turned on `OC_ASYNC_UPLOADS` by default which leads to unwanted side-effects when enabling virus-scanning or fulltext-search indexing (tika)

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/573
- Fixes https://github.com/opencloud-eu/opencloud/issues/656

## Motivation and Context
OC_ASYNC_UPLOADS is enabled by default for posixfs and decomposedfs but currently not for decomposedfss3, to align with the other implementations it should be turned on by default.

## How Has This Been Tested?
- local reproduction

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

PR #2967, opened 2026-06-17
[stable-7.2] chore: bump reva to 2.46.5

rhafer:
### 🐛 Bug Fixes

- [stable-2.46] fix: status codes on DeleteStorageSpace [[#692](https://github.com/opencloud-eu/reva/pull/692)]

### 🐛 Bug Fixes

- [stable-2.46] Backport #681 [[#686](https://github.com/opencloud-eu/reva/pull/686)]
- Stable backports [[#679](https://github.com/opencloud-eu/reva/pull/679)]
- https://github.com/opencloud-eu/reva/pull/672
- https://github.com/opencloud-eu/reva/pull/675
- https://github.com/opencloud-eu/reva/pull/677

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,7 +64,7 @@ require (
 	github.com/open-policy-agent/opa v1.15.2
 	github.com/opencloud-eu/icap-client v0.0.0-20250930132611-28a2afe62d89
 	github.com/opencloud-eu/libre-graph-api-go v1.0.8-0.20260310090739-853d972b282d
-	github.com/opencloud-eu/reva/v2 v2.46.3-0.20260610093751-a33d8108dd91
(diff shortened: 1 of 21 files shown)


micbar (approved):

rhafer:
The tests are failing because the server is returning a different error than expected when a user with the `Can manage` role tries to delete a disabled space.

Currently the implementation returns a `404 (Not Found)` the tests expect a `400 (Bad Request)`. Question is, what is the right this to do here.

IMO neither 404 nor 400 are really correct here. I'd think the server should return a 403 as the user does  have access to the space (i.e. can re-enable it) but is just not allowed to purge it.

I'm working on a fix. (Likely requires yet another reva bump).

v-scharf (approved):

PR #2974, opened 2026-06-18
Handle events asynchronously

aduffeck:


fschade:
fixes: `nats: slow consumer, messages dropped on connection [63] for subscription on "main-queue"`, e.g. https://github.com/opencloud-eu/opencloud/issues/2801#issuecomment-4508905014

fschade (approved):

PR #2975, opened 2026-06-18
fix status codes for Space Disable/Delete, also bump reva to latest main

rhafer:
fix: Status codes for Space Disable/Delete

Allow a "permission denied error" from reva to bubble up to the client.
Reva was fixed to return "permission denied" only when the space to be
delete can actually be listed by the user. Other wise it will return
"not found". See reva commit 1bf72cb76394671f373e87f15f23f978cf41ab08.

So when a user with the 'can manage' role tries to purge an already
disabled space it will now get "Forbidden" status instead of a "Not
found".

Also fixes the expected status codes in the tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,7 +64,7 @@ require (
 	github.com/open-policy-agent/opa v1.17.1
 	github.com/opencloud-eu/icap-client v0.0.0-20250930132611-28a2afe62d89
 	github.com/opencloud-eu/libre-graph-api-go v1.0.8-0.20260310090739-853d972b282d
-	github.com/opencloud-eu/reva/v2 v2.46.4-0.20260615073558-209c2cd3b52b
+	github.com/opencloud-eu/reva/v2 v2.46.4-0.20260618143111-1bf72cb76394
 	github.com/opensearch-project/opensearch-go/v4 v4.6.0
 	github.com/orcaman/concurrent-map v1.0.0
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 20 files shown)


v-scharf (approved):

rhafer:
```
Scenario Outline: space admin user tries to disable the personal space       # /woodpecker/src/github.com/opencloud-eu/opencloud/tests/acceptance/features/apiSpaces/spaceManagement.feature:153
    When user "<user>" disables a space "Alice Hansen" owned by user "Alice" # SpacesContext::userDisablesSpace()
    Then the HTTP status code should be "404"                                # FeatureContext::thenTheHTTPStatusCodeShouldBe()

    Examples:
      | user  |
      | Brian |
        Failed step: Then the HTTP status code should be "404"
        HTTP status code 403 is not the expected value 404
        Failed asserting that 403 matches expected '404'.
      | Carol |
```

This needs more investigation. This seems to imply that SpaceManagers can list Personal Spaces, which they shouldn't be able to. Might be another reva issue.

v-scharf:
> This needs more investigation. This seems to imply that SpaceManagers can list Personal Spaces, which they shouldn't be able to. Might be another reva issue.

unfortunately that’s true: 
Screenshot 2026-06-18 at 18 38 34

v-scharf:
but it's not new. 
checked on demo `7.1.0 rolling` and localy on stable 4.0.7 - same behavior

rhafer:
> but it's not new.
> checked on demo `7.1.0 rolling` and localy on stable 4.0.7 - same behavior

I've created a new bug report for that: https://github.com/opencloud-eu/opencloud/issues/2979
For Space Admins it's currently also possible to update space attributes.

v-scharf (approved):

PR #3001, opened 2026-06-22
fix: make the collaboration service events optional

### LINKED ISSUE #2966: Collabora breaks without setting `OC_EVENTS_ENDPOINT`
## Describe the bug
Using external apps like Collabora without setting `OC_EVENTS_ENDPOINT` seems to break completely on current main. I guess this was introduced by https://github.com/opencloud-eu/opencloud/pull/2862. This is bad because it's introducing a breaking change.

## Steps to reproduce
1. Start a dev stack with collabora, e.g. in our `opencloud-compose` setup without having `OC_EVENTS_ENDPOINT` set (basically undo https://github.com/opencloud-eu/opencloud-compose/pull/307/changes)
2. Access web in the browser and open a odt file

## Expected behavior
The file opens. Collaboration events like for the mention feature don't work, but they shouldn't affect Collabora as a whole.

## Actual behavior
Collabora doesn't start.

fschade:
## Description

the collaboration service died without OC_EVENTS_ENDPOINT set,
this pr makes setting (and enabling) the endpoint optional.

## Related Issue

- Fixes https://github.com/opencloud-eu/opencloud/issues/2966#issuecomment-4751775996

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/collaboration/pkg/command/server.go
+++ b/services/collaboration/pkg/command/server.go
@@ -176,7 +176,9 @@ func Server(cfg *config.Config) *cobra.Command {
 
 			var optionalHTTPServerOptions []http.Option
 			var notificationService notification.Service
-			if cfg.Events.Endpoint != "" { // notifications are optional
+			if cfg.Events.Endpoint == "" {
+				logger.Warn().Msg("Events endpoint is not configured, notifications from the collaboration service will not work")
+			} else {
 				connName := generators.GenerateConnectionName(cfg.Service.Name, generators.NTypeBus)
 				natsStream, err := stream.NatsFromConfig(connName, true, stream.NatsConfig(cfg.Events))
 				if err != nil {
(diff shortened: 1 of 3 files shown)


services/collaboration/pkg/command/server.go
```diff
@@ -175,8 +174,9 @@ func Server(cfg *config.Config) *cobra.Command {
 				}
 			}
 
+			var optionalHTTPServerOptions []http.Option
 			var notificationService notification.Service
-			{
+			if cfg.Events.Endpoint != "" { // notifications are optional
```
kulmann: could you add an `else` with a fat warning that notifications wont' work because the events endpoint is not configured?

kulmann (changes requested):

services/collaboration/pkg/command/server.go
fschade: sure, no problem, can you add that, im a bit under preasure preparing a talk?

kulmann: done 👍

kulmann (approved):

PR #3019, opened 2026-06-25
chore(skills): add skill for bumping web

JammingBen:
Add a skill that let's agents bump web to a specified version, commit the changes and create a PR for it. The user invokes it via e.g. "bump web to v7.0.0". The agent then handles all the rest.

I successfully tried this in a dry run, but obviously no real world experience yet. So this might need some further adjustments down the line. Also, you need GitHub Cli (`gh`) installed on your system to be able to use this skill.

fyi @kulmann @ScharfViktor , since you also often do these bumps.


v-scharf (commented):
That is cool. I'm looking forward to testing it on the next rolling release.
Just one complaint from the static code analysis.

JammingBen:
> Just one complaint from the static code analysis.

It complains about the markdown formatting, but to be honest, I don't see any issue with that. Especially since there is no markdown formatter in this repository.

v-scharf (approved):

PR #3034, opened 2026-06-28
feat: support admin password file env

### LINKED ISSUE #2982: Add support for IDM_ADMIN_PASSWORD_FILE
## Is your feature request related to a problem? Please describe.
This Issue is related to this one [opencloud-compose/#306
](https://github.com/opencloud-eu/opencloud-compose/issues/306)

In short, the request was to add the variable IDM_ADMIN_PASSWORD_FILE in addition to the variable IDM_ADMIN_PASSWORD, so that, for example, Docker secrets could be used. Currently, `opencloud-compose` can pass `IDM_ADMIN_PASSWORD`, but file-based secret handling would require a shell wrapper or entrypoint workaround in the compose stack.

## Describe the solution you'd like
Add native server-side support for reading the admin password from a file.

   Suggested behavior:

   - Support `IDM_ADMIN_PASSWORD_FILE`
   - Optionally also support `ADMIN_PASSWORD_FILE`, matching the existing `ADMIN_PASSWORD` alias
   - If `*_FILE` is set, read the password from that file
   - Trim a trailing newline from the file content
   - If both `IDM_ADMIN_PASSWORD` and `IDM_ADMIN_PASSWORD_FILE` are set, return a clear configuration error instead of silently choosing one

   This would allow compose deployments to use Docker secrets like:

   ```yaml
   environment:
     IDM_ADMIN_PASSWORD_FILE: /run/secrets/idm_admin_password
   secrets:
     - idm_admin_password
 ```

## Describe alternatives you've considered
The alternative would be adding a shell wrapper in opencloud-compose that reads the secret file and exports IDM_ADMIN_PASSWORD before starting the server.

 I think this is less ideal because:

 - it only helps the compose stack
 - it duplicates secret-handling logic outside the server
 - it is harder to test and maintain
 - other deployment methods would not benefit from it

 Native server-side support would work for Docker Compose, Kubernetes, systemd, and any other deployment method that can mount secrets as files.

bxn-dev:
Adds file-based admin password support for the IDM admin user.

   Supported env vars:

   - `ADMIN_PASSWORD_FILE` for `opencloud init`
   - `IDM_ADMIN_PASSWORD_FILE` for `opencloud init`
   - `IDM_ADMIN_PASSWORD_FILE` for the IDM service config parser

   The file content is read and trailing `\n` / `\r\n` is trimmed. Supplying both a direct password env var and a password file env
 var returns a config error.

   ## Related Issue

   - Fixes https://github.com/opencloud-eu/opencloud/issues/2982

   ## Motivation and Context

   This allows deployments to provide the IDM admin password via Docker secrets or mounted secret files instead of exposing the
 password directly through environment variables.

   ## How Has This Been Tested?

   - test environment: local checkout on branch `feat/2982_idm_admin_passwort_file`
   - test case 1: helper tests for no file env, file env, trailing newline trimming, direct-value conflict, and missing file
   - test case 2: IDM parser test for `IDM_ADMIN_PASSWORD_FILE`
   - test case 3:

   ```sh
   go test ./pkg/config ./opencloud/pkg/command ./services/idm/pkg/...
 ```

 Screenshots (if appropriate):

 N/A

 Types of changes

 - [ ] Technical debt

 Checklist:

 - [x] Code changes
 - [x] Unit tests added
 - [ ] Acceptance tests added
 - [ ] Documentation added
 ```

### DIFF
--- a/opencloud/pkg/command/init.go
+++ b/opencloud/pkg/command/init.go
@@ -38,7 +38,11 @@ func InitCommand(_ *config.Config) *cobra.Command {
 			quietFlag, _ := cmd.Flags().GetBool("quiet")
 			configPathFlag := viper.GetString("config-path")
 			adminPasswordFlag := viper.GetString("admin-password")
-			err := ocinit.CreateConfig(insecure, forceOverwriteFlag, diffFlag, configPathFlag, adminPasswordFlag, quietFlag)
+			adminPasswordFlag, _, err := config.ReadFileEnv(adminPasswordFlag, "ADMIN_PASSWORD_FILE", "IDM_ADMIN_PASSWORD_FILE")
+			if err != nil {
+				return err
+			}
+			err = ocinit.CreateConfig(insecure, forceOverwriteFlag, diffFlag, configPathFlag, adminPasswordFlag, quietFlag)
 			if err != nil {
 				log.Fatalf("Could not create config: %s", err)
 			}
--- a/pkg/config/helpers.go
+++ b/pkg/config/helpers.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"io/fs"
 	"os"
 	"path"
@@ -49,6 +50,29 @@ func bindSourcesToStructs(fileSystem fs.FS, filePath, service string, dst any) e
 	}
 
 	return nil
+}
+
+// ReadFileEnv returns value or the trimmed contents of the first set file env var.
+func ReadFileEnv(value string, fileEnvNames ...string) (string, bool, error) {
+	for _, name := range fileEnvNames {
+		file, ok := os.LookupEnv(name)
+		if !ok {
+			continue
+		}
+
+		if value != "" {
+			return "", true, fmt.Errorf("%s cannot be used together with direct value", name)
+		}
+
+		content, err := os.ReadFile(file)
+		if err != nil {
+			return "", true, fmt.Errorf("read %s: %w", name, err)
+		}
+
+		return strings.TrimRight(string(content), "\r\n"), true, nil
+	}
+
+	return value, false, nil
 }
 
 // LocalEndpoint returns the local endpoint for a given protocol and address.
--- a/pkg/config/helpers_test.go
+++ b/pkg/config/helpers_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"os"
 	"testing"
 	"testing/fstest"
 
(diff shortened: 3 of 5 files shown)


opencloud/pkg/command/init.go
```diff
@@ -38,7 +38,11 @@ func InitCommand(_ *config.Config) *cobra.Command {
 			quietFlag, _ := cmd.Flags().GetBool("quiet")
 			configPathFlag := viper.GetString("config-path")
 			adminPasswordFlag := viper.GetString("admin-password")
-			err := ocinit.CreateConfig(insecure, forceOverwriteFlag, diffFlag, configPathFlag, adminPasswordFlag, quietFlag)
+			adminPasswordFlag, _, err := config.ReadFileEnv(adminPasswordFlag, "ADMIN_PASSWORD_FILE", "IDM_ADMIN_PASSWORD_FILE")
+			if err != nil {
+				return err
+			}
+			err = ocinit.CreateConfig(insecure, forceOverwriteFlag, diffFlag, configPathFlag, adminPasswordFlag, quietFlag)
```
parasite-lost: *suggestion*: the following would preserve the original error behaviour

```
			adminPasswordFlag, _, err := config.ReadFileEnv(adminPasswordFlag, "ADMIN_PASSWORD_FILE", "IDM_ADMIN_PASSWORD_FILE")
			if err == nil {
				err = ocinit.CreateConfig(insecure, forceOverwriteFlag, diffFlag, configPathFlag, adminPasswordFlag, quietFlag)
			}
```

pkg/config/helpers.go
```diff
@@ -51,6 +52,29 @@ func bindSourcesToStructs(fileSystem fs.FS, filePath, service string, dst any) e
 	return nil
 }
 
+// ReadFileEnv returns value or the trimmed contents of the first set file env var.
+func ReadFileEnv(value string, fileEnvNames ...string) (string, bool, error) {
```
parasite-lost: *question*: is there a purpose to the bool return value I am missing?

It looks like it is generally ignored in production code. Its semantics seem to be essentially just confirming that one of the provided environment variables has been set. Since it is only used in test code, where it is already obvious from the test setup whether such a variable has been set or not and the correct behaviour is observable from the other two return values, is this bool return value really needed?

pkg/config/helpers_test.go
```diff
@@ -13,6 +14,64 @@ type TestConfig struct {
...
+	})
+
+	t.Run("missing file", func(t *testing.T) {
+		t.Setenv("SECRET_FILE", t.TempDir()+"/missing")
+
+		_, ok, err := ReadFileEnv("", "SECRET_FILE")
+		assert.Equal(t, ok, true)
+		assert.ErrorContains(t, err, "read SECRET_FILE")
+	})
+}
```
parasite-lost, line 76: *suggestion*: consider adding two more tests:
* two environment variable passed, both are set (first file is read)
* two environment variable passed, first is not set, second is set (second file is read)

parasite-lost (changes requested):

pkg/config/helpers.go
bxn-dev: No, you're right. I'll remove the bool.

pkg/config/helpers_test.go
bxn-dev: I'll add those. Thanks for the feedback.

PR #3068, opened 2026-07-06
add posixfs index command

butonic:
The `opencloud posix index` command allows adding spaces on disk to the posixfs space `by-type`, `by-user-id` and `by-group-id` indexes. It needs to be run when restoring a space from a backup.

### DIFF
--- a/opencloud/pkg/command/posixfs.go
+++ b/opencloud/pkg/command/posixfs.go
@@ -19,6 +19,7 @@ import (
 	"github.com/opencloud-eu/reva/v2/pkg/storage/fs/posix/ignore"
 	"github.com/opencloud-eu/reva/v2/pkg/storage/fs/posix/options"
 	"github.com/opencloud-eu/reva/v2/pkg/storage/fs/registry"
+	"github.com/opencloud-eu/reva/v2/pkg/storage/utils/decomposedfs/spaceidindex"
 
 	"github.com/pkg/xattr"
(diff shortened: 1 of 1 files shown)


opencloud/pkg/command/posixfs.go
```diff
@@ -138,6 +140,148 @@ func scanCmd(ocCfg *config.Config) *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   "index",
+		Short: "Add spaces from the filesystem to the index",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return rebuildIndexes(cmd)
+		},
+	}
+	cmd.Flags().StringP("root", "r", "", "Path to the root directory of the posixfs storage")
+	cmd.Flags().StringSliceP("dirs", "d", []string{"users", "projects"}, "Subdirectories to scan for spaces")
+	cmd.Flags().StringP("space", "s", "", "Only index a specific space by ID")
```
aduffeck, line 153: Is there a reason why we're filtering by ID instead of directory name? They don't have to match and considering the backup-restore case I would think the directory name is what you have on hand anyway.

dragonchaser (approved):

aduffeck (approved):

opencloud/pkg/command/posixfs.go
butonic: true, I'll allow dirname and spaceid

Issue #2409, opened 2026-02-27
No Etag after Last Chunk


zerox80:
The Tusd library has a hook called PreFinishResponseCallback, with that REVA could calculate the etag after the last chunk and write it to the response header, but this hook is never used....

How are the plans? Both the Desktop and soon Android Client already support that way, with a fallback using Propfind to force the etag from server. But that is a another Request, causing Latency. 

in tusd library, vendor/config.go

and

handler/unrouted_handler.go

It would be a very nice stuff if this could be implemented? 

https://github.com/opencloud-eu/android/pull/85

At bottom we already discussed why this isn't working out of the box.

I really appreciate your thoughts!

micbar:
@kulmann @dragotin What do you think?

micbar:
@zerox80 why is this a bug?

guruz:
I changed type to Enhancement (as in Performance enhancement).
It would save one round trip to get the ETag after the last chunk.


(@dragotin For nostalgia.. Note that we kind of already had this with the old chunking protocol. https://dragotin.wordpress.com/2015/06/22/owncloud-chunking-ng/ 
Or maybe not, that one had a MOVE and here it is just PATCHes.. :-) )

zerox80:
> [@zerox80](https://github.com/zerox80) why is this a bug?

Not a real bug, i just didn't know if i should tag is as "feature request" "bug" or "enhancement".

zerox80:
Hi, is there any update?

michaelstingl:
I went ahead and prototyped this on the reva side, as a proposal to make the discussion concrete rather than a finished solution:

- https://github.com/opencloud-eu/reva/pull/666

It registers the `PreFinishResponseCallback` from the issue description so the chunked finalize returns the same etag and permissions as the creation-with-upload path. The PR carries one open design question: the data path is authenticated by the transfer token, which has no user identity, so the callback restores the upload's executant to stat the file; a small storage interface could surface the etag instead.

**Edit (2026-06-30):** the storage-interface version is now up as a separate PR, with the earlier one kept for comparison:

- https://github.com/opencloud-eu/reva/pull/718

<sub>🤖 drafted with Claude Code, reviewed before submitting.</sub>

guruz:
**EDIT: Ignore my comment, the mtime is just set to currentTime and resolveFinalEtagIfNeeded really only does ETag**

> with a fallback using Propfind to force the etag from server. 

Just thinking aloud here @zerox80 

[...]

PR #3078, opened 2026-07-06
test(coreApiWebdavUploadTUS): assert etag and permissions on the finalizing TUS chunk

michaelstingl:
> [!WARNING]
> This scenario asserts headers that only a reva carrying the following change returns on the chunked TUS finalize:
>
> - https://github.com/opencloud-eu/reva/pull/718
>
> This repo's reva pin (`v2.46.4-0.20260625152426-8cff2a7032ec`, 2026-06-25) predates that merge, so the scenario is red until the pin is bumped past it. It goes green with the bump; no code change here.

## Description

Adds one scenario to `tests/acceptance/features/coreApiWebdavUploadTUS/lowLevelUpload.feature`: after a chunked TUS upload's finalizing PATCH, the response carries `OC-ETag`, `ETag` and `OC-Perm`, so a client does not need a follow-up PROPFIND for the new etag and permissions.

Reuses existing step definitions only (TUS create + `sends a chunk` + `the following headers should match these regular expressions`); the etag is a dynamic quoted hash, hence the regex assertion.

## Related Issue

Locks the behaviour from #2409 (closed) in the acceptance suite. The fix itself is in reva:

- https://github.com/opencloud-eu/reva/pull/718

## Motivation and Context

The chunked finalize returning the etag was covered only by a reva unit test. This pins it at the product boundary so a later regression fails the suite, and states the intended behaviour as an executable spec rather than an implicit claim.

## How Has This Been Tested?

Locally, one feature against a running server, on two builds:

- **stock `opencloud-rolling:7.1` (before the reva change):** create, chunks and `204` pass; the three header rows fail because the headers are absent.
- **opencloud built on a reva carrying the change:** the whole `lowLevelUpload.feature` passes on posix and decomposed, the new scenario on old/new/spaces.

## Types of changes

- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation added (n/a: tests-only, no changelog fragment)

<sub>🤖 drafted with Claude Code, reviewed before submitting.</sub>

### DIFF
--- a/tests/acceptance/features/coreApiWebdavUploadTUS/lowLevelUpload.feature
+++ b/tests/acceptance/features/coreApiWebdavUploadTUS/lowLevelUpload.feature
@@ -90,3 +90,23 @@ Feature: low level tests for upload of chunks
       | old              |
       | new              |
       | spaces           |
+
+  @issue-2409
+  Scenario Outline: finalize a chunked upload and get the etag and permissions
+    Given using <dav-path-version> DAV path
+    And user "Alice" has created a new TUS resource on the WebDAV API with these headers:
+      | Upload-Length   | 10                    |
+      #    ZmlsZS50eHQ= is the base64 encode of file.txt
+      | Upload-Metadata | filename ZmlsZS50eHQ= |
+    When user "Alice" sends a chunk to the last created TUS Location with offset "0" and data "123" using the WebDAV API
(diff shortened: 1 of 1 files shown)


rhafer:
@michaelstingl reva has be bumped meanwhile, please rebase to check if the tests are green now.

v-scharf (approved):

PR #3091, opened 2026-07-08
chore: don't log non-error message at error level

rhafer:
When walking the tree for creating the search index, ignore "not found" errors. Search indexing happens asynchronously so the underlying space, file or folder might just not exist anymore when the indexing starts.

### DIFF
--- a/services/search/pkg/search/service.go
+++ b/services/search/pkg/search/service.go
@@ -2,6 +2,7 @@ package search
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


butonic (approved):

fschade (dismissed):
can’t we use errors is?

rhafer:
> can’t we use errors is?

@fschade Not without changes to the `errtypes` package in reva I think.

What is your concern? Are you thinking about wrapper errors here? Then I guess we could leverage `errors.As`.

fschade:
> > can’t we use errors is?
> 
> @fschade Not without changes to the `errtypes` package in reva I think.
> 
> What is your concern? Are you thinking about wrapper errors here? Then I guess we could leverage `errors.As`.

Yes, exactly, I just didn't want us to overlook it.

rhafer:
@fschade PTAL

aduffeck (approved):

Issue #3092, opened 2026-07-08
Search engine schema update documentation // migration


fschade:
updating the search engine schema right now is a painful process which could include re-indexing the whole instance, making it batteries included is not an option in my opinion because we have no insights what each individual installation contains and what were allowed to touch or not.

Because of this we have to document one or two cases how such an schema update could look like, what needs to be done and evaluate feasible ways of a non re-indexing workflow.

This needs to be done for:

- bleve search (our default build in search engine) 
- openSearch

dschmidt:
We have an explicit message, that opensearch index is incompatible.
Can/should we have an explicit message for bleve as well at least?
I would like to avoid silent failure or an unreadable (and "unactionable" error message).

I agree that a batteries included is very hard to get right, especially when thinking of kubernetes environments with multiple instances of the backend: which instance is allowed to update?

butonic:
We discussed today how we can handle search index migration in a non breaking way. We could move the breaking change to a v2 package and keep the old implementation in place as the default? the process can start and we can nag the admin (IIRC we can now show messages to the admin) to rebuild the index and then flip an env var switch to move to the new search index. 

For main we could bring back the old index with a few bugfixes, right before we rewrote the service implementation. 

This approach can be used for future migrations, as long as they only affect the index and the backend code. The web ui might not find all new properties when the index changes or the search results might be too unspecific eg. [for paths, but there will be more migrations in the future](https://github.com/opencloud-eu/opencloud/pull/3495). This would be the first step to handle them gracefully in large deployments.

The next step would be to be able to maintain both: the old and the new index and let the migration set a flag on a space to indicate which search index version should be used for search. The problem is that we would have to aggregate results in process and query both backends. Opensearch already supports merging results from multiple indexes but the mechanism cannot handle type changes. This does need more research, but being able to start building the index at runtime is the first step.

dschmidt:
FWIW: I've extracted the most important breaking change out of the path pr referenced above into #3510. 
Stumbled upon the issue only when someone reported it and tried to fix it by treating a symptom. To be clear: it's not a regression in my work, it was a preexisting issue. 

Everything I "wanted" to break was in #3345. All my other PRs are non breaking and strictly additive

PR #3098, opened 2026-07-08
feat(search): check the index schema on startup and refuse breaking changes

dschmidt:
**Superseded by #3197.** Reopened from the upstream repo with identical content and target so it can serve as a native-stack parent.

---

Stacked on #2659. The base `tmp/refactor-search-mapping` is a temporary upstream snapshot of the #2659 branch so this PR shows only its own commits; it gets retargeted to `main` (and the temp branch deleted) once #2659 is merged. Both PRs need to ship in the same release. Refs #3092.

Two complementary mechanisms keep the index schema and the code in sync.

**Versioned indices (from the base #2659).** The index name/path carries the schema version from `search.SchemaVersion` (OpenSearch `opencloud-resource-v3`, bleve `bleve-v3`). An intended breaking schema change is made by bumping the constant: the service targets a fresh, empty index and leaves the old one in place. No migration, no index-to-index reindex; content lives only in the index, the files are the source of truth, so a rescan (`opencloud search index --all-spaces`) rebuilds everything into the new index.

**Startup schema check (this PR).** The version bump is manual, so this is the safety net for the schema changing without one. On startup both engines diff the existing (current-version) index schema against the schema generated from code, using one shared recursive classifier:

- **equal**: start normally.
- **additive** (new fields with no indexed data): applied in place, no version bump needed. OpenSearch: `PUT _mapping` with the full code properties. bleve: the code mapping is persisted into the index (`SetInternal` + reopen), so new fields are typed correctly right away and the next start classifies equal. A startup warning lists the new fields; documents indexed before the upgrade lack them until re-indexed.
- **breaking** (changed types/analyzers, removed or renamed fields, new fields that already contain data of unknown shape): refuse to start. Because the index is versioned, a released instance never reaches this: an intended break bumps `SchemaVersion` and gets a fresh index. It fires in development when the mapping is changed in a breaking way without bumping, so the error is developer-facing (bump `search.SchemaVersion`, or revert the change) and lists the diff that triggered it.

Why this shape:

- The classifier is the only oracle. `PUT _mapping` is just the apply mechanism: its merge semantics hide removals/renames and it acks in-place updatable param changes, so it must not judge.
- bleve keeps no schema trace for dynamically indexed fields, so the classifier additionally checks `idx.Fields()` (exact name + path prefix). A field that is explicitly mapped now but already holds dynamically indexed data of unknown shape is breaking rather than a source of silently wrong results, no search beats wrong search.
- No migration framework, no index-to-index reindex, no blue-green: versioning provides the fresh index, a rescan repopulates it.

Known and accepted, no code changes:

- Non-JSON 5xx proxy bodies lose their status code in the opensearch-go error parsing; irrelevant while every non-conflict error is plain fatal.
- Trashed items are missing from the index after a rebuild until they are restored, the rescan walks only the live tree (docs caveat for #3092).
- The bleve datapath must not be shared between processes; a second opener now fails after 5s (`bolt_timeout`) instead of hanging forever.

Deliberate consequences of the classifier rules (not bugs):

- Additive upgrades are one-way within a schema version: once the new release widened the index schema (`PUT _mapping` / persisted bleve mapping), the previous release sees a stored-only field and refuses to start. Roll forward or rebuild; refuse-in-both-directions is what keeps mismatched schemas from ever serving queries.
- Adding a new custom analyzer classifies as breaking even when only a brand-new field uses it: OpenSearch cannot change analysis settings on an open index, and both engines must behave identically.
- Removing or renaming a field in code is breaking by design; tolerating removals would make renames undetectable (a rename is a removal plus an addition).


dschmidt:
Reopened from the upstream repo as #3197 (identical content and target). The head branch now lives in the upstream repo instead of my fork so follow-up PRs can natively stack on it. Continuing there.

PR #3113, opened 2026-07-11
graph: expose following state, tags and allowed actions on driveItems

dschmidt:
Implements opencloud-eu/libre-graph-api#37 and opencloud-eu/libre-graph-api#39: driveItems now carry the current user's favorite state and the assigned tags (read from ArbitraryMetadata), plus `@libre.graph.permissions.actions.allowedValues` when requested via `$select`.

### DIFF
--- a/services/graph/pkg/service/v0/driveitems.go
+++ b/services/graph/pkg/service/v0/driveitems.go
@@ -29,7 +29,23 @@ import (
 
 	"github.com/opencloud-eu/opencloud/pkg/log"
 	"github.com/opencloud-eu/opencloud/services/graph/pkg/errorcode"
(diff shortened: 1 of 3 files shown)


services/collaboration/pkg/connector/fileconnector.go
```diff
@@ -1578,8 +1578,6 @@ func (f *FileConnector) GetAvatar(ctx context.Context, userID string) (*Connecto
 		}
 		return nil, NewConnectorError(http.StatusBadGateway, "failed to fetch avatar")
 	}
-	defer photoFile.Close()
```
dschmidt: https://github.com/opencloud-eu/libre-graph-api/pull/47

micbar (approved):
LGTM

PR #3115, opened 2026-07-11
fix: oidcHTTPClient now use proxy from environment

### LINKED ISSUE #3114: oidcHTTPClient doesn't use proxy settings from environment (http_proxy) causing authentication with external IdP to fail
## Describe the bug
oidcHTTPClient doesn't use proxy settings from environment (http_proxy), causing connection and timeout issues during authentication with external IdP for systems without internet access or otherwise limited firewall access, or where the use of a forward proxy is required.

In Go, the default http.Transport uses http.ProxyFromEnvironment as value for Proxy field. This allows using a proxy defined in the HTTP_PROXY or HTTPS_PROXY environment in the entire application. The issue is that oidcHTTPClient uses a new contructed http.Transport where Proxy is left `nil`. 

## Steps to reproduce
1. setup opencloud on a system without internet access, outgoing http requests must go via http proxy
2. configure external idp
3. authenticate with a user, observe the following error message in opencloud: 
```
2026-07-10T14:11:34Z ERR failed to authenticate the request error="failed to verify access token: Get \"https://auth.example.com/.well-known/openid-configuration\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)" authenticator=oidc client.address="10.0.0.1" network.peer.address= network.peer.port= path=/graph/v1.0/me service=proxy user_agent="Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36"
```

## Expected behavior
I expect any outgoing http requests to use the environment variables:
- http_proxy
- https_proxy
- no_proxy

## Actual behavior
Proxy settings are ignored and HTTP request and authentication fails

## Setup
OpenCloud configured to use Authelia as external IdP after successfully using internal IdP for a few weeks. Authentication works after adding a temporary dynamic firewall rule (external IdP using Cloudflare proxy without static IP).

<details>
<p>

```console
      HTTP_PROXY={{ proxy_url }}
      HTTPS_PROXY={{ proxy_url }}
      NO_PROXY=localhost,127.0.0.1,host.example.com
      OC_OIDC_ISSUER={{ opencloud_oidc_url }}
      PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD=none
      PROXY_AUTOPROVISION_ACCOUNTS=true
      PROXY_ROLE_ASSIGNMENT_DRIVER=oidc
      PROXY_USER_OIDC_CLAIM=preferred_username
      PROXY_USER_CS3_CLAIM=username
      PROXY_ROLE_ASSIGNMENT_OIDC_CLAIM=groups
      WEB_OIDC_CLIENT_ID={{ opencloud_oidc_client_id }}
      WEB_OIDC_SCOPE=openid profile email groups
      OC_EXCLUDE_RUN_SERVICES=idp
```

</p>
</details>

## Additional context
Systems are hosted in an environment where the baseline is no internet or firewall access in or out. Firewall rules for static IPs can be added when required, but for endpoints with non-static IPs, using a http proxy is required. Since our external IdP is protected by Cloudflare Proxy without a static IP, our OpenCloud instance needs to use a http proxy to access it.

sigurdvaa:
## Description
Update transport setting for oidcHTTPClient so it can use proxy setting from the environment. This can be required for systems with limited firewall access, so that oidc issuer can be reached via a proxy.

## Related Issue

- Fixes https://github.com/opencloud-eu/opencloud/issues/3114

## Motivation and Context
Makes it possible to authenticate using external IdP when OpenCloud system needs to use a http proxy for outoing http requests.

## How Has This Been Tested?

- Needed firewall rule for authentication with external IdP even though OpenCloud system was configured to use http proxy. Verified Firewall rule was no longer needed after applying patch.    

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/proxy/pkg/command/server.go
+++ b/services/proxy/pkg/command/server.go
@@ -104,6 +104,7 @@ func Server(cfg *config.Config) *cobra.Command {
 						InsecureSkipVerify: cfg.OIDC.Insecure, //nolint:gosec
 					},
 					DisableKeepAlives: true,
+					Proxy: http.ProxyFromEnvironment,
 				},
 				Timeout: time.Second * 10,
 			}
@@ -279,6 +280,7 @@ func loadMiddlewares(logger log.Logger, cfg *config.Config,
 				InsecureSkipVerify: cfg.OIDC.Insecure, //nolint:gosec
 			},
 			DisableKeepAlives: true,
+			Proxy: http.ProxyFromEnvironment,
 		},
 		Timeout: time.Second * 10,
 	}


micbar (approved):
Nice catch!

PR #3140, opened 2026-07-15
fix(runtime): log service startup errors instead of printing them beside the log

dschmidt:
Service startup errors were printed as plain text into the log stream (`SilenceErrors`/`SilenceUsage` unset on the cobra app), and the one record about the failure carried `"error":{}` (suture's `ev.Err` is an `interface{}`, so it got marshaled instead of passed to `Err()`).

The example below is the schema mismatch error from #3098, but this is not specific to it: every startup error of every service has this shape.

before:

```
{"level":"warn","service":"proxy","time":"...","message":"No tls certificate provided, using a generated one"}
Error: manual action required: search index /var/lib/opencloud/search/bleve was built with a different schema:
  - field Deleted is explicitly mapped now, but the index already holds data that was indexed dynamically for it, of an unknown type
  [... 10 more ...]
There is no in-place migration: with the OpenCloud search service stopped, delete the index directory /var/lib/opencloud/search/bleve, then start it again ...
Usage:
  search server [flags]

Flags:
  -h, --help   help for server

{"level":"error","service":"opencloud","event":"opencloud: Failed service 'search' (1.000000 failures of 5.000000), restarting: true, error: manual action required: search index ...","service":"search","supervisor":"opencloud","restarting":true,"failures":1,"threshold":5,"error":{},"time":"...","message":"service terminated"}
```

after:

```
{"level":"error","service":"opencloud","event":"opencloud: Failed service 'search' (1.000000 failures of 5.000000), restarting: true, error: manual action required: search index ...","service":"search","supervisor":"opencloud","restarting":true,"failures":1,"threshold":5,"error":"manual action required: search index /var/lib/opencloud/search/bleve was built with a different schema:\n  - field Deleted is explicitly mapped now, but the index already holds data that was indexed dynamically for it, of an unknown type\n  [... 10 more ...]\nThere is no in-place migration: with the OpenCloud search service stopped, delete the index directory /var/lib/opencloud/search/bleve, then start it again ...","time":"...","message":"service terminated"}
```

### DIFF
--- a/opencloud/pkg/runtime/service/service.go
+++ b/opencloud/pkg/runtime/service/service.go
@@ -389,9 +389,16 @@ func Start(ctx context.Context, o ...Option) error {
 				if ev.Restarting {
 					l = s.Log.Error()
 				}
-				l.Str("event", e.String()).Str("service", ev.ServiceName).Str("supervisor", ev.SupervisorName).
-					Bool("restarting", ev.Restarting).Float64("failures", ev.CurrentFailures).Float64("threshold", ev.FailureThreshold).
-					Interface("error", ev.Err).Msg("service terminated")
+				l = l.Str("event", e.String()).Str("service", ev.ServiceName).Str("supervisor", ev.SupervisorName).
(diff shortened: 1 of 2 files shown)


aduffeck (approved):
lgtm

PR #3142, opened 2026-07-15
fix(search): report rejected documents when indexing to OpenSearch

dschmidt:
`Batch.Push` discarded the bulk response. The bulk API answers 200 even when it rejects individual items, so a rejected document went missing from the index without a trace. `Purge` already checks `resp.Failures` for `DeleteByQuery`, this does the same for `Bulk`.

While at it: those `DeleteByQuery` failures were printed as byte arrays, `resp.Failures` is a `[]json.RawMessage`.

This only makes the loss visible, it does not prevent it. The event is acked either way and nothing retries.

### DIFF
--- a/services/search/pkg/opensearch/batch.go
+++ b/services/search/pkg/opensearch/batch.go
@@ -211,12 +211,21 @@ func (b *Batch) Push() error {
 		case err != nil:
 			return fmt.Errorf("failed to execute bulk operations: %w", err)
 		case resp.Errors:
-			items, err := json.Marshal(resp.Items)
-			if err != nil {
-				return fmt.Errorf("failed to marshal bulk response: %w", err)
+			var failed []opensearchgoAPI.BulkRespItem
+			for _, item := range resp.Items {
+				for _, result := range item {
+					if result.Error != nil {
+						failed = append(failed, result)
+					}
+				}
 			}
 
-			return fmt.Errorf("failed to execute bulk operations, response: %s", items)
+			failures, err := json.Marshal(failed)
+			if err != nil {
+				return fmt.Errorf("failed to marshal bulk failures: %w", err)
+			}
+
+			return fmt.Errorf("failed to execute bulk operations, failures: %s", failures)
 		}
 
 		bulkOperations = nil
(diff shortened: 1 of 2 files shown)


services/search/pkg/opensearch/batch.go
```diff
@@ -167,7 +167,7 @@ func (b *Batch) Purge(id string, onlyDeleted bool) error {
 			case err != nil:
 				return fmt.Errorf("failed to delete by query: %w", err)
 			case len(resp.Failures) != 0:
-				return fmt.Errorf("failed to delete by query, failures: %v", resp.Failures)
+				return fmt.Errorf("failed to delete by query, failures: %s", resp.Failures)
```
dschmidt: `resp.Failures` is a `[]json.RawMessage`, and `json.RawMessage` is a `[]byte` without a `String()` method, so `%v` rendered the failures as byte arrays:

```
[[123 34 105 110 100 101 120 34 58 34 111 112 101 110 ...]]
```

`%s` prints what the API actually returned:

```
[{"index":"opencloud-demo-dbq","id":"1$1!3","cause":{"type":"cluster_block_exception","reason":"index [opencloud-demo-dbq] blocked by: [FORBIDDEN/8/index write (api)];"},"status":403}]
```

services/search/pkg/opensearch/batch.go
```diff
@@ -204,10 +204,19 @@ func (b *Batch) Push() error {
...
+		switch {
+		case err != nil:
 			return fmt.Errorf("failed to execute bulk operations: %w", err)
+		case resp.Errors:
+			items, err := json.Marshal(resp.Items)
+			if err != nil {
+				return fmt.Errorf("failed to marshal bulk response: %w", err)
+			}
+
+			return fmt.Errorf("failed to execute bulk operations, response: %s", items)
```
dschmidt: `resp.Items` holds every item, not only the rejected ones, hence `response:` and not `failures:`. A batch is up to `defaultBatchSize` (50), so a single rejected document drags 49 successful ones into the log.

If that is too noisy, filter for the rejected items (compiles, `TestBatch_Push` stays green):

```suggestion
			var failed []opensearchgoAPI.BulkRespItem
			for _, item := range resp.Items {
				for _, result := range item {
					if result.Error != nil {
						failed = append(failed, result)
					}
				}
			}

			failures, err := json.Marshal(failed)
			if err != nil {
				return fmt.Errorf("failed to marshal bulk failures: %w", err)
			}

			return fmt.Errorf("failed to execute bulk operations, failures: %s", failures)
```

PR #3144, opened 2026-07-16
add header to release pr. remove 📦️ Dependencies from release descrip…

v-scharf:
- [x] added header: `This is a rolling release. Learn here about the`
tested here: https://github.com/v-scharf/rrg-test/pull/2

  
  
  - Now we have header in PR, in release description- ✅️

image

  -  but not in changelog - ✅️:
image

- [x] excluded `📦️ Dependencies` from release pr and release description  -> but stills in changelog ✅️

### DIFF
--- a/release-config.ts
+++ b/release-config.ts
@@ -73,6 +73,38 @@ export default {
     const latestTag = tags.pop() || "v0.0.0";
     return compareVersions(latestTag, nextVersion) === -1;
   },
+  getReleaseDescription: async ({ nextVersion }) => {
+    const note =
+      "> [!IMPORTANT]\n" +
+      "> This is a rolling release. Learn here about the [release types and lifecycle](https://docs.opencloud.eu/docs/admin/resources/lifecycle#release-types).\n\n";
+
+    // Exclude the "📦️ Dependencies" section from the release description
+    const section = removeChangelogSubsection(
+      await getChangelogSection(nextVersion),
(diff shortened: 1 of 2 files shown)


release-config.ts
```diff
@@ -73,6 +73,38 @@ export default {
...
+  getReleaseDescription: async ({ nextVersion }) => {
+    const note =
+      "> [!IMPORTANT]\n" +
+      "> This is a rolling release. Learn here about the [release types and lifecycle](https://docs.opencloud.eu/docs/admin/resources/lifecycle#release-types).\n\n";
+
+    // Exclude the "📦️ Dependencies" section from the release description
+    const section = removeChangelogSubsection(
+      await getChangelogSection(nextVersion),
+      "📦️ Dependencies"
+    );
```
JammingBen, line 85: You could just remove "📦️ Dependencies" from `changeTypes` in this file. It would never land in `CHANGELOG.md` in the first place. Then you don't need to manually strip out strings.

Or do we want to keep it in `CHANGELOG.md`? 🤔

v-scharf: > You could just remove "📦️ Dependencies" from changeTypes in this file.

Yes, I know, I just wanted to exclude "📦️ Dependencies" from release but keep it in changelog.md (I supposed it'll usefull for us dev teams)

JammingBen (approved):

PR #3146, opened 2026-07-16
feat(search): search index migration for both backends

dschmidt:
Stacked on #3098 (base = #3098 + the Mtime `*time.Time` fix), so this shows only its own commits.

In-process index migration for bleve (dir swap) and OpenSearch (alias flip), triggered by a shared `search.SchemaRevision` bump, with an `opencloud search migrate` CLI. Reindexes from the existing index instead of a full Tika rescan, preserves trash, paginates with search_after, logs start/progress/end.

bleve migrates on startup by default (opt-out via `SEARCH_ENGINE_BLEVE_AUTO_MIGRATE=false`; safe because a single instance holds the data-path lock). OpenSearch is CLI-only, meant as a pre-rollout hook: the migration builds the new index and flips the alias atomically while the old index stays live, so the new version starts against a ready index.

The `Apply` schema classifier stays as the runtime safety net: it refuses to start on an index that doesn't match the code schema (e.g. a forgotten revision bump), rather than serving wrong results.

Unsure about the last commit (refresh/replicas off during reindex, ~19% on a 2-node cluster) being worth the extra code; happy to drop it.

Not here (follow-up): live-write catch-up during migration (seq_no delta) and the readiness gate.


dschmidt:
We're most likely not going down this road for now.
Just keeping it open for reference until the other refactor PRs are merged 

- https://github.com/opencloud-eu/opencloud/pull/2659
- https://github.com/opencloud-eu/opencloud/pull/3098

dschmidt:
The SchemaVersion handling has been moved to #2659 and the NewClient refactor to #3098.
As we're not doing the migration at this point, I think we can safely close this PR.

PR #3158, opened 2026-07-21
Only log a debug message when an item is still in processing state

aduffeck:
Depends on https://github.com/opencloud-eu/reva/pull/742.


rhafer:
@aduffeck rebase? Reva has been bumped meanwhile

rhafer (approved):

PR #3189, opened 2026-07-28
feat: add announcement banner

dschmidt:
Runtime managed announcement banner shown above the top bar to all users, with a dialog showing Markdown detail text.

Model: `{ enabled, bannerText, infoText }`, persisted in a NATS JetStream key-value store.

- `GET /announcement` (authenticated, `Announcement.ReadWrite` permission): the full stored state, for the admin UI.
- `PUT /announcement`: upsert; an empty banner text removes the record. There is no DELETE endpoint (an empty banner text is the implicit delete, and `enabled` hides/shows without deleting).
- `config.json`: `options.announcement.{bannerText, infoText}` is injected only while `enabled` and a banner text are set, so prepared/disabled announcements stay private. The store is the single source of truth: config.json always reflects it, and a statically configured `web.config.options.announcement` is ignored (the field is `yaml:"-"`, output only).

## Public by design
The message is part of the unauthenticated `config.json` (loaded before login), so the banner also shows on the login and other pre-auth pages, e.g. to announce that a login provider is unavailable. Admins must not put sensitive information into it.

Web frontend: opencloud-eu/web#2967


services/web/pkg/announcement/service.go
```diff
@@ -0,0 +1,162 @@
...
+		store:           options.store,
+		gatewaySelector: options.gatewaySelector,
+	}, nil
+}
+
+// Get returns the full stored announcement (including disabled ones) for management.
+func (s Service) Get(w http.ResponseWriter, r *http.Request) {
+	gatewayClient, err := s.gatewaySelector.Next()
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
aduffeck, line 94: Maybe log the errors here and below before returning?

services/web/pkg/service/v0/service.go
```diff
@@ -75,6 +86,14 @@ func NewService(opts ...Option) (Service, error) {
...
 			r.Delete("/", themeService.LogoReset)
 		})
+		r.Route("/announcement", func(r chi.Router) {
+			r.Use(middleware.ExtractAccountUUID(
+				account.Logger(options.Logger),
+				account.JWTSecret(options.Config.TokenManager.JWTSecret),
+			))
+			r.Get("/", announcementService.Get)
+			r.Put("/", announcementService.Set)
+		})
```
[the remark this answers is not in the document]
aduffeck: I think Copilot has a point here. How does web clear the announcement, does it just disable it?

services/web/pkg/announcement/announcement.go
```diff
@@ -0,0 +1,74 @@
...
+type Announcement struct {
+	// Enabled controls whether the announcement is live (injected into config.json).
+	Enabled    bool   `json:"enabled"`
+	BannerText string `json:"bannerText"`
+	InfoText   string `json:"infoText"`
+}
+
+// Store persists a single announcement in a key-value store.
+type Store struct {
+	store microstore.Store
```
aduffeck: We've been trying to avoid go-micro stores for a while as it's a dependency we might want to get rid of in the future. Could you rewrite this to do plain nats instead like we already do in a few places (see https://github.com/opencloud-eu/reva/blob/main/pkg/storage/cache/kv.go, for example) ?

aduffeck (changes requested):

services/web/pkg/announcement/service.go
dschmidt: Sure!

services/web/pkg/service/v0/service.go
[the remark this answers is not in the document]
dschmidt: The point of Copilot was that the description was outdated. In my first iteration I had an explicit DELETE call and then found it confusing what happens when text is empty vs explicit delete. So I decided I would rather treat PUT with empty values as "implicit DELETE". 
Since then I also added the enabled semantics, so the message can be shown or hidden without changing it. (i.e. we can prepare a message and enable it when its time has come)

services/web/pkg/announcement/announcement.go
dschmidt: Ah, fair enough. Certainly can

services/web/pkg/announcement/service.go
dschmidt: Done in 7d59a83c18: the handlers now log server-side errors before returning, tagged with the request id (via a small `logError` helper, same pattern as graph's auth middleware).

services/web/pkg/announcement/announcement.go
dschmidt: Done: reworked to a plain NATS JetStream KV bucket in 27bdda265f (no more go-micro store); 6d2d6c38ec then makes the connection fail fast like the graph service instead of retrying.

services/web/pkg/service/v0/service.go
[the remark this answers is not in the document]
dschmidt: The API is intentionally GET + PUT only: a PUT with an empty `bannerText` removes the record (implicit delete), and the `enabled` flag lets an announcement be hidden/shown without deleting it. There is no DELETE endpoint; the outdated PR description has been fixed.

services/web/pkg/service/v0/service.go
```diff
@@ -116,25 +136,53 @@ func (p Web) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
 	// ensure that the server url has a trailing slash
-	p.config.Web.Config.Server = strings.TrimRight(p.config.Web.Config.Server, "/") + "/"
+	webConfig.Server = strings.TrimRight(webConfig.Server, "/") + "/"
+
+	// inject the runtime managed announcement banner; keep a statically configured one
+	// (web.config.options.announcement) when nothing is stored
+	if a := p.currentAnnouncement(); a != nil {
+		webConfig.Options.Announcement = a
+	}
```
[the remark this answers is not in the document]
dschmidt: Resolved by removing static config entirely (d27b7ef40a): the runtime store is now the single source of truth for the announcement, so there is no static-vs-runtime precedence to reconcile.

services/web/pkg/server/http/server.go
```diff
@@ -76,11 +78,27 @@ func Server(opts ...Option) (http.Service, error) {
...
+		AuthPassword:         options.Config.Store.AuthPassword,
+		TLSEnabled:           options.Config.Store.EnableTLS,
+		TLSInsecure:          options.Config.Store.TLSInsecure,
+		TLSRootCACertificate: options.Config.Store.TLSRootCACertificate,
+	}, &options.Logger.Logger)
+	if err != nil {
+		return http.Service{}, fmt.Errorf("could not initialize announcement store: %w", err)
+	}
+	announcementStore := announcement.NewStore(kv)
+
```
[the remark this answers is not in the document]
dschmidt: Addressed in 6d2d6c38ec: replaced `cache.NewNatsKeyValue` (backoff/hang) with a direct `nats.Connect` + jetstream setup like the graph service, so an unreachable or misconfigured store fails fast with a clear startup error. We keep the store as a hard dependency rather than degrading to static config, since a statically configured announcement is not supported (d27b7ef40a).

aduffeck (approved):
lgtm 😍

Issue #3182, opened 2026-07-28
Implement subtree scanning in "posixfs scan"


aduffeck:
The `posixfs scan` command currently scans the full storage tree without an option to limit the scan to a particular subtree.
We should implement an according option to make use cases, e.g. where only one particular space was changed, more efficient.

aduffeck:
This has been implemented with #3235

PR #3191, opened 2026-07-29
feat(posixfs): #3182 add basepath option in the "posixfs scan" command

pbleser-oc:
## Description
Adds support for specifying a basepath using `-p` when running the `posixfs scan` command, to indicate a directory under which to start scanning, or a singular file to scan, as opposed to scanning from the storage root directory as is the default behaviour

## Related Issue

- implements https://github.com/opencloud-eu/opencloud/issues/3182

## Motivation and Context
In some scenarios, a system administrator might want to run the `posixfs scan` command under a specific subdirectory of the storage root instead of the whole tree under the storage root. This is made possible by specifying the basepath introduced with this change.

## How Has This Been Tested?

Tested manually in a local environment, by disabling posixfs watching (with `STORAGE_USERS_POSIX_WATCH_FS` set to `false`) and then copying files into the storage, checking that they had no xattrs set on them (using `getfattr -d`), running the `posixfs scan` command on them, both on individual files and on the containing directory, and then checking that OpenCloud posixfs metadata has been set on those files afterwards (again using `getfattr -d`).

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/pkg/x/path/filepathx/path.go
+++ b/pkg/x/path/filepathx/path.go
@@ -1,7 +1,9 @@
 package filepathx
 
 import (
+	"fmt"
 	"path/filepath"
+	"strings"
 )
 
 // JailJoin joins any number of path elements into a single path,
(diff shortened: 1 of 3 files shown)


pkg/x/path/filepathx/path.go
```diff
@@ -10,3 +12,30 @@ import (
...
+
+// Determines whether the file or directory 'child' is underneath the directory 'parent'.
+//
+// Note that 'parent' is expected to be a directory.
+func IsSameOrContainedBy(parent string, child string) (bool, error) {
+	absParent, err := filepath.Abs(parent)
+	if err != nil {
+		return false, fmt.Errorf("failed to make parent directory absolute: %q: %w", parent, err)
+	}
+	absParent = filepath.Clean(absParent)
```
aduffeck: `filepath.Abs` actually calls `Clean` on the result before returning it, so this line is redundant and can be removed.

pkg/x/path/filepathx/path.go
```diff
@@ -10,3 +12,30 @@ import (
...
+
+	absChild, err := filepath.Abs(child)
+	if err != nil {
+		return false, fmt.Errorf("failed to make child file/directory absolute: %q: %w", child, err)
+	}
+	absChild = filepath.Clean(absChild)
```
aduffeck: same here.

pkg/x/path/filepathx/path.go
```diff
@@ -10,3 +12,30 @@ import (
...
+
+	rel, err := filepath.Rel(absParent, absChild)
+	if err != nil {
+		return false, fmt.Errorf("failed to determine the relative path between the parent directory %q and the child file/directory: %q: %w", absParent, absChild, err)
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+		return false, nil
+	} else {
+		return true, nil
+	}
```
aduffeck: maybe simplify this to

```suggestion
	}
	return true, nil
```

aduffeck (changes requested):
lgtm besides the nitpicks.

pkg/x/path/filepathx/path.go
pbleser-oc: Thanks, made the changes, they were trivial and not worth keeping in the history, so I recommitted and force-pushed.

aduffeck (approved):
nice, lgtm

PR #3197, opened 2026-07-29
feat(search): check the index schema on startup and refuse breaking changes

dschmidt:
Stacked on #3345 (base `refactor/search-mapping` is #3345's branch, so this PR shows only its own commits; retarget to `main` once #3345 merges, both ship together). Refs #3092.

Two mechanisms keep the index schema and the code in sync.

**Versioned indices (from #3345).** The index name carries `search.SchemaVersion` (`opencloud-resource-v<N>`, `bleve-v<N>`). An intended breaking change bumps the constant: the service targets a fresh, empty index and leaves the old one in place. No migration or index-to-index reindex; content lives only in the index and the files are the source of truth, so `opencloud search index --all-spaces --force-rescan` rebuilds everything.

**Startup schema check (this PR).** The bump is manual, so this is the safety net for a schema change made without one. On startup both engines diff the existing index schema against the schema generated from code, via one shared recursive classifier:

- **equal**: start normally.
- **additive** (new fields, no indexed data): applied in place, no bump. OpenSearch `PUT _mapping`; bleve persists the code mapping (`SetInternal` + reopen). A startup warning lists the new fields (documents from before the upgrade lack them until re-indexed).
- **breaking** (changed types/analyzers, removed/renamed fields, new fields that already hold data): refuse to start. Versioning means a released instance never reaches this; it fires in development when the mapping changes breakingly without a bump, so the error is developer-facing (bump `SchemaVersion` or revert) and lists the diff.

Why this shape:
- The classifier is the only oracle; `PUT _mapping` is just the apply step (its merge semantics hide removals/renames), so it must not judge.
- bleve keeps no schema trace for dynamically indexed fields, so the classifier also checks `idx.Fields()`: a now-explicit field that already holds dynamic data of unknown shape is breaking (no search beats wrong search).
- No migration framework or blue-green: versioning gives the fresh index, a rescan repopulates it.

Known and accepted:
- Non-JSON 5xx proxy bodies lose their status code in opensearch-go error parsing (irrelevant while every non-conflict error is fatal).
- Trashed items are absent after a rebuild until restored (the rescan walks only the live tree; docs caveat for #3092).
- The bleve datapath must not be shared between processes; a second opener now fails after 5s (`bolt_timeout`) instead of hanging.

Deliberate consequences (not bugs):
- Additive upgrades are one-way within a version: once a release widened the schema, the previous release sees a stored-only field and refuses to start. Roll forward or rebuild; refusing both directions keeps mismatched schemas from serving queries.
- A new custom analyzer classifies as breaking even if only a new field uses it (OpenSearch can't change analysis on an open index, and both engines must match).
- Removing/renaming a field is breaking by design; tolerating removals would make renames undetectable.


butonic:
🤔 so how do we migrate in k8s? we can update all pods. the old ones will answer from the old index, the new ones will automagically create a new _v3 index, which is empty. a change in a space will trigger a reindex of that space. Or the admin executes a reindex cmd manually ... so forward update.

search requests from new pods will go to the new index, from old pods to the old ... that is at least consistent.

dschmidt:
> 🤔 so how do we migrate in k8s? we can update all pods. the old ones will answer from the old index, the new ones will automagically create a new _v3 index, which is empty. a change in a space will trigger a reindex of that space. Or the admin executes a reindex cmd manually ... so forward update.
> 
> search requests from new pods will go to the new index, from old pods to the old ... that is at least consistent.

Yes, you are right about all of that - BUT that's an already existing issue. Completely independent from my first refactor and also this one. This PR is just a mechanism to detect divergence and add fields to the index

A real waterproof approach would probably take care of migration, possibly with event replay for different versions, I don't know. I agree it would be nice to have, but totally out of scope right now

butonic (approved):

fschade (approved):
LGTM

PR #3207, opened 2026-07-30
Reindex spaces concurrently

aduffeck:
This speeds up reindexing all spaces which happend sequentially space after space until now. The level of concurrency can be configured using the SEARCH_REINDEX_CONCURRENCY env var (3 by default).

### DIFF
--- a/services/search/pkg/config/config.go
+++ b/services/search/pkg/config/config.go
@@ -28,7 +28,7 @@ type Config struct {
 	Extractor                  Extractor             `yaml:"extractor"`
 	ContentExtractionSizeLimit uint64                `yaml:"content_extraction_size_limit" env:"SEARCH_CONTENT_EXTRACTION_SIZE_LIMIT" desc:"Maximum file size in bytes that is allowed for content extraction." introductionVersion:"1.0.0"`
 	BatchSize                  int                   `yaml:"batch_size" env:"SEARCH_BATCH_SIZE" desc:"The number of documents to process in a single batch. Defaults to 500." introductionVersion:"1.0.0"`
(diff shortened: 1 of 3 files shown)


dschmidt (commented):
Two findings:

1) Should we just log the indexing errors in the server and otherwise swallow them?
I would probably prefer to collect them and report to the caller (the cli) that indexing (some) spaces failed.. Reporting the actual spaces in a structured format would probably mean touching the proto files. If you want to avoid that, a simple summary of failed spaces in a single error message would still be better than nothing imho.

2) It might be more complex and maybe not necessary right now, but I could imagine that an operator might want to set the concurrency value depending on the load of the cluster without restarting the search service ... 
Maybe let's name the option `_MAX_CONCURRENCY` so we can introduce a cli arg later?
This would obviously need a proto change as well

aduffeck:
> Two findings:
> 
> 1. Should we just log the indexing errors in the server and otherwise swallow them?
>    I would probably prefer to collect them and report to the caller (the cli) that indexing (some) spaces failed.. Reporting the actual spaces in a structured format would probably mean touching the proto files. If you want to avoid that, a simple summary of failed spaces in a single error message would still be better than nothing imho.

Yes you're right, the experience when using the CLI needs to be improved. We also still have the problem that the CLI doesn't wait indefinitely until the reindex process has completed, so returning an error message usually wouldn't reach the user (See https://github.com/opencloud-eu/opencloud/issues/2592). I would defer that to a proper fix alongside #2592 and just go with the server log messages for now.

> 2. It might be more complex and maybe not necessary right now, but I could imagine that an operator might want to set the concurrency value depending on the load of the cluster without restarting the search service ...
>    Maybe let's name the option `_MAX_CONCURRENCY` so we can introduce a cli arg later?
>    This would obviously need a proto change as well

Good point, I'll rename the option as suggested.

dschmidt:
Sounds reasonable, thanks!

dschmidt (approved):
LGTM

PR #3208, opened 2026-07-30
fix(thumbnails): pick audio cover art deterministically

dschmidt:
# Superseded by #3332 (if we decide to go down that road)

Stacked on #3198.

The thumbnails `AudioDecoder` used `dhowden/tag`'s single `Picture()`, which is non-deterministic when an audio file embeds multiple pictures (e.g. front + back cover). This switches to my fork (`dschmidt/tag`, adds `Pictures()` + a `RawType` byte) and selects deterministically: the tagged front cover (picture type 0x03) if present, otherwise the first available picture. Many taggers store the cover as type "Other" (0x00) rather than front, so the fallback is the common case.

Fork PR: https://github.com/dschmidt/tag/pull/1 (merged to master).


dschmidt:
Superseded by the Tika-based thumbnail pipeline (#3332 + upstream TIKA-4801, which makes Tika emit the front cover first): with covers coming from Tika there is no dhowden/tag code path left to make deterministic. Branch stays around in case we revive it.

PR #3210, opened 2026-07-30
feat(search): honest has-preview prop and thumbnails relationship

dschmidt:
Stacked on #3198. Depends on the audio cover art support from apache/tika#2983 (TIKA-4801) for the audio preview signal.

Makes preview availability honest instead of a blanket mimetype match, and adds the driveItem thumbnails relationship.

- Split `SupportedMimeTypes` into `Unconditional`/`Embedded` preview types + `HasPreview` as the single source of truth (unconditional types follow from the mimetype, embedded types like audio cover art depend on an indexed preview).
- Index embedded cover-art dimensions as `oc.preview` (preferring the front cover, matching the thumbnailer). Also fixes facets leaking from embedded resources into the container (audio no longer gets a bogus image facet / wiped audio facet).
- Honest `oc:has-preview` for audio in search results (new `preview` field on the search message entity).
- Graph driveItem listings expose the `thumbnails` relationship on `$expand=thumbnails`: small/medium/large thumbnail URLs plus a `source` thumbnail carrying the original dimensions. Per-size dimensions are omitted because the served size depends on the configured resolution set, not knowable at listing time.

Follow-ups (kept out to stay focused): consolidate the shared-view thumbnail blocks via a generic `odata` expand-set, and switch reva propfind has-preview to `thumbnail.HasPreview`.

> [!NOTE]
> Even once this PR **and** #3211 (graph search query endpoint) are both merged, hits from `POST /graph/v1beta1/search/query` still won't carry thumbnails — the `$expand=thumbnails` wiring for search hits builds on both PRs and currently only exists on the consolidation branch ([`dschmidt/opencloud@feat/graph-search-full-2`](https://github.com/dschmidt/opencloud/tree/feat/graph-search-full-2), commit [`0d94199a`](https://github.com/dschmidt/opencloud/commit/0d94199a): `searchEntityThumbnailSet` in `searchquery.go`, preview presence via `thumbnail.HasPreviewForMimeType` + indexed `oc.preview`/image-facet dimensions). It needs a follow-up PR once both are in.


services/thumbnails/pkg/thumbnail/has_preview.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/opencloud-eu/opencloud/pkg/conversions"
+)
+
+// Arbitrary-metadata keys for an embedded preview's dimensions, written at index
+// time; their presence signals a preview exists.
+const (
+	PreviewWidthKey  = "oc.preview.width"
+	PreviewHeightKey = "oc.preview.height"
+)
```
butonic, line 14: I would prefer a single `oc.preview` property with `200x300` as the dimension. While we do cache the properties in nats, I think splitting by `x` is more efficient.

butonic:
Hm ...
>  I need the availability and dimensions of embedded previews at index time (#3210).

🤔 

You mean you need them to be created when indexing (after postprocessing), so they are available during dirlistings?

dschmidt:
> > I need the availability and dimensions of embedded previews at index time (#3210).

> You mean you need them to be created when indexing (after postprocessing), so they are available during dirlistings?

Basically, yes - I need the dimensions in ArbitraryMetadata during listings and in the index during search, so I can fill the thumbnails relationship.
It's basically the same as all the other extracted metadata like audio artists, geo data in images, etc. 
This was the original motivation to implement the extraction in Tika, so it's completely anologous to the other metadata extraction, because that seemed like a completely natural fit.
The only downside of doing it like this is, that previews for files out of the `EmbeddedPreviewMimeTypes` group  become available only when they have been indexed.

services/thumbnails/pkg/thumbnail/has_preview.go
dschmidt: I'm not the biggest fan tbh.

Would need concatenation on the indexing site and parsing on the consuming side, while making querying for it more awkward and require special handling.
I could imagine for my music server it could be nice for some views to query only files with specific (minimal)  preview sizes e.g.  

If you insist, we can do it.. but I'd prefer not to

PR #3220, opened 2026-08-03
Extend posixfs consistency check

aduffeck:
Extend the `posixfs consistency` command to also check the blobsize and checksums attributes.

It's also possible to limit the check to one or more files or folders.

```
Check the consistency of the posixfs storage.

You can specify one or more paths to limit the scope of the check.
If no path is provided, the whole storage is checked.

The provided arguments determines the scope of the check:
  - a storage root:   the whole storage (all personal and project spaces) is checked
  - a space root:     only that space is checked
  - a file or folder: only that single entity is checked (and its children, if it is a folder)

Usage:
  opencloud posixfs consistency [path ...] [flags]

Flags:
      --fix-checksums   Recalculate and fix the file checksums. This reads every file and can be slow on large storages.
  -h, --help            help for consistency
```

### DIFF
--- a/opencloud/pkg/command/posixfs.go
+++ b/opencloud/pkg/command/posixfs.go
@@ -1,14 +1,9 @@
 package command
 
 import (
-	"bufio"
-	"bytes"
-	"context"
 	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/opencloud-eu/opencloud/opencloud/pkg/register"
@@ -16,7 +11,6 @@ import (
 	"github.com/opencloud-eu/opencloud/pkg/config/configlog"
 	"github.com/opencloud-eu/opencloud/pkg/config/parser"
 	"github.com/opencloud-eu/opencloud/pkg/x/path/filepathx"
-	storageUsersConfig "github.com/opencloud-eu/opencloud/services/storage-users/pkg/config"
 	storageUsersParser "github.com/opencloud-eu/opencloud/services/storage-users/pkg/config/parser"
 	"github.com/opencloud-eu/opencloud/services/storage-users/pkg/event"
 	"github.com/opencloud-eu/opencloud/services/storage-users/pkg/revaconfig"
@@ -25,17 +19,9 @@ import (
 	"github.com/opencloud-eu/reva/v2/pkg/storage/fs/posix/options"
 	"github.com/opencloud-eu/reva/v2/pkg/storage/fs/registry"
 	"github.com/opencloud-eu/reva/v2/pkg/storage/pkg/decomposedfs/metadata/prefixes"
-	"github.com/opencloud-eu/reva/v2/pkg/storage/pkg/decomposedfs/node"
 
 	"github.com/pkg/xattr"
 	"github.com/spf13/cobra"
-	"github.com/vmihailenco/msgpack/v5"
-)
-
-var (
-	restartRequired      = false
-	recalculateChecksums = false
-	ignorer              *ignore.Ignorer
 )
 
 type IDCacher interface {
@@ -205,7 +191,27 @@ The provided arguments determines the scope of the check:
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cfg := ocCfg.StorageUsers
-			return checkPosixfsConsistency(cfg, cmd, args)
+			if len(args) == 0 {
+				args = []string{cfg.Drivers.Posix.Root}
+			}
+			log := logger("posixfs")
+			recalculateChecksums, _ := cmd.Flags().GetBool("fix-checksums")
+
+			drivers := revaconfig.StorageProviderDrivers(cfg)
+			drivers["posix"] = revaconfig.Posix(cfg, false, false)
+			opts, err := options.New(drivers["posix"].(map[string]any))
+			if err != nil {
+				return err
+			}
+			ignorer := ignore.NewIgnorer(opts, &log)
+
+			checker := &consistencyChecker{
+				cfg:                  cfg,
+				ignorer:              ignorer,
+				recalculateChecksums: recalculateChecksums,
+			}
+
+			return checker.Check(args)
 		},
 	}
 	consCmd.Flags().Bool("fix-checksums", false, "Recalculate and fix the file checksums. This reads every file and can be slow on large storages.")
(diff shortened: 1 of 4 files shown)


opencloud/pkg/command/posixfs.go
```diff
@@ -188,75 +178,130 @@ func scanCmd(ocCfg *config.Config) *cobra.Command {
...
+		Short: "Check the consistency of the posixfs storage",
+		Long: `Check the consistency of the posixfs storage.
+
+You can specify one or more paths to limit the scope of the check.
+If no path is provided, the whole storage is checked.
+
+The provided arguments determines the scope of the check:
+  - a storage root:   the whole storage (all personal and project spaces) is checked
+  - a space root:     only that space is checked
+  - a file or folder: only that single entity is checked (and its children, if it is a folder)`,
```
pbleser-oc: Please use "directory" instead of "folder", as we discussed in chat, since we are literally referring to a filesystem directory, and "directory" is UNIX/Linux parlance, while "folder" is rather part of UI and Windows vocabulary.

opencloud/pkg/command/posixfs.go
```diff
@@ -2,46 +2,40 @@ package command
...
-	nameAttrName     = "user.oc.name"
-	spaceIDAttrName  = "user.oc.space.id"
-	ownerIDAttrName  = "user.oc.owner.id"
-)
-
 var (
-	spinner         *yacspin.Spinner
-	restartRequired = false
-	ignorer         *ignore.Ignorer
+	restartRequired      = false
```
pbleser-oc: Obvious, but returning a `bool` just like the `int` for counting the fixes (or a "report" struct that contains both) instead of using a global variable would make it possible to implement unit tests for the functions below (e.g. `checkNodeAttributes`).

pbleser-oc (changes requested):
lgtm, even though I only have limited ability to verify whether it's functioning properly.

Maybe have a look at not using global variables for enhancing the "unit-testability" of the implementation, up to you :)

aduffeck:
> lgtm, even though I only have limited ability to verify whether it's functioning properly.
> 
> Maybe have a look at not using global variables for enhancing the "unit-testability" of the implementation, up to you :)

:+1: I refactored the consistency check into a struct which keeps the state instead. Test coverage can be increased in a follow-up PR.

opencloud/pkg/command/posixfs_consistency.go
```diff
@@ -0,0 +1,492 @@
...
+	if c.restartRequired {
+		fmt.Println("\n\n  ⚠️  Please restart your openCloud instance to apply changes.")
+	}
+	return nil
+}
+
+func (c *consistencyChecker) checkSpaces(basePath string) {
+	dirEntries, err := os.ReadDir(basePath)
+	if err != nil {
+		logFailure("Error reading spaces directory '%s': %v", basePath, err)
```
pbleser-oc, line 76: How about counting and/or collecting errors instead of just logging and then skipping?
I wonder whether it wouldn't be safer if we either abort the execution altogether, or, instead, log but also count errors and include them in the report at the end, along side with the number of fixed files.
And where we abort the function, maybe return the error instead (and count them in the calling site) ?

opencloud/pkg/command/posixfs_consistency.go
```diff
@@ -0,0 +1,492 @@
...
+// updateOwnerIndexFile handles the logic of reading, modifying, and writing the MessagePack index file.
+func updateOwnerIndexFile(basePath string, obsoleteIDs []string) error {
+	fmt.Printf("  Rewriting index file '%s'\n", basePath)
+
+	ownerID, err := xattr.Get(basePath, prefixes.OwnerIDAttr)
+	if err != nil {
+		return fmt.Errorf("could not get owner ID from oldest entry '%s' to find index: %w", basePath, err)
+	}
+
+	indexPath := filepath.Join(basePath, "../../indexes/by-user-id", string(ownerID)+".mpk")
```
pbleser-oc: `filepath.Join(basePath, "..", "..", "indexes", "by-user-id")` instead for portability?

Do we need to check this one to ensure it doesn't escape the storage root "jail" ?
If not, adding a comment to explain why this is safe might be a good idea.

opencloud/pkg/command/posixfs_consistency.go
```diff
@@ -0,0 +1,492 @@
...
+	"github.com/pkg/xattr"
+	"github.com/shamaton/msgpack/v2"
+)
+
+type consistencyChecker struct {
+	cfg                  *storageUsersConfig.Config
+	ignorer              *ignore.Ignorer
+	recalculateChecksums bool
+
+	restartRequired bool
```
pbleser-oc, line 31: Note that a downside of using the struct to hold state is that we can't parallelize the execution in the future, or we'd have to use atomic integers and booleans to update the inner state, whereas returning and aggregating those would make that safe and transparent (but with the downside of having to return that thing everywhere, and recurse).

opencloud/pkg/command/posixfs_consistency.go
```diff
@@ -0,0 +1,492 @@
...
+			fmt.Printf("    - %s\n", id)
+		}
+
+		fmt.Printf("\n  ⏳ Oldest entry is '%s' (modified on %s).\n",
+			filepath.Base(oldestEntry.Path), oldestEntry.ModTime.Format(time.RFC1123))
+
+		targetID := oldestEntry.ParentID
+		fmt.Printf("  ✅ Proposed target Parent ID: %s\n", targetID)
+
+		fmt.Printf("\n  Do you want to unify all parent IDs to '%s'? This will modify %d entries, the directory, and the user index. (y/N): ", targetID, len(entries))
```
pbleser-oc, line 132: This screams for having a `--yes` and a `--no` option in the future to make it non-interactive :)

opencloud/pkg/command/posixfs_consistency.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+
+	fixes := c.walkNodes(spacePath, string(rootID))
+
+	if fixes > 0 {
+		fmt.Printf("  ✓ Fixed %d incorrect node attributes in %s\n", fixes, filepath.Base(spacePath))
+	}
+}
+
+func fixSpaceID(spacePath string, obsoleteIDs []string, targetID string, entries []EntryInfo) {
```
pbleser-oc: Would you mind a high-ish-level comment here to explain what this one does (like with the other functions above) ?

pbleser-oc (changes requested):
More comments after the updated implementation.
If time doesn't permit, maybe just add a couple of comments so we can improve/address them in the future if necessary.

opencloud/pkg/command/posixfs_consistency.go
aduffeck: done

opencloud/pkg/command/posixfs_consistency.go
aduffeck: We're now inferring the indexes path from the config instead.

opencloud/pkg/command/posixfs_consistency.go
aduffeck: The consistency check can take a substantial amount of time on large storages, so I would argue that it's better to keep running and log the errors so that the work that can be done gets done, e.g. when the command is running over night. The admin can then check the errors and run the command for these paths again if needed.
Counting and reporting at the end sounds like a good idea though. Defered to #3227

opencloud/pkg/command/posixfs_consistency.go
aduffeck: Deferred to #3227

opencloud/pkg/command/posixfs_consistency.go
aduffeck: Deferred to #3227

aduffeck:
I fixed the comment and the base path boundary issue and created #3227 for the rest. I think an issue is a better place for these things as code comments tend to get forgotten and rot.

pbleser-oc (approved):
Looks good, thanks for the changes.

PR #3228, opened 2026-08-03
feat(posixfs): #3182 add basepath option in the "posixfs scan" command

pbleser-oc:
## Description
* add support for specifying a set of resources when running the posixfs scan command scanning, or a singular file to scan, as opposed
* refactors the resource walking function implemented in https://github.com/opencloud-eu/opencloud/pull/3220 in order to reuse it for the posixfs scan command

## Related Issue
* implements https://github.com/opencloud-eu/opencloud/issues/3182
* documentation: https://github.com/opencloud-eu/docs/pull/1084

## Motivation and Context
Addresses use-cases where files are copied into the PosixFS storage directory tree while the server is not running or with collaborative mode disabled, e.g. during large migrations or in disaster recovery scenarios.

With these changes, it is possible to perform the storage scanning in a much more granular way, which may help in cases where the storage tree is especially large and scanning everything would be too slow.

## How Has This Been Tested?
Has been tested locally/manually by
* running the server with collaborative mode disabled (`STORAGE_USERS_POSIX_WATCH_FS=false`)
* copying files into a space storage directory, `cp ... ~/.opencloud/storage/users/projects/9c4feaa1-0a8f-4f9d-bcf7-2d79b714d9d5/`
* checking that no xattrs were set on those files using `getfattr -d`
* running the `posixfs scan` command with different use-cases:
  * on the storage root
  * without arguments
  * on a subdirectory of a space
  * on the space directory
  * on individual files
* and then checking that the xattrs were set (`user.oc.id`, `user.oc.parentid`, etc...), again using `getfattr -d`

Unfortunately, it is not possible to add unit tests as it would require a lot of changes to abstract away the filesystem operations, or a significant amount of test harness setup to be able to run the command with a NATS Jetstream and everything else that is needed for it.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


opencloud/pkg/command/posixfs.go
```diff
@@ -56,10 +56,22 @@ func init() {
...
-		PreRunE: func(cmd *cobra.Command, args []string) error {
+		Long: `Perform a filesystem scan and update the ID and filemetadata cache.
+
+You can specify one or more paths to limit the scope of the scan.
+If no path is provided, the whole storage is checked, starting at the storage root folder.
+
+The provided arguments determines the scope of the check:
+  - a storage root:   the whole storage (all personal and project spaces) is scanned
+  - a space root:     only that space is scanned
+  - a file or folder: only that single resource is scanned (and its children, if it is a folder)
```
aduffeck: We should use "directory" here and below as well as discussed ;)

opencloud/pkg/command/posixfs.go
```diff
@@ -77,80 +89,79 @@ func scanCmd(ocCfg *config.Config) *cobra.Command {
...
+				if !ok {
+					fmt.Fprintf(os.Stderr, "The posix driver does not expose WarmupIDCache.\n")
+					os.Exit(1)
+				}
 
-			cacher, ok := fs.(IDCacher)
-			if !ok {
-				fmt.Fprintf(os.Stderr, "The posix driver does not expose WarmupIDCache.\n")
-				os.Exit(1)
+				scan = func(path string) error { return cacher.WarmupIDCache(path, true, false) }
```
aduffeck: If I understand the code correctly you deliberately overwrite the `scan` var so that the initialization happens only once? I think it would be more clear if you'd use `sync.Once*` for that, or in this case I think I would just do it directly here before iterating over the paths.
The initialization isn't that expensive and there's no deliberate code path where it wouldn't be needed anyway.

opencloud/pkg/command/posixfs.go
```diff
@@ -254,3 +264,39 @@ func isSpaceRoot(path string) bool {
...
+func processPosixFsResources(paths []string,
+	personalSpaceDir func(string) error,
+	projectSpaceDir func(string) error,
+	spaceRoot func(string) error,
+	entity func(string) error,
+) error {
+	for _, path := range paths {
+		rootPath, err := findStorageRoot(path)
+		if err != nil {
+			return err
```
aduffeck: Could we change this function to log the errors and continue instead? The commands using this method can take a very long time to finish, so people are likely to let it run while they are away from the keyboard. I think we should try to get as much of the work done as possible and not bail out on any error - which might be intermittent or specific to one resource.

aduffeck (changes requested):

opencloud/pkg/command/posixfs.go
pbleser-oc: We don't need to, since the functions that are passed as parameters can simply choose to log and return `nil`, or the caller of the `processPosixFsResources` function can also choose to ignore the `error` that is returned, log, and continue.

Which is how I ported your previous implementation in `posixfs_consistency.go`

pbleser-oc: Ah, but you probably mean that the same should be done for the `scan` command, where I currently bail when an error is encountered for one `path`.
I cannot influence what the `WarmupIDCache` function does without changing that one deep in Reva, but I can certainly change it to continue processing the subsequent paths when an error occurs, rather than quitting altogether.

aduffeck: Right, my worry is that a number of spaces are given, one of them can't be `stat`ed or is outside of the storage root and that causes the subsequents spaces to be skipped, for example.

aduffeck: So it's actually not only about the errors, but we also can't `return` in the space, `contained` or default case becasue that would mean that paths are skipped.

opencloud/pkg/command/posixfs.go
pbleser-oc: I implement it that way to make sure I don't inadvertently use any of the variables that are used to initialize the scan function, since I don't need them below, and I don't want them either.

Just a habit I have of scoping things to avoid accidental references.

Let me know if this is too un-idiomatic compared to what we do everywhere else and I'll remove it, if it is confusing.

aduffeck: if it's just about scoping variables you can use blocks, e.g. 
```
{
  a := 1
}
fmt.Prinln(a) <-- fails, doesn't see `a`
```

but yeah, maybe it's just me but I found it a bit confusing at first, esp. the overwriting of the var inside the func which was assigned to the var initially.

pbleser-oc: I see where the confusion comes from, I added a `= nil` assignment there with 732d4028f2da50e46296f85950040afcfc649159 to make it clear that the block that follows is not the implementation of the func

opencloud/pkg/command/posixfs.go
pbleser-oc: Yes, got it now, thanks, good point

opencloud/pkg/command/posixfs.go
aduffeck: ah! yes indeed, my brain added a `=` there which wasn't there... Thanks.

aduffeck (approved):

PR #3252, opened 2026-08-05
Fix missing favorite flag on opensearch hits

aduffeck:
Fix favorites Flag in search response.


rhafer (approved):

dschmidt (approved):
LGTM.

During review I stumbled upon etags missing in the search results as well. Maybe worth filing an issue to not forget about it...

PR #3257, opened 2026-08-06
fix: notification handling for share removal and space membership expiry 

maki5:
### Notifications service

- Handle the `ShareRemoved` event (backport of owncloud/ocis#10915):
  new email template, event dispatch and grouped-mail support, and the
  mail option is enabled by default in the notification settings
- Rework `handleSpaceMembershipExpired`: resolve the space name via a
  prepare helper and send from the default email sender, dropping the
  dependency on the unresolvable space owner lookup

### Userlog service 

- Add a dedicated conversion path for `SpaceMembershipExpired` that
  resolves only the space (the template does not reference a user),
  so the in-app notification no longer fails on the unresolvable space
  owner placeholder
- Handle policy violation and antivirus postprocessing steps for
  granular in-app notifications (previously we didn't handle it's disabling)

### Dependencies

- Bump reva to v2.48.1-0.20260812174127-12b756ed83c5 (PR
  opencloud-eu/reva#766), bringing:
  - Backport of owncloud/ocis#11079: emit a `SpaceMembershipExpired`
    event when the decomposedfs cleans up expired space grants "on
    access"
  - Sharing manager fix: publish `SpaceMembershipExpired` instead of
    `ShareExpired` for expired space-root shares

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -73,8 +73,7 @@
 
         // service accounts
         "OC_SERVICE_ACCOUNT_ID": "service-account-id",
-        "OC_SERVICE_ACCOUNT_SECRET": "service-account-secret",
-        "FRONTEND_CONFIGURABLE_NOTIFICATIONS": "true"
+        "OC_SERVICE_ACCOUNT_SECRET": "service-account-secret"
       }
     },
     {
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -73,7 +73,8 @@
 
         // service accounts
         "OC_SERVICE_ACCOUNT_ID": "service-account-id",
-        "OC_SERVICE_ACCOUNT_SECRET": "service-account-secret"
+        "OC_SERVICE_ACCOUNT_SECRET": "service-account-secret",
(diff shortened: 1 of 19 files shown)


micbar:
Please add a description.

micbar (approved):
Looking good to me 👍

micbar:
the last merge from main somehow messed this up a bit.

@butonic @pbleser-oc Can you help to rebase this properly?

@maki5 I tested you changes, looking good to me! 👍 

We should enable this feature by default, i checked with @tbsbdr 

That means we need to change the default config by setting this StructValue https://github.com/opencloud-eu/opencloud/blob/main/services/frontend/pkg/config/config.go#L62 in the defaultconfig.go to `true`

butonic:
I rebased this to get rid of the merge conflict.

butonic (approved):

maki5:
@butonic can you please reviews this one as well? https://github.com/opencloud-eu/reva/pull/766

it's a dependency without this PR cannot be merged, so we should merge it first

PR #3263, opened 2026-08-07
api-test: add posixfs scan and consistency CLI tests

v-scharf:
test for: 
```
/var/lib/opencloud # opencloud posixfs scan --help
Perform a filesystem scan and update the ID and filemetadata cache.

You can specify one or more paths to limit the scope of the scan.
If no path is provided, the whole storage is checked, starting at the storage root directory.

The provided arguments determines the scope of the check:
  - a storage root:   the whole storage (all personal and project spaces) is scanned
  - a space root:     only that space is scanned
  - a file or directory: only that single resource is scanned (and its children, if it is a directory)

Any specified file or directory must be underneath the storage root directory and if that is not the case,
the command is aborted with an error before performing any scanning.

Usage:
  opencloud posixfs scan [path ...] [flags]

Flags:
  -E, --halt-on-error   Halt at once when an error occurs when processing one of the paths (default behaviour is to keep going and attempt to process all paths).
  -h, --help            help for scan
```

and 

```
/var/lib/opencloud # opencloud posixfs consistency --help
Check the consistency of the posixfs storage.

You can specify one or more paths to limit the scope of the check.
If no path is provided, the whole storage is checked.

The provided arguments determines the scope of the check:
  - a storage root:   the whole storage (all personal and project spaces) is checked
  - a space root:     only that space is checked
  - a file or directory: only that single entity is checked (and its children, if it is a directory)

Usage:
  opencloud posixfs consistency [path ...] [flags]

Flags:
      --fix-checksums   Recalculate and fix the file checksums. This reads every file and can be slow on large storages.
  -h, --help            help for consistency
```

### DIFF
--- a/tests/acceptance/features/cliCommands/posixfsConsistency.feature
+++ b/tests/acceptance/features/cliCommands/posixfsConsistency.feature
@@ -39,12 +39,12 @@ Feature: posixfs consistency CLI command
     Given user "Alice" has uploaded file with content "content" to "textfile.txt"
     When the administrator sets the extended attribute "user.oc.cs.sha1" of file "textfile.txt" of user "Alice" to "corrupted" on the POSIX filesystem
     Then the command should be successful
-    And the administrator checks the posixfs consistency using the CLI
-    And the command should be successful
+    When the administrator checks the posixfs consistency using the CLI
+    Then the command should be successful
     And the command output should not contain "Fixed checksum"
-    And the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
-    And the command should be successful
+    When the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
+    Then the command should be successful
     And the command output should contain "Fixed checksum"
-    And the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
-    And the command should be successful
+    When the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/cliCommands/posixfsConsistency.feature
```diff
@@ -0,0 +1,50 @@
...
+    Then the command should be successful
+    And the administrator checks the posixfs consistency using the CLI
+    And the command should be successful
+    And the command output should not contain "Fixed checksum"
+    And the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
+    And the command should be successful
+    And the command output should contain "Fixed checksum"
+    And the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
+    And the command should be successful
+    And the command output should not contain "Fixed checksum"
```
saw-jan:
```suggestion
    When the administrator checks the posixfs consistency using the CLI
    Then the command should be successful
    And the command output should not contain "Fixed checksum"
    When the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
    And the command should be successful
    And the command output should contain "Fixed checksum"
    Then the administrator checks the posixfs consistency using the CLI with flag "--fix-checksums"
    And the command should be successful
    And the command output should not contain "Fixed checksum"
```

saw-jan (approved):

PR #3277, opened 2026-08-10
Expose upload status and message

aduffeck:
Show the upload status and message in the `storage-users uploads sessions` command.

Depends on https://github.com/opencloud-eu/reva/pull/764


rhafer:
merged as part of #3329

PR #3241, opened 2026-08-04
Activitylog event handler split

butonic:
As part of https://github.com/opencloud-eu/opencloud/issues/1312 we can now disable the event handler or http handler part of the activitylog service.

Unfortunately, the code was very intertwined. The http handler was started as a go routine of the event handler. Cleaning that up took quite a bit of work.

PR #3279, opened 2026-08-10
ability to disable grpc and/or event consumer for event history service

maki5:
for event history service now it's possible to enable/disable from config

- grpc server
- events consumer


services/eventhistory/pkg/command/server.go
```diff
@@ -76,22 +90,26 @@ func Server(cfg *config.Config) *cobra.Command {
...
+				)
 
-			gr.Add(runner.NewGoMicroGrpcServerRunner(cfg.Service.Name+".grpc", service))
+				gr.Add(runner.NewGoMicroGrpcServerRunner(cfg.Service.Name+".grpc", service))
+			} else {
+				logger.Info().Msg("gRPC server disabled, not starting gRPC service")
+			}
 
-			{
+			if !cfg.Debug.Disabled {
```
butonic: Hm, while we could make the debug endpoint optional, I would leave that out for now. If we want to change how we expose metrics, allow readyniess checks etc. we should track that in a separate issue. It would totally make sense to reduce the number of debug servers we run, especially because we run them in a single binary anyway. However we would need to look into metrics collection and how we can expose the metrics of all services and the readyness / healthyness in a common debug server.

For now, do not make this configurable.

services/eventhistory/pkg/config/parser/parse.go
```diff
@@ -34,5 +34,12 @@ func ParseConfig(cfg *config.Config) error {
 
 // Validate validates the config
 func Validate(cfg *config.Config) error {
+	// this need to be discussed, I put it here for now because I belive we need to ensure that the service is doing
+	// at least something, if we have events and grpc disabled that means that the service is doing actually nothing
+	// which can be a siletlly ignorred and cause issues in the future
+	if cfg.Events.Disabled && cfg.GRPC.Disabled {
```
butonic, line 38: makes sense!

services/eventhistory/pkg/command/server.go
```diff
@@ -58,10 +59,23 @@ func Server(cfg *config.Config) *cobra.Command {
...
+					TLSInsecure:          cfg.Events.TLSInsecure,
+					TLSRootCACertificate: cfg.Events.TLSRootCACertificate,
+					EnableTLS:            cfg.Events.EnableTLS,
+					AuthUsername:         cfg.Events.AuthUsername,
+					AuthPassword:         cfg.Events.AuthPassword,
+				})
+				if err != nil {
+					return err
+				}
+			} else {
```
butonic: This only sets up the nats consumer. Unfortunately, the actual events.ConsumeAll(...) call still lives in the NewEventHistoryService. You will have te extract that into a separate module, like in https://github.com/opencloud-eu/opencloud/pull/3241/changes#diff-654db365a382a61398b04610d91188cc8b6d777398bd1b7a25793977e33231e0 ... although I still need to fix the failing tests it shows the different services that are started by the server command.

services/eventhistory/pkg/service/service.go
```diff
@@ -32,17 +32,17 @@ type EventHistoryService struct {
...
+	if consumer != nil {
+		ch, err := events.ConsumeAll(consumer, "evhistory")
+		if err != nil {
+			return nil, err
+		}
 
-	eh := &EventHistoryService{ch: ch, store: store, cfg: cfg, log: log}
-	go eh.StoreEvents()
+		eh.ch = ch
+		go eh.StoreEvents()
```
butonic: This consumer will never be started if grpc has been disabled

butonic (approved):

PR #3285, opened 2026-08-11
ability to disable http and/or event consumer for user log service

maki5:
for user log service it's possible now to disable/enable from config
- events listener(consumer)
- http server


services/userlog/pkg/config/config.go
```diff
@@ -30,6 +30,9 @@ type Config struct {
 
 	DisableSSE bool `yaml:"disable_sse" env:"OC_DISABLE_SSE,USERLOG_DISABLE_SSE" desc:"Disables server-sent events (sse). When disabled, clients will no longer receive sse notifications." introductionVersion:"1.0.0"`
 
+	EventsDisabled bool `yaml:"events_disabled" env:"USERLOG_EVENTS_DISABLED" desc:"Disables listening for events. Set this to true if the service should only handle HTTP requests." introductionVersion:"%NEXT%"`
+	HTTPDisabled   bool `yaml:"http_disabled" env:"USERLOG_HTTP_DISABLED" desc:"Disables the HTTP service. Set this to true if the service should only handle events." introductionVersion:"%NEXT%"`
+
```
butonic: There are dedicated structs for HTTP and Events, move the flags there.

services/userlog/pkg/command/server.go
```diff
@@ -143,6 +149,36 @@ func Server(cfg *config.Config) *cobra.Command {
 				}
 
 				gr.Add(runner.NewGoMicroHttpServerRunner(cfg.Service.Name+".http", server))
+			} else {
+				logger.Info().Msg("HTTP server disabled, not starting HTTP service")
+
+				if !cfg.EventsDisabled {
```
butonic: the logic here seems wrong. If `!cfg.HTTPDisabled` is true the else branch is not reached ... meaning the Events service cannot be enabled when the HTTPServer is enabled. Or am I misunderstanding the logic?

pkg/shared/errors.go
```diff
@@ -101,3 +101,11 @@ func MissingURLSigningSecret(service string) error {
 		"the config/corresponding environment variable).",
 		service, defaults.BaseConfigPath())
 }
+
+func AllComponentsDisabledError(service string) error {
```
pbleser-oc: I'd suggest naming that about "API handlers" instead of "components", since "components" is a super generic term that can mean anything, that will allow for a more specific error message as well (the error text below mentions "request handlers and event consumers", that's a lot more specific than "components")
```suggestion
func AllApiHandlersDisabledError(service string) error {
```

butonic (changes requested):
the pkg/service package now implements both:
1. service.go contains event processing that stores events in nats + publishes sse events
2. http.go implements the http handler that reads from nats.

In the [activitylog PR](https://github.com/opencloud-eu/opencloud/pull/3241/changes#diff-80f16457a97d68a3b298268edc44d66f5580ceaf876e3518c671e55c92a910a2) I extracted an activitylog service that wraps the nats kv store with the business logic to add and remove events. It is injected into the http service and the events service (but I like your name 'consumer' better). 

I know this is more work, but I think we should seperate the implementation aspects. AFAICT the http service then does not need the event stream at all, so we only need to initiate it for the consumer (hm, it also publishes sse events, so maybe events is a better name in this case).

maki5:
> the pkg/service package now implements both:
> 
> 1. service.go contains event processing that stores events in nats + publishes sse events
> 2. http.go implements the http handler that reads from nats.
> 
> In the [activitylog PR](https://github.com/opencloud-eu/opencloud/pull/3241/changes#diff-80f16457a97d68a3b298268edc44d66f5580ceaf876e3518c671e55c92a910a2) I extracted an activitylog service that wraps the nats kv store with the business logic to add and remove events. It is injected into the http service and the events service (but I like your name 'consumer' better).
> 
> I know this is more work, but I think we should seperate the implementation aspects. AFAICT the http service then does not need the event stream at all, so we only need to initiate it for the consumer (hm, it also publishes sse events, so maybe events is a better name in this case).

I refactored the code to follow the format from activitylog PR

butonic:
@maki5 can you rebase?

maki5:
> @maki5 can you rebase?

@butonic sure, rebased

services/userlog/pkg/command/server.go
```diff
@@ -120,20 +119,40 @@ func Server(cfg *config.Config) *cobra.Command {
...
+					handle,
+					httpSvc.Logger(logger),
+					httpSvc.Config(cfg),
+					httpSvc.GatewaySelector(gatewaySelector),
+					httpSvc.ValueClient(vClient),
+					httpSvc.RegisteredEvents(_registeredEvents),
+					httpSvc.TraceProvider(tracerProvider),
+				)
+				if err != nil {
+					logger.Info().Err(err).Str("transport", "http").Msg("Failed to initialize server")
```
pbleser-oc, line 145:
```suggestion
					logger.Error().Err(err).Str("transport", "http").Msg("Failed to initialize server")
```
Shouldn't this be `ERROR` instead of `INFO` ?

services/userlog/pkg/config/config.go
```diff
@@ -72,6 +73,7 @@ type CORS struct {
 
 // HTTP defines the available http configuration.
 type HTTP struct {
+	Disabled  bool                  `yaml:"disabled" env:"USERLOG_HTTP_DISABLED" desc:"Disables the HTTP service. Set this to true if the service should only handle events." introductionVersion:"%NEXT%"`
```
pbleser-oc, line 76:
```suggestion
	Disabled  bool                  `yaml:"disabled" env:"USERLOG_HTTP_DISABLED" desc:"Disables the HTTP service. Set this to true if the service should only handle events." introductionVersion:"%%NEXT%%"`
```
We should use `%%NEXT%%`, not `%NEXT%`

services/userlog/pkg/config/config.go
```diff
@@ -53,6 +53,7 @@ type Persistence struct {
 
 // Events combines the configuration options for the event bus.
 type Events struct {
+	Disabled             bool   `yaml:"disabled" env:"USERLOG_EVENTS_DISABLED" desc:"Disables listening for events. Set this to true if the service should only handle HTTP requests." introductionVersion:"%NEXT%"`
```
pbleser-oc, line 56:
```suggestion
	Disabled             bool   `yaml:"disabled" env:"USERLOG_EVENTS_DISABLED" desc:"Disables listening for events. Set this to true if the service should only handle HTTP requests." introductionVersion:"%%NEXT%%"`
```
We should use `%%NEXT%%`, not `%NEXT%`

PR #3288, opened 2026-08-11
perf(policies): reuse the compiled rego query across evaluations

fschade:
## Description
as pointed out by @pbleser-oc we re-load the rego policies on every evaluation, on small instances that is not nice but also not noticeable. As soon as the user base and usage grows, the allocation are measurable.

This PR changes the handling of how we load rego policies, it keeps the compiled query which reduces the allocations by a measurable amount! After changing the rego, a server restart is needed.

it contains a bench test and the numbers prove the idea behind it.

| 2 policy files | before | after | factor |
| -------------------------- | ------ | ----- | ------ |
| time                       | 690 us | 45 us |
| memory                     | 332 KB | 20 KB | 17x    |
| allocations                | 6496   | 472   | 14x    |

take the numbers with a pinch of salt, its measured in my leb env!

## How Has This Been Tested?
- lab installation & load test
- perf test
- unit tests

## Types of changes

- [X] Technical debt

### DIFF
--- a/services/policies/pkg/engine/opa/engine.go
+++ b/services/policies/pkg/engine/opa/engine.go
@@ -6,64 +6,86 @@ import (
 	"os"
 	"time"
 
+	"github.com/open-policy-agent/opa/loader"
 	"github.com/open-policy-agent/opa/rego"
-	"github.com/open-policy-agent/opa/topdown/print"
+	"github.com/opencloud-eu/reva/v2/pkg/rhttp"
 
 	"github.com/opencloud-eu/opencloud/pkg/log"
 	"github.com/opencloud-eu/opencloud/services/policies/pkg/config"
 	"github.com/opencloud-eu/opencloud/services/policies/pkg/engine"
 )
 
+// downloads go to the internal data gateway, its certificate is not verified.
+const rHTTPInsecure = true
+
 // OPA wraps open policy agent makes it possible to ask if an action is granted.
 type OPA struct {
-	printHook print.Hook
-	policies  []string
-	timeout   time.Duration
-	options   []func(r *rego.Rego)
+	timeout  time.Duration
+	policies *loader.Result
+	options  []func(r *rego.Rego)
 }
 
 // NewOPA returns a ready to use opa engine.
-func NewOPA(timeout time.Duration, logger log.Logger, conf config.Engine) (OPA, error) {
+func NewOPA(timeout time.Duration, logger log.Logger, conf config.Engine) (*OPA, error) {
 	var mtReader io.ReadCloser
 
 	if conf.Mimes != "" {
 		var err error
 		mtReader, err = os.Open(conf.Mimes)
 		if err != nil {
-			return OPA{}, err
+			return nil, err
 		}
 
-		defer mtReader.Close()
+		defer func() {
+			_ = mtReader.Close()
+		}()
 	}
 
 	rfMimetypeExtensions, err := RFMimetypeExtensions(mtReader)
 	if err != nil {
-		return OPA{}, err
+		return nil, err
 	}
 
-	return OPA{
-		policies:  conf.Policies,
-		timeout:   timeout,
-		printHook: logPrinter{logger: logger},
-		options: []func(r *rego.Rego){
-			RFMimetypeDetect,
-			RFResourceDownload,
-			rfMimetypeExtensions,
-		},
+	policies, err := loader.NewFileLoader().WithProcessAnnotation(true).Filtered(conf.Policies, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	options := []func(r *rego.Rego){
+		rego.EnablePrintStatements(true),
+		rego.PrintHook(logPrinter{logger: logger}),
+		RFMimetypeDetect,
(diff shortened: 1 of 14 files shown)


services/policies/pkg/engine/opa/engine.go
```diff
@@ -2,62 +2,109 @@ package opa
...
+	revision string
+	queries  map[string]rego.PreparedEvalQuery
 }
 
 // NewOPA returns a ready to use opa engine.
-func NewOPA(timeout time.Duration, logger log.Logger, conf config.Engine) (OPA, error) {
+func NewOPA(timeout time.Duration, logger log.Logger, conf config.Engine) (*OPA, error) {
+	for _, path := range conf.Policies {
+		if info, err := os.Stat(path); err == nil && info.IsDir() {
+			return nil, fmt.Errorf("policy path %q is a directory, list the rego files instead", path)
```
pbleser-oc: Why this?
When a directory is specified, the engine loads all the `.rego` files that are in that directory, which sounds like a really convenient option for admins, no?

services/policies/pkg/engine/opa/engine.go
```diff
@@ -2,62 +2,109 @@ package opa
...
+			RFResourceDownload(rhttp.GetHTTPClient(rhttp.Insecure(rHTTPInsecure))),
 			rfMimetypeExtensions,
 		},
 	}, nil
 }
 
 // Evaluate evaluates the opa policies and returns the result.
-func (o OPA) Evaluate(ctx context.Context, qs string, env engine.Environment) (bool, error) {
+func (o *OPA) Evaluate(ctx context.Context, qs string, env engine.Environment) (bool, error) {
+	q, err := o.preparedQuery(ctx, qs)
```
pbleser-oc: Is it on purpose that we use a different `ctx` here than when calling `Eval()` below?

If it's on purpose, would you mind adding a comment to explain why that is?

services/policies/pkg/engine/opa/engine.go
fschade: we can change that, till today only .rego files were allowed, you think it makes sense?

`./**/*.rego` or `./*.rego`?

pbleser-oc: It worked for me in my tests when just indicating a directory there, I believe the recursive `.rego` file loading is in the OPA engine, if you look at `loadFiles` in `rego.go`, and the `allRec` func in `loader.go` in the `opa` library.

pbleser-oc: So there is nothing to add to be able to recurse on directories, the OPA library can already do that on its own, which is why I'm asking why we would restrict it from doing that through that `IsDir()` check there.

services/policies/pkg/engine/opa/engine.go
fschade: that context is passed down to the extensions, if the caller has no timeout set a potential download has no timeout, if the caller has set a timeout our timeout never extends it.

services/policies/pkg/engine/opa/engine.go
fschade: rego never loads files, it just "imports" whats there

services/policies/pkg/engine/opa/engine.go
pbleser-oc: I'm confused...
* the `ctx` passed to `preparedQuery` up here governs the timeout for the call to `PrepareForEval`, which parses everything
* the `ctx` passed to `Eval` below governs the timeout for actually evaluating the rules

Okay, so what you're saying is that if the caller doesn't have a timeout set on the `ctx` that is passed to `Evaluate`, then we don't care, we just don't have a timeout and that's fine for parsing the files.

But we enforce having a timeout on the evaluation.

Alright, makes sense. Just wasn't immediately clear to me, which is why I'm suggesting a comment in the code there, above the call to `preparedQuery()`

services/policies/pkg/engine/opa/engine.go
```diff
@@ -2,62 +2,111 @@ package opa
...
```
pbleser-oc:
```suggestion
    // note that we use the caller's context here because having a timeout is optional and up to the caller,
    // since this part only parses the rules, and the configured timeout
	q, err := o.preparedQuery(ctx, qs)
```

services/policies/pkg/engine/opa/engine.go
```diff
@@ -2,62 +2,111 @@ package opa
...
+	if err != nil {
+		return false, err
+	}
+
+	// only Eval gets the timeout. it hands the context down to the builtins, so
+	// a hanging download can be cut off.
```
pbleser-oc:
```suggestion
	// Eval(), in turn, gets the timeout that is set via the configuration, since we *always* want to have
	// a timeout here, regardless of whether the caller has set a timeout in the context that is passed
	// to this function: this context is then handed down to the builtin functions, in order to have e.g.
	// a hanging download be terminated if it exceeds the configured timeout duration
```

butonic (approved):

butonic:
critical errors are in tests ... merging

Issue #1312, opened 2025-08-06
[FR] Starting Event Listeners should be optional


butonic:
In https://github.com/opencloud-eu/opencloud/issues/1124 I analyzed which services start what. In order to scale request handlers and event listeners independently we need to be able to run only the handler or the listener part of a service. 

The affected services are:
- [ ] activitylog
- [ ] eventhistory
- [ ] frontend (related: https://github.com/opencloud-eu/opencloud/issues/1310)
- [ ] graph
- [ ] policies
- [ ] search (related: https://github.com/opencloud-eu/opencloud/pull/1290)
- [ ] sse
- [ ] userlog

We cannot set the service address to emptystring, eg. `ACTIVITYLOG_HTTP_ADDR=""`, because that would prevent services from using a random port. So, we should introduce a new env var eg. `ACTIVITYLOG_HTTP_DISABLE="true"` or `POLICIES_GRPC_DISABLE="true"`. 

To disable event listeners we cannot just set the service specific `OC_EVENTS_ENDPOINT` to emptystring because we may need to differentiate between publishing and listening. So, we need a new env var, eg. `GRAPH_EVENTS_DISABLE_CONSUMER` to not listen for events.

<details><summary>Using grep we can find services that Consume</summary>

```
vendor/github.com/opencloud-eu/reva/v2/pkg/share/manager/jsoncs3/jsoncs3.go:            ch, err := events.Consume(m.eventStream, "jsoncs3sharemanager", _registeredEvents...)
vendor/github.com/opencloud-eu/reva/v2/pkg/storage/pkg/decomposedfs/decomposedfs.go:            ch, err := events.Consume(fs.stream, "dcfs", _registeredEvents...)
vendor/github.com/opencloud-eu/reva/v2/pkg/storage/utils/decomposedfs/decomposedfs.go:          ch, err := events.Consume(fs.stream, "dcfs", _registeredEvents...)
services/userlog/pkg/service/service.go:        ch, err := events.Consume(o.Stream, "userlog", o.RegisteredEvents...)
services/sse/pkg/server/http/server.go: ch, err := events.Consume(options.Consumer, "sse-"+uuid.New().String(), options.RegisteredEvents...)
services/graph/pkg/service/v0/service.go:       evChannel, err := events.Consume(g.eventsConsumer, "graph", _registeredEvents...)
services/antivirus/pkg/service/service.go:      ch, err := events.Consume(natsStream, "antivirus", events.StartPostprocessingStep{})
[...]

PR #3293, opened 2026-08-12
chore(graph): disable HTTP or eventhandlers by configuration

pbleser-oc:
## Description
In the scope of the broader issue #1312, this PR deals with performing those changes for the `graph` service, namely to add the ability to disable the HTTP API or to disable the events API handler by configuration.

It also adds numerous metrics, and tests for the events processing.

### Making Event Processing and HTTP Optional (#1312)

The previous implementation was combining the HTTP server service and the events consumption, which is why this PR refactors the composition of those services:

 * the event consumption has been moved into its own service
 * the identity.Backend is created beforehand, and then injected as a collaborator in both the HTTP service as well as the event consumer service

To encourage re-use in latter implementations and changes, it also introduces two top-level package changes:

 * `internal/eventstest/events_test_helpers`: contains a `TestBus` implementation to unit-test event consumers without the need to engage NATS
 * `internal/metricstest/metrics_test_helpers`: contains assertion functions to test Prometheus metrics

Run-down of changes for this aspect:
 * add the ability to disable the HTTP API handler (`GRAPH_HTTP_DISABLED`)
 * add the ability to disable the Events API handler (`GRAPH_EVENTS_DISABLE_CONSUMER`)
 * add metrics, especially for event processing
 * add metrics to the documentation
 * introduces a new top-level packages `internal/eventstest` and `internal/metricstest` with utilities for testing event consumers and Prometheus metrics, respectively

Includes metrics for whether the HTTP API handler is enabled, and another for whether the Event consumer is enabled, to allow counting the number of instances that do either, and potentially define alerts when one of those reaches 0 (or another low threshold).

image

### Adding Metrics

Introducing [gowrap](https://github.com/hexdigest/gowrap) as a build-time tool to generate interface delegate structs from templates:

 * added as a `make generate` target in `services/graph`,
 * added as a build-time dependency in `.bingo/`

#### LDAP Client Metrics

Introduces LDAP client abstraction interface to be able to wrap the `go-ldap` client API with metrics transparently (and possibly hooks and such in the future), in order to use delegation patterns to measure the time LDAP (client) operations take to finish, as well as to track their results (success, failure, not-found).

Has two implementations that are generated using `gowrap`:
 * a `go-ldap` adapter implementation that directly delegates
 * a time measuring and metrics collecting implementation that delegates to another `LdapClient`

The metrics collecting one is disabled by default, can be enabled with `GRAPH_LDAP_METRICS_DISABLE=false`

It collects the duration of outbound LDAP client operations into a histogram, as well as the number of concurrent outbound LDAP operations in a gauge (via an atomic int and a gauge func for best performance).

image

image

#### HTTP Requests Metrics

Add an HTTP middleware that measures how long Graph HTTP API requests take, storing taken time into a histogram along with labels for
 * method,
 * path pattern (from the chi routes),
 * Graph API version prefix,
 * Graph API resource name,
 * and the resulting status code.

It also tracks the number of concurrent inbound Graph API HTTP requests using a gauge (also via an atomic int and a gauge func for best performance).

Disabled by default, can be enabled with `GRAPH_HTTP_METRICS_DISABLE=false`

image

#### LDAP Identity Backend Metrics

Add `Backend` and `EducationBackend` delegate implementations that measure execution time on the level of the higher API call operations there (`CreateUser`, `DeleteUser`, ..., `CreateSchool`, ...), generated using `gowrap`.

Disabled by default, can be enabled with `GRAPH_IDENTITY_BACKEND_METRICS_DISABLE=false`

image

### LDAP Client API Change

Made an **internal** change to how the singular entry search operations work in the LDAP identity backend:

 * check whether searches for a singular entry returns more than one result, in which case a new error `TooManyResults` is returned, instead of leaving that undetected, blindly taking the first result, and potentially risking data inconsistencies

### Logging Improvements

Improve the loggers in identity backends by adding attributes for their request targets (Reva gateway address or LDAP URI, respectively).

Also add a "backend" attribute for all Graph API logs (set to `"ldap"` or `"cs3"`), to help debug potential issues.

The LDAP identity backend logger also has two new attributes to help debugging with logs:
 * `write` (bool): whether write operations are enabled
 * `refint` (bool): whether refint is enabled or not

Also adds a dedicated counter for user password change operations.

### Additional boy-scouting

 * add a constructor func for the CS3 backend
 * add a constructor func for the LDAP backend
 * in the LDAP identity backend, in `searchLDAPEntryByFilter` (used by all search/get public functions), errors that occur when performing LDAP `SEARCH` operations were blindly mapped to a `ItemNotFound` error, instead of being analyzed as it could be caused by a technical error
 * in the `requireadmin` middleware, add debug logging to explain why a request is denied
 * when an LDAP password change fails because the user entry was not found in LDAP, we now have a log message that tracks that

## Related Issue
- fixes a part of #1312 

## Motivation and Context
Details can be found in #1312

### HTTP API
Tested the HTTP API (when enabled) using:
```bash
xh --verify=no -a alan:demo --https :9200/graph/v1.0/me
```
and
```bash
xh --verify=no -a alan:demo --https :9200/graph/v1.0/users/b1f74ec4-dd7e-11ef-a543-03775734d0f7
```
and
```bash
xh --verify=no -a admin:admin --https :9200/graph/v1.0/users \$orderby=='displayName asc'
```
and
```bash
xh --verify=no -a admin:admin --https POST :9200/graph/v1.0/users accountEnabled:=true displayName="Jane Doe" onPremisesSamAccountName=jdoe mail="jdoe@example.com" passwordProfile:='{"password": "secret12345"}'
```

### Events API
Tested the event API (when enabled) using:
```bash
echo -n '{"metadata":{"eventtype":"events.UserSignedIn"},"payload":"'$(echo -n '{"Executant":{"opaque_id":"alan"}}'|base64)'"}'|nats pub main-queue
```

Tested combinations of enabling/disabling the events and HTTP APIs.
When both are disabled, the service refuses to start and exits early with an error message.

### Metrics
Metrics were tested with a small [`k6`](https://k6.io/) script included in this PR (namely `services/graph/load_test.js`) that performs various read-only operations with parallel VUs.

```bash
k6 run --vus 20 --duration 10m ./services/graph/load_test.js
```

## Types of changes
- [x] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation added (in `services/graph/README.md`)


services/graph/pkg/middleware/requireadmin.go
```diff
@@ -12,15 +12,21 @@ import (
...
 			if !ok {
 				errorcode.AccessDenied.Render(w, r, http.StatusUnauthorized, "Unauthorized")
 				return
 			}
+			if u == nil {
+				l.Debug().Msg("Bad request: user is missing")
+				errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "user is missing an id")
+				return
+			}
 			if u.Id == nil || u.Id.OpaqueId == "" {
```
butonic: with the generated protobuf code we can safely do this:
```suggestion
			if u.GetId().GetOpaqueId() == "" {
```

services/graph/pkg/service/v0/users.go
```diff
@@ -79,6 +79,12 @@ func (g Graph) GetMe(w http.ResponseWriter, r *http.Request) {
 			errorcode.RenderError(w, r, err)
 			return
 		}
+		if me == nil {
+			// this is an error
+			logger.Debug().Err(err).Interface("query", r.URL.Query()).Msg("could not get users: user not found in backend")
```
butonic: is this a leftover?

services/graph/pkg/config/parser/parse.go
```diff
@@ -39,6 +39,14 @@ func ParseConfig(cfg *config.Config) error {
 }
 
 func Validate(cfg *config.Config) error {
+	if cfg.HTTP.Disabled && cfg.Events.DisabledConsumer {
+		// might be debatable, but this situation should be treated as an error,
+		// as the process wouldn't be able to serve either API and would thus be
+		// completely useless -- in that case, just don't start this service
+		// in the first place (especially since it's optional)
+		return errors.New("both HTTP and events consumption APIs are disabled by configuration; at least one must be enabled")
```
butonic: there is now shared.AllComponentsDisabledError

services/graph/pkg/identity/ldap_group.go
```diff
@@ -616,12 +633,14 @@ func (i *LDAP) isLDAPGroupReadOnly(e *ldap.Entry) bool {
...
 }
 
-func (i *LDAP) groupsFromLDAPEntries(e []*ldap.Entry) []libregraph.Group {
+func (i *LDAP) groupsFromLDAPEntries(e []*ldap.Entry) ([]libregraph.Group, error) {
 	groups := make([]libregraph.Group, 0, len(e))
 	for _, g := range e {
-		if grp := i.createGroupModelFromLDAP(g); grp != nil {
+		if grp, err := i.createGroupModelFromLDAP(g); err != nil {
+			return nil, err
+		} else if grp != nil {
```
butonic, line 645: this stops after the first error

services/graph/pkg/middleware/requireadmin.go
```diff
@@ -48,6 +49,7 @@ func RequireAdmin(rm *roles.Manager, logger log.Logger) func(next http.Handler)
 				return
 			}
 
+			l.Debug().Str("userid", u.Id.OpaqueId).Str("permission", settings.AccountManagementPermissionID).Msg("Access denied: necessary permission %q not present in user's roles")
```
butonic:
```suggestion
			l.Debug().Str("userid", u.Id.OpaqueId).Str("permission", settings.AccountManagementPermissionID).Msg("Access denied: necessary permission not present in user's roles")
```

services/graph/pkg/service/v0/users.go
```diff
@@ -778,33 +786,41 @@ func (g Graph) DeleteUser(w http.ResponseWriter, r *http.Request) {
...
 		us.Reason = "User soft deleted via Graph API" // TODO: this needs a proper implementation through the request
 		us.TimeStamp = time.Now()
-		err = g.setUserStateToNatsKeyValue(r.Context(), userID, us)
-		if err != nil {
+		if err := g.setUserStateToNatsKeyValue(r.Context(), userID, us); err != nil {
 			logger.Error().Err(err).Str("id", userID).Msg("could not set user state")
 			errorcode.RenderError(w, r, err)
 			return
+		} else {
+			// note: logging these as WARN for backwards compatibility reason since they previously were not logged at all
```
butonic: the else branch is not needed here because we return in case there was an error.

butonic (approved):
This touches a lot of LDAP code and also introduces gowrap tooling, so I'd like to get @rhafer to take a look at this as well.

services/graph/pkg/identity/backend.go
```diff
@@ -1,5 +1,8 @@
 package identity
 
+//go:generate gowrap gen -g -i Backend -t ./backend_prometheus.tmpl -o backend_prometheus.go
```
rhafer: This does not work with the bingo installed version-pinned `gowrap` as it needs `gowrap` got be installed in `$PATH`.

To make it work with the bingo installed `gowrap` I think you need to replace the `go generate ./...` in the `Makefile` with `GOWRAP=$(GOWRAP) go generate ./...` and use

```
//go:generate $GOWRAP gen ....
```

`<rant>` IMO at some point we should really get away from `bingo` and switch to using `go tool` for managing our external go based dependencies. (reva does that already) `</rant>`

services/graph/pkg/identity/ldap_client.go
```diff
@@ -0,0 +1,26 @@
+package identity
+
+//go:generate gowrap gen -g -i LdapClient -t ./ldap_client_prometheus.tmpl -o ldap_client_prometheus.go
```
rhafer: Same here about the bingo installed `gowrap`

services/graph/pkg/identity/ldap.go
```diff
@@ -454,16 +546,33 @@ func (i *LDAP) getEntryByDN(dn string, attrs []string, filter string) (*ldap.Ent
...
 	}
-	if len(res.Entries) == 0 {
+
+	switch len(res.Entries) {
+	case 0:
 		return nil, ErrNotFound
+	case 1:
+		return res.Entries[0], nil
+	default:
+		return nil, ErrTooManyResults
```
rhafer, line 567: we're already searching with a sizelimit of 1, here (see the NewSearchRequest call above), so this line will never be reached.

If a search would have returned more that 1 result, `Search` will return `ldap.LDAPResultSizeLimitExceeded` in `err`.

services/graph/pkg/identity/ldap.go
```diff
@@ -487,14 +596,24 @@ func (i *LDAP) searchLDAPEntryByFilter(basedn string, attrs []string, filter str
...
+		return nil, i.mapLDAPError(err, errMap)
 	}
-	if len(res.Entries) == 0 {
+	switch len(res.Entries) {
+	case 0:
 		return nil, ErrNotFound
+	case 1:
+		return res.Entries[0], nil
+	default:
+		return nil, ErrTooManyResults
```
rhafer, line 617: Same here. We need to map the `LDAPResultSizeLimitExceeded` error instead.

services/graph/pkg/identity/ldap_education_school.go
```diff
@@ -715,26 +759,37 @@ func (i *LDAP) getSchoolByFilter(filter string) (*ldap.Entry, error) {
...
 	if err != nil {
-		var errmsg string
 		if lerr, ok := err.(*ldap.Error); ok {
 			if lerr.ResultCode == ldap.LDAPResultSizeLimitExceeded {
-				errmsg = fmt.Sprintf("too many results searching for school '%s'", filter)
-				i.logger.Debug().Str("backend", "ldap").Err(lerr).
+				errmsg := "too many results searching for school"
+				i.logger.Warn().Err(lerr).
 					Str("schoolfilter", filter).Msg("too many results searching for school")
+				return nil, errorcode.New(errorcode.TooManyResults, errmsg)
```
rhafer: Can't we include this into the `ldapResultToErrMap` below?

services/graph/pkg/identity/ldap_group.go
```diff
@@ -501,11 +510,15 @@ func (i *LDAP) getLDAPGroupByFilter(filter string, requestMembers bool) (*ldap.E
...
-	}
 
-	return e[0], nil
+	switch len(e) {
+	case 0:
+		return nil, ErrNotFound
+	case 1:
+		return e[0], nil
+	default:
+		return nil, ErrTooManyResults
```
rhafer, line 520: One more about the `LDAPResultSizeLimitExceeded` error. `getLDAPGroupsByFilter` sets a sizelimit for us and returns `errcode.ItemNotFound` when it is hit. I guess the error handling in `getLDAPGroupsByFilter` needs to be adjusted.

rhafer (changes requested):
I've focussed the review of the LDAP related changes. Main issue is that we need to map the `ldap.LDAPResultSizeLimitExceeded` error instead of checking the number of returned entries, since we already set a sizelimit of 1 for all request that expect just a single result entry.

Other than that looks nice. I like the logging improvements, and improved error handling. 

I must admin that am not really a fan of generating code from templates, but I also see the benefit here. I am already wondering if we could "re-implement" the LDAP reconnect wrapper (https://github.com/opencloud-eu/reva/blob/main/pkg/utils/ldap/reconnect.go) in a similar way :).

services/graph/pkg/identity/backend.go
pbleser-oc: Good point, went unnoticed on my end because I had a `mise` managed `gowrap` in my `PATH`.

Changed as outlined above, works.

Also removed `gowrap` from the `mise.toml` again, as we should always use the `bingo` managed one via `make`, at least for now.

services/graph/pkg/identity/ldap.go
pbleser-oc: Fair, missed it because it's just a casual `1` parameter in `ldap.NewSearchRequest()` :)

I changed the `mapLDAPError` calls to include checking for `ldap.LDAPResultSizeLimitExceeded` and return `errorcode.TooManyResults`.

But I also chose to keep the result entries length check in there as well, because
* it needs to be there anyway for the "found no element" case
* in case someone changes the size limit by mistake, it doesn't harm to have the additional check down there as well, especially since it's always in functions that, by semantics, have the purpose of finding a single entry

rhafer: Sounds good.

rhafer (approved):
Looks good to me now. I've kicked the CI once more, something seems to be flaky there :( .

rhafer:
@pbleser-oc The `test-lint-unit` failure is actually a real one:

```
=== RUN   TestGetGroup
2026-09-01T07:39:29Z DBG GetGroup line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:30 refint=false request-id= service=opencloud write=true
2026-09-01T07:39:29Z DBG getLDAPGroupsByFilter attributes=["cn","entryUUID","member"] base=ou=groups,dc=test filter=(&(objectClass=groupOfNames)(|(cn=group)(entryUUID=group))) line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:553 refint=false scope=2 service=opencloud sizelimit=1 write=true
    ldap_group_test.go:63: 
        	Error Trace:	/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group_test.go:63
        	Error:      	Error "tooManyResults: failed to search for LDAP groups" does not contain "itemNotFound:"
        	Test:       	TestGetGroup
2026-09-01T07:39:29Z DBG GetGroup line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:30 refint=false request-id= service=opencloud write=true
2026-09-01T07:39:29Z DBG getLDAPGroupsByFilter attributes=["cn","entryUUID","member"] base=ou=groups,dc=test filter=(&(objectClass=groupOfNames)(|(cn=group)(entryUUID=group))) line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:553 refint=false scope=2 service=opencloud sizelimit=1 write=true
    ldap_group_test.go:65: 
        	Error Trace:	/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group_test.go:65
        	Error:      	Error "tooManyResults: failed to search for LDAP groups" does not contain "itemNotFound:"
        	Test:       	TestGetGroup
2026-09-01T07:39:29Z DBG GetGroup line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:30 refint=false request-id= service=opencloud write=true
2026-09-01T07:39:29Z DBG getLDAPGroupsByFilter attributes=["cn","entryUUID","member"] base=ou=groups,dc=test filter=(&(objectClass=groupOfNames)(|(cn=group)(entryUUID=group))) line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:553 refint=false scope=2 service=opencloud sizelimit=1 write=true
    ldap_group_test.go:67: 
        	Error Trace:	/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group_test.go:67
        	Error:      	Error "tooManyResults: failed to search for LDAP groups" does not contain "itemNotFound:"
        	Test:       	TestGetGroup
2026-09-01T07:39:29Z DBG GetGroup line=/go/src/github.com/opencloud-eu/opencloud/services/graph/pkg/identity/ldap_group.go:30 refint=false request-id= service=opencloud write=true
```

pbleser-oc:
Fixed the broken tests in `ldap_group_test.go`, spliced it into the commit with a fp to keep meaningfully squashed commits for the merge.

PR #3297, opened 2026-08-12
fix(collaboration): harden the mention notification endpoint

### LINKED ISSUE #3128: collaboration @user mentions review
## Issues with the `/collaboration/notify` endpoint

We introduced a new public endpoint here that is  un-documented and completely unspecified. I guess this should be reworked into something on the graph service. 

Do we really need the functionality to bulk-notifiy a list of users? This provides quite some potential for abuse. If this was just a single request per notified user it should be far easier to impose rate-limits and other protection.

### 1. unused `type` field
The `type` field is validated as required but never read. It isn't put on the event or used for branching. The API contract forces clients to send a field that has no effect. It's documented nowhere what this field is supposed to be.

### 2. Requester's access to `fileID` is never checked, allows to send spoofed mentions

The only file-access check is that each mentioned user can `Stat` the file. Nothing verifies the executant has any relationship to `fileID`. Any user who knows/guesses a valid `fileID` and the `userIDs` of people who do have access can generate a real notification:

> "<attacker display name> mentioned you in Q3-Financials.docx" with a working `resourceLink` to the actual file.
 
This also means that the handler can be abused to figure out stuff like, "does user xyz have access to file abc". This could be a possible information disclosure issue.

### 3. No bound on `userIDs`, no dedup

`userIDs` is caller-controlled and iterated with no length cap, each iteration doing a gRPC `Authenticate` + `Stat`. A single request can fan out into a large number of gateway calls. There's also no dedup: repeating the same `userID` produces duplicate entries in `event.UserIDs`, and the downstream `handleResourceMention` builds recipients from that list without dedup, so a user can be emailed N times from one request. This allows to flood users with notifications.

### 4. A single failed recipient lookup or stat request aborts the whole notification - `service.go:115-145`

If any one mentioned user fails machine auth or lacks `Stat` access to the file, the handler returns 500 and no event is published for anyone.

### 6 . Minor issues

-  `HandleNotification()` in service.go:
    - storagespace.ParseID(data.FileID) is called in every loop iteration, could be done outside, same is true for setting event.Ref
    -  `authResponse.Status.Code` / `statResponse.Status.Code` should be  `GetStatus().GetCode()` to avoid potential nil pointer panics

-  `handleResourceMention` in `services/notification/pkg/service/resource.go`:
    -  `logs "could not select next gateway client" for an error from `GetServiceUserContextWithContext`.

fschade:
## Description

- the requester now has to have access to the file
- recipients are deduped
- a recipient that cannot be resolved no longer drops the notification for everyone else
- errors map to 400/403/404 instead of 500
- remove unused type field is gone from the request body.
- move the mention endpoint to graph

## Related Issue
- Fixes https://github.com/opencloud-eu/opencloud/issues/3128
- Needs https://github.com/opencloud-eu/libre-graph-api/pull/64
- Dep https://github.com/opencloud-eu/web/pull/3114
- Update https://github.com/opencloud-eu/web/pull/3060

## How Has This Been Tested?
- unit tests

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,155 @@
...
+		return
+	}
+
+	recipientStat, err := gatewayClient.Stat(
+		metadata.AppendToOutgoingContext(ctx, revactx.TokenHeader, authResponse.GetToken()),
+		&storageprovider.StatRequest{Ref: ref},
+	)
+	if err != nil || recipientStat.GetStatus().GetCode() != rpc.Code_CODE_OK {
+		g.logger.Debug().Str("userID", userID).Msg("the recipient has no access to the item")
+		errorcode.AccessDenied.Render(w, r, http.StatusForbidden, "the recipient cannot access the item")
```
butonic: I wonder if we should expose to the client that makes this request that the recipient has no access to the file. It would allow users to probe who has access.

🤔 we should stat the file with the currently logged in user to prevent users from mentioning people in files they don't have access to ... oh that would even allow them to probe who has access to a file they don't have access to.

I would probably send a 200 OK after stating the file with the senders context and then use a go routine to check if the recipient has acces and then send the actual event. And only log a debug message that explains why we are not returning an error in the case the recipient has no access. Or ... we just send out the event. If the recipient has no access he wil lge t a permission denied and can ping the sender that he forgot to share the file with him ... ? Who is allowed to notify whom? 🤔

We don't want a way for users to annoy other users by sending anyone on the platform notifications ... so ... we need to check access permissions for sender and recipient. And then we can also return a 403 Forbidden to indicate that the sender is not allowed to notify the recipient for whatever reason. In debug mode we can return the recipients lacking permissions as an inner error ... if we want to go that far.

butonic (changes requested):

services/graph/pkg/service/v0/teamwork.go
fschade: done, please check if the status is fine for you

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,156 @@
...
+		g.logger.Debug().Err(err).Msg("user is not allowed to publish activity notifications")
+		errorcode.AccessDenied.Render(w, r, http.StatusForbidden, "not allowed to publish activity notifications")
+		return
+	}
+
+	userID, err := url.PathUnescape(chi.URLParam(r, "userID"))
+	if err != nil || userID == "" {
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "invalid user id")
+		return
+	}
```
butonic, line 64: This Bad Request check is cheaper than the Permission check and shoud happen before it

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,156 @@
...
+	case notification.ActivityType != _activityTypeMentioned:
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unsupported activityType")
+		return
+	case notification.TeamsAppID != WebOfficeAppID:
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unknown teamsAppId")
+		return
+	case notification.Topic.Source != _topicSourceText:
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "unsupported topic source")
+		return
+	}
```
butonic, line 83: same for this

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,156 @@
...
+
+	itemID, err := storagespace.ParseID(notification.Topic.Value)
+	if err != nil || itemID.GetStorageId() == "" || itemID.GetSpaceId() == "" || itemID.GetOpaqueId() == "" {
+		g.logger.Debug().Err(err).Msg("could not parse the topic value")
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "the topic value is no resource id")
+		return
+	}
```
butonic, line 90: aaand this

butonic (changes requested):

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// answer with 202 to hide who is allowed to access the resource.
+	authResponse, err := gatewayClient.Authenticate(ctx, &gateway.AuthenticateRequest{
+		Type:         "machine",
+		ClientId:     "userid:" + userID,
+		ClientSecret: g.config.MachineAuthAPIKey,
+	})
+	if err != nil || authResponse.GetStatus().GetCode() != rpc.Code_CODE_OK {
+		g.logger.Debug().Str("userID", userID).Msg("mention dropped, could not authenticate the recipient")
+		w.WriteHeader(http.StatusAccepted)
```
butonic: If the user is not found we can return a 404 because there are other endpoints where users can be looked up. If there is another error we can map that properly as well (IIRC we have something that can be reused for mapping CS3 errors to graph errors). Maybe a simple catch all with internal server error is good enough

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,156 @@
...
+		return
+	}
+
+	recipientStat, err := gatewayClient.Stat(
+		metadata.AppendToOutgoingContext(ctx, revactx.TokenHeader, authResponse.GetToken()),
+		&storageprovider.StatRequest{Ref: ref},
+	)
+	if err != nil || recipientStat.GetStatus().GetCode() != rpc.Code_CODE_OK {
+		g.logger.Debug().Str("userID", userID).Msg("mention dropped, the recipient has no access to the item")
```
butonic, line 170: Only if the CS3 error is a not found or forbidden error. If there is a real error we should return an internal server error.

butonic (changes requested):

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -105,38 +105,60 @@ func (g Graph) SendActivityNotification(w http.ResponseWriter, r *http.Request)
...
 		g.logger.Error().Err(err).Msg("could not stat item")
 		errorcode.GeneralException.Render(w, r, http.StatusInternalServerError, "could not stat item")
 		return
-	case statResponse.GetStatus().GetCode() != rpc.Code_CODE_OK:
+	case statResponse.GetStatus().GetCode() == rpc.Code_CODE_NOT_FOUND,
+		statResponse.GetStatus().GetCode() == rpc.Code_CODE_PERMISSION_DENIED:
 		errorcode.ItemNotFound.Render(w, r, http.StatusNotFound, "item not found")
 		return
+	case statResponse.GetStatus().GetCode() != rpc.Code_CODE_OK:
+		g.logger.Error().Str("code", statResponse.GetStatus().GetCode().String()).Msg("could not stat item")
```
butonic: there is also a `GetStatus().GetMessage()` we should log

butonic: ah, in graph/pkg/errorcode/cs3.go there is FromCS3Status that you can reuse!

butonic (approved):

services/graph/pkg/service/v0/teamwork.go
```diff
@@ -0,0 +1,189 @@
...
+			Str("userID", userID).
+			Str("code", authResponse.GetStatus().GetCode().String()).
+			Str("message", authResponse.GetStatus().GetMessage()).
+			Msg("could not authenticate the recipient")
+		errorcode.RenderError(w, r, errorcode.FromCS3Status(authResponse.GetStatus(), nil))
+		return
+	}
+
+	recipientStat, err := gatewayClient.Stat(
+		metadata.AppendToOutgoingContext(ctx, revactx.TokenHeader, authResponse.GetToken()),
```
rhafer: Hm, the context already carries a `x-access-token` for the original user (added by the graph/pkg/middleware/auth middleware).  This will not overwrite the existing token, so the Stat request is still executed in the context of the original user.

rhafer (changes requested):

services/graph/pkg/service/v0/teamwork.go
rhafer:
```
md, _ := metadata.FromOutgoingContext(ctx)
md = md.Copy()
md.Set(revactx.TokenHeader, authResponse.GetToken())

recipientCtx := metadata.NewOutgoingContext(ctx, md)
recipientStat, err := gatewayClient.Stat(
	recipientCtx,
	&storageprovider.StatRequest{Ref: ref},
)
```

fschade: GOOD CATCH, thx, fixed!

rhafer (approved):

PR #3299, opened 2026-08-12
feat(search): scope searches to a drive via the driveId field

dschmidt:
`scope:` takes an opaque resource id, which is hostile to hand-written queries. Accept `driveId:"<storage$space>"` as a regular KQL field instead: it resolves to the indexed `RootID` and a bare drive id is completed to the root resource id. Combined with `path:` this gives a readable location scope with no token stripping, both compose with groups, OR and NOT like any other field.

Second commit prunes the space fan-out when top-level AND conjuncts pin the query to a single root; conservative rules, the restriction stays in the query, mountpoints are kept for path mapping.

Follow-up (deliberately not here): parse the query once in the service and hand the AST to the engines, which currently re-parse per space; that changes the engine interface.

Stacked on #3345 (reflection-based search mapping).


dschmidt:
Replaced by #3424: same commits rebased onto `main` after the base branch was deleted on the #3345 merge.

PR #3307, opened 2026-08-14
[docs-only] feat: add mise toml

fschade:
## Description
add support for mise

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,133 @@
+[tools]
+go = "1.25"
+node = "24"
+pnpm = "11.1.3"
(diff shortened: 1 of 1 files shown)


pbleser-oc:
Works for me, haven't tested all the mise tasks, but they obviously work for you. And even if they didn't, it wouldn't break anything.

Can be merged, and once we get enough people using `mise`, we could add a bit of developer documentation somewhere to signal that `mise` is the proper tool to do all this (and to manage versions especially).

pbleser-oc (approved):
lgtm

PR #3313, opened 2026-08-14
feat(web): add rclone-crypt to default apps

JammingBen:
Enable the `rclone-crypt` app per default for the upcoming release.


kulmann (approved):
😍

PR #3315, opened 2026-08-16
feat(search): semantic image search with CLIP embeddings

dschmidt:
`semantic:"..."` in KQL embeds the query text via a CLIP inference service (immich machine-learning API, multilingual model) and ranks image vectors by cosine similarity. Images are embedded at index time by a `ClipExtractor` decorating the configured extractor. bleve searches via faiss KNN behind the new `vectors` build tag (RRF fusion for hybrid queries; vectors round-trip through a stored-only sibling field because bleve cannot return vector-typed fields), OpenSearch via `knn_vector` plus client-side RRF. The filter part of a query keeps its regular meaning and stays the only source of totals; purely semantic queries return only the KNN ranking with raw similarity scores. The semantic clause is split off the parsed KQL tree (`query.ExtractSemantic`), so quoted values like `name:"*semantic:x*"` stay literals.

Enable with `SEARCH_EXTRACTOR_CLIP_URL` pointing at an immich-machine-learning instance (works with `basic` and `tika`); bleve additionally needs a binary built with `ENABLE_VECTORS=true` (see the faiss stage in Dockerfile.multiarch, pinned to the blevesearch/faiss commit matching the vendored go-faiss). Existing content needs a force reindex to get vectors.

Open questions:
- configurable vector dimensionality and model choice (currently 512, a schema constant; the startup probe hard-fails on mismatch)
- send originals or thumbnails to the inference service (currently originals with a 50MB cap, reversible per reindex)
- capability flag so clients can discover semantic support
- web UX: the search bar wraps input as `name:"*...*"`, needs a client change to issue semantic queries
- `semantic:` in OR/NOT position degrades silently to an AND filter, could be rejected instead
- release story: faiss in the official images, static binaries stay without bleve vector support, CI job for the vectors build tag
- AGPL: immich-machine-learning is referenced as an external, unmodified service

### DIFF
--- a/opencloud/Makefile
+++ b/opencloud/Makefile
@@ -7,6 +7,12 @@ ifdef ENABLE_VIPS
 	TAGS := ${TAGS},enable_vips
 endif
 
+# bleve KNN support; needs libfaiss_c at build and run time (see the vectors
+# stage in docker/Dockerfile.multiarch)
+ifdef ENABLE_VECTORS
+	TAGS := ${TAGS},vectors
+endif
+
 ifneq (, $(shell command -v go 2> /dev/null)) # suppress `command not found warnings` for non go targets in CI
 include ../.bingo/Variables.mk
 endif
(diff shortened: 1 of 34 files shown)


dschmidt:
superseded, reopened from the upstream repo

PR #3324, opened 2026-08-17
feat: update space template image

AlexAndBear:
## Summary
- Updated the space template image to a new version

## Test plan
- [ ] Verify that the new space template image is displayed correctly
- [ ] Check that space creation with templates works as expected


services/graph/pkg/service/v0/spacetemplate/image.png
JammingBen, line 1: 1MB is huge, it was only a few kb before. Can you reduce the size?

.woodpecker.star
```diff
@@ -401,7 +401,8 @@ config = {
         "4": {
             "skip": False,
             "suites": [
-                "user-settings/",
+                # skip user-settings due to failing pagination test. see https://github.com/opencloud-eu/opencloud/pull/3324
+                #"user-settings/",
```
JammingBen, line 405: Why does this fail now? 🤔 Or is it just flaky?

AlexAndBear: Because of the size change, the e2e tests now detects a change in the toc. We need to update the the test in web and enable it after that here

JammingBen (approved):

PR #3325, opened 2026-08-17
fix(activitylog): log missing parent id cache entry at debug level

### LINKED ISSUE #1043: Unknown cache errors
## Describe the bug
When moving or deleting files on my Opencloud server, I have come across some errors in the container logs. The files appear to move and delete successfully, but I wanted to raise this report in-case something is going wrong under the hood.

## Steps to reproduce
1. Create a folder
2. Move a file to that folder
3. Delete a file

## Expected behaviour
Actions are performed with no errors in the container logs

## Actual behaviour
Three types of errors occur at various stages (logs provided)
- ERR failed to handle moved away item error=error: not found: path not found in cache
- ERR could not delete parent id cache error="key not found"
- ERR could not remove lock file

## Setup
- I only wish to use the Opencloud fileserver, I am not interested in things like Collabra etc.
- I am running Opencloud-rolling:latest using Docker, on a Ubuntu 24.04 Virtual Machine. This VM is managed by Proxmox.
- I use Caddy for my reverse proxy
- I use Zitadel for OAuth
- I am pointing the POSIX root to a virtoFS mount on my VM (/mnt/cloud/opencloud). This virtioFS mount points to a ZFS pool on my Proxmox host where I plan on keeping my files for Opencloud. I do not want to store the files on the VM itself. 
- Extended attributes are enabled on the virtioFS configuration for this VM
- The directories on both the VM and HOST are set to 1000:1000 permissions, as specified in OC docs
- I have defined a CSP.yaml, using the example given in the opencloud_full deployments
- I have defined a proxy.yaml, this simply looks for my Client ID from Zitadel and makes me an admin when I login, for testing purposes.
- Container log is attached

Here is my compose.yaml

```yaml
services:
  opencloud:
    image: opencloudeu/opencloud-rolling:latest
    container_name: opencloud
    restart: unless-stopped
    volumes:
      - config:/etc/opencloud
      - data:/var/lib/opencloud
      - /mnt/cloud/opencloud:/var/lib/opencloud-posix
      # Local CSP with my auth domain added
      - ./config/csp.yaml:/etc/opencloud/csp.yaml
      # Local proxy.yaml contains role assignment instructions to make myself an admin
      - ./config/proxy.yaml:/etc/opencloud/proxy.yaml
    ports:
      - "{MY PORT}:9200"   
    logging:
      driver: local
    entrypoint:
      - /bin/sh
    command: ["-c", "opencloud init || true; opencloud server"]
    environment:
      OC_URL: https://{MY DOMAIN}
      OC_LOG_LEVEL: error
      OC_LOG_PRETTY: true
      # Reverse proxy handles certificates, so these two are set to False
      PROXY_TLS: false
      OC_INSECURE: true
      
      # These four are from the deployment example on the OC github, no idea what they do
      GATEWAY_GRPC_ADDR: 0.0.0.0:9142
      MICRO_REGISTRY_ADDRESS: 127.0.0.1:9233
      NATS_NATS_HOST: 0.0.0.0
      NATS_NATS_PORT: 9233
      # custom CSP file
      PROXY_CSP_CONFIG_FILE_LOCATION: /etc/opencloud/csp.yaml     

      # use posixFS for more human readable file structure
      STORAGE_USERS_DRIVER: posix
      # Recommended cache storage driver from the OC docs when using posix
      STORAGE_USERS_ID_CACHE_STORE: nats-js-kv
      # Not sure if I need this line as well
      # STORAGE_USERS_FILEMETADATA_CACHE_STORE: nats-js-kv
      STORAGE_USERS_POSIX_ROOT: /var/lib/opencloud-posix
      # I am using virtioFS, to link to a ZFS pool on the host, do I need this??
      STORAGE_USERS_POSIX_WATCH_FS: true 
            
      ####### INTERNAL AUTH #######
      # PROXY_ENABLE_BASIC_AUTH: true
      #############################

      ########### EXTERNAL AUTH ###########
      OC_OIDC_ISSUER: https://{MY ZITADEL DOMAIN}
      EXTERNAL_OIDC_DOMAIN: {MY ZITADEL DOMAIN}
      # Needed to access all the relevant sub-URLs for Auth
      PROXY_OIDC_REWRITE_WELLKNOWN: true
      # App ID from Zitadel
      WEB_OIDC_CLIENT_ID: {MY ZITADEL CLIENT ID}
      # Doesn't work without this, no idea why
      PROXY_OIDC_ACCESS_TOKEN_VERIFY_METHOD: none
      # Make accounts for OC if they don't already exist, based on claims from Zitadel
      PROXY_AUTOPROVISION_ACCOUNTS: true
      # sub is used due to the ClientID never changing, where as users sometimes change their preferred_name
      PROXY_USER_OIDC_CLAIM: sub
      PROXY_USER_CS3_CLAIM: username
      PROXY_AUTOPROVISION_CLAIM_EMAIL: email
      PROXY_AUTOPROVISION_CLAIM_DISPLAYNAME: name
      PROXY_AUTOPROVISION_CLAIM_GROUPS: groups
      PROXY_AUTOPROVISION_CLAIM_USERNAME: sub
      OC_EXCLUDE_RUN_SERVICES: idp
      OC_ADMIN_USER_ID: ""
      GRAPH_ASSIGN_DEFAULT_USER_ROLE: false
      GRAPH_USERNAME_MATCH: none
      GRAPH_LDAP_SERVER_WRITE_ENABLED: true
      ######################################

volumes:
  config:
  data:
```

Here is my csp.yaml

```yaml
directives:
  child-src:
    - '''self'''
  connect-src:
    - '''self'''
    - 'blob:'
    # - 'https://${COMPANION_DOMAIN|companion.opencloud.test}/'
    # - 'wss://${COMPANION_DOMAIN|companion.opencloud.test}/'
    - 'https://raw.githubusercontent.com/opencloud-eu/awesome-apps/'
    # - 'https://${KEYCLOAK_DOMAIN|keycloak.opencloud.test}/'
    - 'https://{MY-ZITADEL-DOMAIN}'
  default-src:
    - '''none'''
  font-src:
    - '''self'''
  frame-ancestors:
    - '''self'''
  frame-src:
    - '''self'''
    - 'blob:'
    - 'https://embed.diagrams.net/'
    # In contrary to bash and docker the default is given after the | character
    # - 'https://${ONLYOFFICE_DOMAIN|onlyoffice.opencloud.test}/'
    # - 'https://${COLLABORA_DOMAIN|collabora.opencloud.test}/'
    # This is needed for the external-sites web extension when embedding sites
    - 'https://docs.opencloud.eu'
  img-src:
    - '''self'''
    - 'data:'
    - 'blob:'
    - 'https://raw.githubusercontent.com/opencloud-eu/awesome-apps/'
    # In contrary to bash and docker the default is given after the | character
    # - 'https://${ONLYOFFICE_DOMAIN|onlyoffice.opencloud.test}/'
    # - 'https://${COLLABORA_DOMAIN|collabora.opencloud.test}/'
  manifest-src:
    - '''self'''
  media-src:
    - '''self'''
  object-src:
    - '''self'''
    - 'blob:'
  script-src:
    - '''self'''
    - '''unsafe-inline'''
  style-src:
    - '''self'''
    - '''unsafe-inline'''
```

Here is my proxy.yaml

```yaml
role_assignment:
    driver: oidc
    oidc_role_mapper:
        role_claim: sub
        role_mapping:
            - role_name: admin
              claim_value: {MY ZITADEL USER ID}
```

[OC container logs.txt](https://github.com/user-attachments/files/20690997/OC.container.logs.txt)

junkerderprovinz:
## Description

Issue #1043 reports three ERROR log lines that appear during perfectly normal file moves and deletes. @butonic already diagnosed all three as expected conditions that are simply logged at the wrong level.

Only **one** of the three sites lives in this repository:

| Log message | Location | Fixable here? |
| --- | --- | --- |
| `could not delete parent id cache` | `services/activitylog/pkg/service/service.go` | yes, fixed in this PR |
| `failed to handle moved away item` | `reva/pkg/storage/fs/posix/tree/assimilation.go` | no, vendored `opencloud-eu/reva` |
| `could not remove lock file` | `reva/pkg/storage/fs/posix/tree/tree.go` | no, vendored `opencloud-eu/reva` |

This PR therefore changes the activitylog site only. The other two need a companion PR in `opencloud-eu/reva` followed by a dependency bump here. I am happy to open that reva PR as a follow-up if you would like.

`ActivitylogService.removeCachedParentID` calls `ttlcache.Cache.Remove`, which returns `ErrNotFound` (`"key not found"`) when the key is not present. The parent id cache is filled lazily in `addActivity` and its entries expire, so on an `ItemMoved` event the key is very often simply not there. That is the normal case, not a failure, so it is now logged at debug level.

A short comment was added at the call site so the reason for the debug level is not lost.

## Related Issue

- Fixes https://github.com/opencloud-eu/opencloud/issues/1043 (partially, see the table above)

Small note on process: @robin-thoene mentioned in the issue that they wanted to pick this up and asked exactly the right two clarifying questions, which turned out to be correct on both counts. As the questions had not been answered and there was no visible activity since, I went ahead and prepared this change. Happy to hand it over or close this if they would still prefer to do it.

## Motivation and Context

Users running an otherwise healthy instance see ERROR entries in their logs for successful operations, which is alarming and hard to distinguish from real problems. The reporter in #1043 specifically opened the issue because they thought something was broken.

## How Has This Been Tested?

- test environment: go1.26.4, `CGO_ENABLED=0`, package `services/activitylog/pkg/service`
- test case 1: `go test -count=1 ./services/activitylog/pkg/service/` passes
- test case 2: three new Ginkgo specs added under `Describe("removeCachedParentID")`:
  - at the default `error` level, a missing cache entry produces no log output at all (this is the actual regression from the issue)
  - at `debug` level, the message is emitted with `"level":"debug"`
  - when the entry does exist, nothing is logged and the entry is removed
- test case 3: verified the new specs genuinely guard the fix by temporarily reverting `.Debug()` back to `.Error()`, which fails 2 of the 3 specs as expected
- test case 4: `go vet ./services/activitylog/...` and `go build ./services/activitylog/...` are clean (`GOOS=linux`)
- test case 5: `gofmt` clean on both changed files

## Screenshots (if appropriate):

n/a

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

---

**AI assistance:** Claude Code helped with investigation and implementation of this fix. I reviewed, tested, and take full responsibility for the change.

Assisted-by: Claude Code:claude-opus-5

### DIFF
--- a/changelog/unreleased/fix-parent-id-cache-log-level.md
+++ b/changelog/unreleased/fix-parent-id-cache-log-level.md
@@ -1,9 +1,0 @@
-Bugfix: Log a missing parent id cache entry at debug level
-
-Moving a file made the activitylog service log "could not delete parent id
-cache" at error level, even though the operation succeeded. The parent id
-cache is filled lazily and its entries expire, so removing a key that was
-never cached returns "key not found", which is the expected case and not an
-error. It is now logged at debug level.
-
-https://github.com/opencloud-eu/opencloud/issues/1043
--- a/changelog/unreleased/fix-parent-id-cache-log-level.md
+++ b/changelog/unreleased/fix-parent-id-cache-log-level.md
@@ -1,0 +1,9 @@
+Bugfix: Log a missing parent id cache entry at debug level
+
+Moving a file made the activitylog service log "could not delete parent id
+cache" at error level, even though the operation succeeded. The parent id
+cache is filled lazily and its entries expire, so removing a key that was
+never cached returns "key not found", which is the expected case and not an
+error. It is now logged at debug level.
+
+https://github.com/opencloud-eu/opencloud/issues/1043
--- a/services/activitylog/pkg/service/service.go
+++ b/services/activitylog/pkg/service/service.go
@@ -661,8 +661,10 @@ func (a *ActivitylogService) removeCachedParentID(ref *provider.Reference) {
 		}
 		purgeId = info.GetId()
 	}
+	// The parent id cache is populated lazily and its entries expire, so a
+	// missing key is the expected case rather than an error.
 	if err := a.parentIdCache.Remove(storagespace.FormatResourceID(purgeId)); err != nil {
-		a.log.Error().Interface("event", ref).Err(err).Msg("could not delete parent id cache")
+		a.log.Debug().Interface("event", ref).Err(err).Msg("could not delete parent id cache")
 	}
 }
 
--- a/services/activitylog/pkg/service/service_test.go
+++ b/services/activitylog/pkg/service/service_test.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"bytes"
 	"context"
 	"net"
 	"os"
@@ -13,8 +14,10 @@ import (
 	nserver "github.com/nats-io/nats-server/v2/server"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/opencloud-eu/opencloud/pkg/log"
 	"github.com/opencloud-eu/opencloud/services/activitylog/pkg/config"
 	eventsmocks "github.com/opencloud-eu/reva/v2/pkg/events/mocks"
+	"github.com/opencloud-eu/reva/v2/pkg/storagespace"
 	"github.com/test-go/testify/mock"
 	"go.opentelemetry.io/otel/trace/noop"
 )
@@ -263,6 +266,42 @@ var _ = Describe("ActivitylogService", func() {
 					g.Expect(activities).To(ConsistOf(activitites("activity1", 0, "activity2", 0, "activity3", 0, "activity4", 0)))
 				}).Should(Succeed())
 			})
+		})
+	})
+
+	Describe("removeCachedParentID", func() {
+		var logBuffer *bytes.Buffer
+
+		newLoggerAtLevel := func(level string) log.Logger {
+			logBuffer = &bytes.Buffer{}
+			return log.Logger{Logger: log.NewLogger(log.Level(level)).Output(logBuffer)}
+		}
+
+		It("does not log an error when the entry was never cached", func() {
+			alog.log = newLoggerAtLevel("error")
+
+			alog.removeCachedParentID(reference("never-cached"))
+
+			Expect(logBuffer.String()).To(BeEmpty())
+		})
+
+		It("logs a missing entry at debug level", func() {
+			alog.log = newLoggerAtLevel("debug")
+
+			alog.removeCachedParentID(reference("never-cached"))
+
+			Expect(logBuffer.String()).To(ContainSubstring("could not delete parent id cache"))
+			Expect(logBuffer.String()).To(ContainSubstring(`"level":"debug"`))
+		})
+
+		It("does not log at all when the entry was cached", func() {
+			alog.log = newLoggerAtLevel("debug")
+			ref := reference("cached")
+			Expect(alog.parentIdCache.Set(storagespace.FormatResourceID(ref.GetResourceId()), resourceID("parent"))).To(Succeed())
+
+			alog.removeCachedParentID(ref)
+
+			Expect(logBuffer.String()).To(BeEmpty())
 		})
 	})
 })


changelog/unreleased/fix-parent-id-cache-log-level.md
rhafer, line 1: Please remove this, we've switched away from `calens` for changelog generation last year already. The `changelog` directory is just a leftover (will be removed by #3339)

rhafer (changes requested):
Looks fine codewise. Just remove the the changelog item please, before we can merge.

Also you might want to read up on https://github.com/opencloud-eu/.github/blob/main/profile/AI_GUIDELINE.md

junkerderprovinz:
Done, the changelog fragment is removed and pushed. Thanks for the pointer to the AI guideline too, I hadn't seen it before, I've added an Assisted-by line to the PR description to match it.

rhafer (approved):

Issue #1128, opened 2025-06-27
Thumbnailer should use a push based mechanism


butonic:
The current thumnailer uses a pull based mechanism that fetches an image via CS3+webdav from an internal space and then generates a thumbnail for it.

This requires the thumbnailer service to make requests to other services, which opens up attack vectors in case it gets compromised. Unfortunately creating thumnails requires interpreting user provided bytes, which is a well known attack vector.

In contrast, the antivirus and search service send a byte stream to clamav / tika and get a corresponding response. This allows locking down any container that runs clamav or tika and forbid them from making any external requests.

Our thumbnailer service should work in the same way and make a plain http call to another container/pod/service.

I would recommend working on https://github.com/cshum/imagor, a *Fast, secure image processing server and Go library, using libvips*. While it uses the [thumbor API](https://thumbor.readthedocs.io/en/latest/usage.html#image-endpoint) it has no way to POST an image: https://github.com/cshum/imagor/issues/383#issuecomment-3012311643 We might help with that.

Imagor can cache images on an S3 bucket and there is also https://github.com/cshum/imagorvideo for video thumbnails.

We can use the https://hub.docker.com/r/shumc/imagor/ docker container, which even supports HEIC.

butonic:
There are altenatives;
- https://github.com/atmos/camo has already been archived
- https://github.com/h2non/imaginary supports POST, last release 2020
- https://github.com/imgproxy/imgproxy, no POST support, discussion in https://github.com/imgproxy/imgproxy/issues/1332
- https://github.com/cshum/imagor supports POST https://github.com/cshum/imagor/issues/383#issuecomment-3012311643

micbar:
> In contrast, the antivirus and search service send a byte stream to clamav / tika and get a corresponding response.

This is not entirely true 😄 These services are downloading the bytes from the dataprovider because they are subscribing to an event.

butonic:
yes, antivirus and search 'download' bytes via cs3, but then they send them to clamav/tika. The latter do not make any requests to the outside.

butonic:
imagor supports push since 2025: https://github.com/cshum/imagor/issues/383#issuecomment-3012311643

tbsbdr:
PB
- 😱 HEIC yes, of course appreciated, still porblematic due to https://github.com/opencloud-eu/opencloud/issues/630#issuecomment-2958154356
- does it add to reliability & easy maintenance?
- @dragotin I'd be interested in your opinion

[...]

PR #3332, opened 2026-08-18
feat(webdav): extract thumbnails through Tika

dschmidt:
Thumbnails for everything the in-process decoders cannot read: raw camera images, audio cover art, and whatever else Tika learns to extract. One generic step, no format-specific code here.

webdav asks Tika for `/unpack/all` and takes the embedded document Tika marked as the document's `THUMBNAIL`. Needs Tika 4.1: in 4.0 cover art is only `INLINE` and raw files come back whole, an 8 MB TIFF instead of a preview.

Rendered previews (a pdf page, the metafile an office document carries) are not part of this. Tika renders only when its parsers are configured to, and that cannot be asked for per request: the old design used a `renderThumbnails` flag on an endpoint of its own, which is not going to exist. Doing it from here would mean either turning on per-request config on the Tika server or a catalog preset, so for now those documents have no thumbnail. A rendering that is there anyway is used, nested under a thumbnail or standing for the document itself, so a preset later needs no change here.

Verified against `apache/tika:4.1.0-SNAPSHOT-full`: mp3 yields its cover as `THUMBNAIL`, a NEF yields a 482 KB JPEG preview and is detected as `image/x-raw-nikon`, and a pdf yields nothing even with `pdfbox-renderer` configured. The rendering paths follow TIKA-4855 and are covered by unit tests, not measured against a real document.

Stacked on #3397.


rhafer (commented):
As much as I'd like to have previews for my raw images (DNG in my case), I'd really prefer us to get not into the business of creating/maintaining a TIFF parser (even with the help of AI). Especially given the experience with the recent security flaws in libvips (which were also TIFF related).

I see your PR to evanoberholster/imagemeta got merged. So using that might help.

And then there is https://github.com/opencloud-eu/opencloud/issues/1128 which suggests to replace the embedded thumbnailer with something like https://github.com/cshum/imagor for better scalability and broader images format support. (Unfortunately AFAICS imagor does not extract the embedded previews though, but tries to render the rawdata)

I am not giving this a 👎 , but I think we should not merge without further  consideration.

@butonic any thoughts on this?

dschmidt:
Thanks for the considerate review, @rhafer :)

One more idea - tika has a concept of embedded files. I've already implemented audio artwork support using that mechanism.
Currently I'm trying to get the jpeg extraction from raw tiff into tika as well: https://github.com/apache/tika/pull/3037

I mainly wanted both extractions to implement the thumbnails relationship on driveItems, but thinking about it in the context of your concerns and the possible move to imagor it makes me think:

We could pipe audio files and raw images from the thumbnails service to tika for preview extraction.
If that works out well enough, it could be a blueprint for piping regular images to imagor (or to whatever image server else).
This would even rid us of the vendored https://github.com/dhowden/tag (which seems dead and I had to patch and fork for https://github.com/opencloud-eu/opencloud/pull/3208).

Having implemented the original audio artwork thumbnails feature, I can say I'd happily drop the built in support for something more maintained like tika. Moreover, we'd use the same mechanism as in the search service and the supported file formats and features would always align (unlike with a completely different code path like right now).

Thoughts?

dschmidt:
I liked my idea so much, that I immediately wanted to give it a spin 😁 

+395 now, the built in approach had +936 - I'm satisfied :D
Main downside, we need to wait for the tika upstream pr and a release :)
... but I'm patient, you know ... ;-)

What do you think?

dschmidt:
Tika PR is merged 🙂

PR #3338, opened 2026-08-18
refactor(search): re-name the engine parameters

fschade:
## Description
we use id as param name all over the place, the pr only changes the naming (is it an id, is it an path lookup) to make it easier to understand what is what.

### DIFF
--- a/services/search/pkg/bleve/batch.go
+++ b/services/search/pkg/bleve/batch.go
@@ -40,14 +40,14 @@ func (b *Batch) Upsert(id string, r search.Resource) error {
 	})
 }
 
-func (b *Batch) Move(id, parentID, location string) error {
+func (b *Batch) Move(id string, parentID string, targetPath string) error {
 	return b.withSizeLimit(func() error {
 		rootResource, err := searchResourceByID(id, b.index)
 		if err != nil {
 			return err
 		}
 		currentPath := rootResource.Path
-		nextPath := utils.MakeRelativePath(location)
+		nextPath := utils.MakeRelativePath(targetPath)
 
 		rootResource.Path = nextPath
 		rootResource.Name = path.Base(nextPath)
--- a/services/search/pkg/bleve/index.go
+++ b/services/search/pkg/bleve/index.go
@@ -102,9 +102,9 @@ func searchResourceByID(id string, index bleve.Index) (*search.Resource, error) 
 	return matchToResource(res.Hits[0]), nil
 }
 
-func searchResourcesByPath(rootId, lookupPath string, index bleve.Index) ([]*search.Resource, error) {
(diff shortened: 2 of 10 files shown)


aduffeck (approved):

services/search/pkg/search/service.go
```diff
@@ -47,11 +47,11 @@ const (
 type Searcher interface {
 	Search(ctx context.Context, req *searchsvc.SearchRequest) (*searchsvc.SearchResponse, error)
 
-	IndexSpace(rID *provider.StorageSpaceId, forceRescan bool) error
+	IndexSpace(spaceID *provider.StorageSpaceId, forceRescan bool) error
 	PurgeDeleted(spaceID *provider.StorageSpaceId) error
 
 	TrashItem(rID *provider.ResourceId)
```
dschmidt: Why did you keep the rID name here?

services/search/pkg/opensearch/batch.go
```diff
@@ -63,7 +63,7 @@ func (b *Batch) Upsert(id string, r search.Resource) error {
 	})
 }
 
-func (b *Batch) Move(id, parentID, location string) error {
+func (b *Batch) Move(id, parentID, targetPath string) error {
```
dschmidt:
```suggestion
func (b *Batch) Move(id string, parentID string, targetPath string) error {
```

services/search/pkg/bleve/index.go
```diff
@@ -102,9 +102,9 @@ func searchResourceByID(id string, index bleve.Index) (*search.Resource, error)
 	return matchToResource(res.Hits[0]), nil
 }
 
-func searchResourcesByPath(rootId, lookupPath string, index bleve.Index) ([]*search.Resource, error) {
+func searchResourcesByPath(rootID, lookupPath string, index bleve.Index) ([]*search.Resource, error) {
```
dschmidt:
```suggestion
func searchResourcesByPath(rootID string, lookupPath string, index bleve.Index) ([]*search.Resource, error) {
```

services/search/pkg/bleve/batch.go
```diff
@@ -40,14 +40,14 @@ func (b *Batch) Upsert(id string, r search.Resource) error {
 	})
 }
 
-func (b *Batch) Move(id, parentID, location string) error {
+func (b *Batch) Move(id, parentID, targetPath string) error {
```
dschmidt:
```suggestion
func (b *Batch) Move(id string, parentID string, targetPath string) error {
```

dschmidt (approved):
consistency ftw!

In one of the last hunks you added `string` to each param, you didn't for others... I guess we should stick to one approach. No idea if that's consistent in the code base otherwise...

PR #3343, opened 2026-08-18
fix(search): keep the case of a path in the index

fschade:
### openSearch index

livecycle update for out openSearch index,
it removes the path lowercase filter which fixes an issues which appears when searching for folders!
It also adds strict mappings for Tags & Mtime.

### code changes

there was a bug that incorrectly evaluated the hidden state, this is fixed too.

There is no auto migration, all necessary steps are described in the new migration guide (FYI @Svanvith)

## How Has This Been Tested?
- manually in my lab 
- tests

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/search/pkg/search/service.go
+++ b/services/search/pkg/search/service.go
@@ -616,7 +616,13 @@ func (s *Service) doUpsertItem(ref *provider.Reference, batch BatchOperator) {
 		Type:     uint64(stat.Info.Type),
 		Document: doc,
 	}
-	r.Hidden = strings.HasPrefix(r.Path, ".")
+
+	for name := range strings.SplitSeq(filepath.Clean(r.Path), string(filepath.Separator)) {
+		if name != "." && strings.HasPrefix(name, ".") {
+			r.Hidden = true
+			break
+		}
+	}
 
 	if parentID := stat.GetInfo().GetParentId(); parentID != nil {
 		r.ParentID = storagespace.FormatResourceID(parentID)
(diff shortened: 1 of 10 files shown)


services/search/pkg/search/service.go
```diff
@@ -616,7 +616,13 @@ func (s *Service) doUpsertItem(ref *provider.Reference, batch BatchOperator) {
...
 		Document: doc,
 	}
-	r.Hidden = strings.HasPrefix(r.Path, ".")
+
+	for name := range strings.SplitSeq(filepath.Clean(r.Path), string(filepath.Separator)) {
+		if name != "." && strings.HasPrefix(name, ".") {
+			r.Hidden = true
+			break
+		}
+	}
```
dschmidt, line 625: This makes everything hidden inside a dot-folder, right? 
Great, I guess!

dschmidt (approved):

dschmidt (changes requested):
The new segment-based check makes Hidden depend on the whole path, but engine.Move (both bleve and opensearch) only rewrites Path/Name/ParentID and never touches Hidden. So a rename like foo.txt -> .foo.txt, or moving a subtree out of a dot folder (.stuff/notes/ -> /notes/), leaves stale Hidden values for the moved item and all its descendants until they happen to be upserted again.

This was invisible before because everything was indexed as hidden anyway, but with this fix it becomes observable. Fine as a follow-up in my opinion, but Move should probably recompute Hidden from the new path, and a test case for the rename-to-dotfile scenario would be nice.

PR #3344, opened 2026-08-18
actually log error on exit

butonic:
exiting withoul logging is not nice ... fmt.Errorf ... does not log the error, it just renders the string and returns it ...

### DIFF
--- a/services/antivirus/pkg/command/server.go
+++ b/services/antivirus/pkg/command/server.go
@@ -43,7 +43,7 @@ func Server(cfg *config.Config) *cobra.Command {
 			{
 				svc, err := service.NewAntivirus(cfg, logger, traceProvider)
 				if err != nil {
-					fmt.Errorf("failed to initialize antivirus service: %v", err)
+					logger.Error().Err(err).Str("server", "antivirus").Msg("Failed to initialize antivirus service")
 					os.Exit(1)
 				}
 


pbleser-oc (approved):
lgtm

dschmidt:
Since #3140 you can simply return the error instead:

```go
if err != nil {
    return fmt.Errorf("failed to initialize antivirus service: %w", err)
}
```

Supervised mode: the service runs in-process (`runtime/service/service.go` calls `antivirus.Execute` directly), so `os.Exit(1)` kills the whole opencloud process and bypasses suture's restart logic. Returning the error lets the supervisor log it structured via `Err()`.

Standalone mode: since #3140 cobra is silenced, `main()` prints the error to stderr and exits 1, so nothing is lost there either.

Also matches the debug server block below, which already returns its error. The `Str("server", "antivirus")` field would be redundant anyway, the logger already carries the service name.

dschmidt:
Sorry for disabling auto-merge, I would like you to look at my comment before it's merged 🙈 

If you disagree, just reenable auto-merge/click merge

edit:
The proper way would have probably been to request a change, but then it's more annoying to overrule than just reenabling auto-merge

butonic:
@dschmidt Absolutely true. However, this is not the only place where we exit(1) 😞 I'll create an issue to track that.

PR #3350, opened 2026-08-19
fix(search): analyze Content queries on OpenSearch (match_phrase instead of term)

dschmidt:
`Content` is an analyzed text field, but single-term KQL queries compiled to a `term` query, which matches the raw value against tokenized content and never hits: `Content:https://opencloud.eu/` returned 0 (the content tokenizes to `[https, opencloud.eu, ...]`). Bleve analyzes the query, so this only affected OpenSearch.

Route full-text fields to `match_phrase` so the query is analyzed like the content. Keyword fields keep `term`.


dschmidt:
This is solved better with #3345 , I was just trying to get #3142 green - but it probably makes more sense to wait for #3345.

PR #3355, opened 2026-08-19
feat(graph): add LibreGraphContentType on drive

JammingBen:
See https://github.com/opencloud-eu/web/issues/3029 and https://github.com/opencloud-eu/libre-graph-api/pull/66


services/graph/pkg/service/v0/graph_test.go
```diff
@@ -802,6 +802,7 @@ var _ = Describe("Graph", func() {
 							Map: map[string]*typesv1beta1.OpaqueEntry{
 								"description": {Decoder: "plain", Value: []byte("This space is for testing")},
 								"spaceAlias":  {Decoder: "plain", Value: []byte("project/testspace")},
+								"contentType": {Decoder: "plain", Value: []byte("com.example.test")},
```
rhafer: not that it matters much for the result of the test, but we wanted the values to be somewhat like mimetypes, could you actually use a test string here that looks more like a mimetype?

rhafer (changes requested):
Small, nit. Rest LGTM.

rhafer (approved):

PR #3356, opened 2026-08-19
fix(graph): adding the same user as multiple members in a group (#3354)

### LINKED ISSUE #3354: Adding the same user as multiple members in a single request results in listing the same user twice in the group
## Describe the bug
When adding users to a group using the Graph API, and referencing the same user ID multiple times in the member list, that user ends up being in that group twice.

> [!IMPORTANT]
> Note that this only affects the built-in IDM (see https://github.com/libregraph/idm/issues/209), as OpenLDAP treats those LDAP modify operations as errors instead.

## Steps to reproduce
Setting a few environment variables to make copy-pasting the snippets below more portable:
```bash
groupname="test"
api="https://localhost:9200"
auth="admin:admin"
```
The snippets below require having `curl` and `jq`, optionally `shuf` to automate picking a random user, as well as basic authentication support (`PROXY_ENABLE_BASIC_AUTH=true`) to simplify the code below.

1. Optionally create a new group
```bash
curl -sSk -u "$auth" -X POST "$api/graph/v1.0/groups" -d '{"displayName": "'"$groupname"'"}'
```

Expected output:
```json
{"displayName":"test","groupTypes":[],"id":"..."}
```

2. Pick an existing user at random
```bash
id=$(curl -ksS -u "$auth" -X GET "$api/graph/v1.0/users" | jq -r '.value.[] | select(.accountEnabled == true and .onPremisesSamAccountName != "admin") | .id' | shuf -n 1)
```

3. Add that user twice to a group
```bash
gid=$(curl -ksS -u "$auth" -X GET "$api/graph/v1.0/groups" | jq -r '.value.[] | select(.displayName == "test") | .id')

curl -ksS -u "$auth" -X PATCH "$api/graph/v1.0/groups/$gid" -d '{"members@odata.bind": ['"\"/graph/v1.0/users/$id\""','"\"/graph/v1.0/users/$id\""']}'
```

4. Check the members of that group:
```bash
curl -ksS -u "$auth" -X GET "$api/graph/v1.0/groups/$gid/members" | jq -r '.[].id'
```

Expected (buggy) output being the same user ID twice, for example:
```text
801abee4-dd7f-11ef-a324-83f55a754b62
801abee4-dd7f-11ef-a324-83f55a754b62
```

## Expected behavior
Normalize the list of member IDs to skip duplicates, as it makes no sense to have the same user being a member of a group twice. Ideally, this should also be caught on the IDM side of things, but as part of data sanitization, the Graph API should catch that as well.

When using the snippets above, the user should only show up once in the list of members of the group, instead of twice.

## Actual behavior
The user is listed twice, and is thus stored twice as a member of that group.

With the snippets above, the expected output of the group member listing code (point 4) would be the user being referenced once, not twice, for example
```text
801abee4-dd7f-11ef-a324-83f55a754b62
```

## Setup
No particular setup is necessary (except allowing basic authentication to simplify the code snippets above to reproduce the issue).

## Additional context
Currently a known issue that is covered by [`apiGraphUserGroup/addUserToGroup.feature:429`](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L429) and marked as a known-to-fail test.

Also a known issue in the oci repo: https://github.com/owncloud/ocis/issues/5855

pbleser-oc:
## Description
* make sure the list of members to add to a group is unique, both on the level of the inbound list of member IDs as on the level of the LDAP DNs they resolve to
* remove the corresponding failing test from the list of expected failures

## Related Issue
- Fixes #3354

## Motivation and Context
When users are added to an existing group, specifying the same user identifier multiple times ended up in that user entry being referenced multiple times in that group, in the IDM as well as in Graph API calls.

That should not be the case as it makes no sense to have the same user referenced multiple times in a group.

## How Has This Been Tested?
Test snippets can be found in #3354 

Used those snippets on a locally running instance.

Also verified using the test suite:
```bash
BEHAT_FEATURE=tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature:429 \
TEST_SERVER_URL=https://host.docker.internal:9200 \
make -C tests/acceptance/docker/ run-test-only
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/graph/pkg/identity/ldap_group.go
+++ b/services/graph/pkg/identity/ldap_group.go
@@ -293,6 +293,26 @@ func (i *LDAP) AddMembersToGroup(ctx context.Context, groupID string, memberIDs 
 	if !i.writeEnabled && i.groupCreateBaseDN == i.groupBaseDN {
 		return errorcode.New(errorcode.NotAllowed, "server is configured read-only")
 	}
+
+	// eliminate duplicates in memberIDs by their UUID, to prevent unnecessary LDAP
+	// lookups below, and for https://github.com/opencloud-eu/opencloud/issues/3354
+	{
+		uniqueMemberIDs := make([]string, 0, len(memberIDs))
+		memory := map[string]struct{}{}
+		for _, id := range memberIDs {
+			if len(strings.TrimSpace(id)) == 0 {
+				continue // while we're at it, clean the input parameters and skip empty strings
+			}
+			if _, present := memory[id]; !present {
+				uniqueMemberIDs = append(uniqueMemberIDs, id)
+				memory[id] = struct{}{}
+			} else {
+				logger.Debug().Str("id", id).Msg("Duplicate member in group addition. Skipping")
+			}
+		}
+		memberIDs = uniqueMemberIDs
+	}
+
 	ge, err := i.getLDAPGroupByNameOrID(groupID, true)
 	if err != nil {
 		return err
--- a/tests/acceptance/expected-failures-decomposed-storage.md
+++ b/tests/acceptance/expected-failures-decomposed-storage.md
@@ -56,10 +56,6 @@
 - [apiGraphUserGroup/addUserToGroup.feature:378](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L378)
 - [apiGraphUserGroup/addUserToGroup.feature:392](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L392)
 
-#### [Adding the same user as multiple members in a single request results in listing the same user twice in the group](https://github.com/owncloud/ocis/issues/5855)
-
-- [apiGraphUserGroup/addUserToGroup.feature:429](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiGraphUserGroup/addUserToGroup.feature#L429)
-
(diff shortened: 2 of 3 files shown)


services/graph/pkg/identity/ldap_group.go
```diff
@@ -326,20 +346,28 @@ func (i *LDAP) AddMembersToGroup(ctx context.Context, groupID string, memberIDs
...
+			if _, present := currentSet[nDN]; !present {
+				if _, duplicate := memory[nDN]; !duplicate {
+					newMemberDN = append(newMemberDN, me.DN)
+					memory[nDN] = struct{}{}
+				} else {
+					logger.Debug().Str("memberDN", me.DN).Msg("Duplicate member in group addition. Skipping")
+				}
+			} else {
+				logger.Debug().Str("memberDN", me.DN).Msg("Member already present in group. Skipping")
+			}
```
rhafer: I theory this whole block shouldn't be needed as LDAP doesn't allow duplicate attribute values to be added. The fact that this works with the builtin idm is a short coming of https://github.com/libregraph/idm and should be addressed there I think.

Are you ok with removing this and just keep the duplicate uuid check?

pbleser-oc: Makes sense, I opened https://github.com/libregraph/idm/issues/209 to track that issue there, and I will remove this check, only keeping the duplicate uuid sanitization.

rhafer (approved):

PR #3362, opened 2026-08-19
feat(thumbnails): Tika-based audio cover art extraction

dschmidt:
Stacked on #3332, which lands the shared Tika unpack helper and the generic `ErrNoEmbeddedImage`.

Route audio cover art through Tika, so all embedded-media extraction can eventually run through Tika (and `dhowden/tag` be dropped).

When a Tika server is configured, cover art is extracted via Tika by default: it selects the tagged front cover (ID3 APIC "Cover (front)") via `/unpack/all` sidecar metadata, else the first embedded image, matching the deterministic selection from #3208. Verified byte-identical to the in-process reader on real files (front-cover and untyped). Without a Tika server it stays in-process.

`THUMBNAILS_PREPROCESSOR_AUDIO_PROCESSOR=builtin` forces the in-process extractor even when a Tika server is configured; that `builtin` path is deprecated and slated for removal in the next major.

`TikaDecoder` picks by mime type: `audio/*` front cover, everything else the largest renderable preview.


dschmidt:
If we agree on the Tika approach, this supersedes https://github.com/opencloud-eu/opencloud/pull/3208 :)

services/thumbnails/pkg/config/config.go
```diff
@@ -53,6 +53,8 @@ type Thumbnail struct {
 // thumbnailable image.
 type Preprocessor struct {
 	Tika Tika `yaml:"tika"`
+	// AudioProcessor selects the audio cover-art extractor.
+	AudioProcessor string `yaml:"audio_processor" env:"THUMBNAILS_PREPROCESSOR_AUDIO_PROCESSOR" desc:"The processor for extracting cover art from audio files. When a Tika server is configured it is used by default. Set this to 'builtin' to force the in-process extractor even when a Tika server is configured. The 'builtin' processor is deprecated and will be removed in a future major release." introductionVersion:"%%NEXT%%" deprecationVersion:"%%NEXT%%" removalVersion:"%%NEXT_PRODUCTION_VERSION%%" deprecationInfo:"THUMBNAILS_PREPROCESSOR_AUDIO_PROCESSOR only selects the deprecated in-process audio extractor; once it is removed audio cover art always uses Tika when configured." deprecationReplacement:""`
```
dschmidt, line 57: Not sure we need this - one could argue, Tika is opt-in and when you choose to opt in, you get audio extraction via tika.

Thoughts?

dschmidt:
Superseded by #3332, which takes the cover art (and every other thumbnail Tika provides) through one generic Tika step on top of #3397. The branch is kept as `feat/thumbnails-audio-tika-legacy`.

PR #3366, opened 2026-08-20
docs: clarify custom role bootstrap behavior

Svanvith:
## Description
Clarify how custom roles are initialized and updated via `SETTINGS_BUNDLES_PATH`.

Document that default roles can only be prevented from being created when the variable is configured before the first OpenCloud start. On subsequent starts, configured roles are added or updated, while existing roles are not deleted to avoid dangling role assignments.

Also fix the incorrect `SETTINGS_BUNDLE_PATH` variable name in the example.


rhafer (approved):
👍

PR #3383, opened 2026-08-20
[do-not-merge] wip: run search tests suits against both engines

fschade:
## Description
DO NOT MERGE IT, the only reason for it is to see if both engines are green in the ci now!


fschade:
search related tests are green now 🥳 
i close the PR now and start to split it into smaller chunks

PR #3408, opened 2026-08-26
fix(search): make openSearch and bleve behave the same

fschade:
## Description

since we have support for bleve and openSearch as search backends, we also have to take care that those engines behave the same (at least in all basic cases), which was not the case ash shown in the table.

| Query                       | before bleve  | before OpenSearch     | both now   |
|-----------------------------|---------------|-----------------------|-----------------------|
| `new`                       | -             | `new-folder`          | `new-folder`          |
| `report`                    | -             | -                     | `Report.txt`          |
| `name:new`                  | -             | `new-folder`          | `new-folder`          |
| `name:"new"`                | -             | `new-folder`          | -                     |
| `name:"*report"`            | -             | -                     | `Report.txt`          |
| `name:"*übung*"`            | `Übung.txt`   | -                     | `Übung.txt`           |
| `Title:quarterly`           | -             | `q1.html`             | `q1.html`             |
| `Title:QUARTERLY`           | -             | `q1.html`             | `q1.html`             |
| `title:quarterly`           | -             | -                     | `q1.html`             |
| `Content:report`            | `monthly.txt` | -                     | -                     |
| `Content:"reports monthly"` | `monthly.txt` | -                     | -                     |
| `mediatype:TEXT/MARKDOWN`   | -             | `notes.md`            | `notes.md`            |
| `path:"./parent"`           | nur `parent`  | `parent`, `child.pdf` | `parent`, `child.pdf` |
| `type:file`                 | -             | **error 400**        | jede Datei            |
| `type:folder`               | -             | **error 400**        | jeder Ordner          |
| `size>100`                  | -             | -                     | filtert nach Größe    |
| `hidden:true`               | -             | `hidden.txt`          | `hidden.txt`          |
| `Favorites:"A1B2-Upper"`    | `starred.txt` | -                     | `starred.txt`         |

or more detailed [here](https://github.com/fschade/opencloud/blob/1c6b40883f914d19ffd18e115a77d6b2cbfe04fc/services/search/pkg/parity/README.md
)

for that reason, the pr contains a new test suite which runs the same tests against each search backend which proves the compatibility between the two.

The suite produces a markdown document at the end which lists the test cases and shows the similarity between bleve and openSearch.

while writing that engine i stumbled across some inconsistencies which are fixed now too, in a nutshell:

- field matching on OpenSearch (name, title, tags, path), boolean and negated group queries on bleve, cased values (type:File, hidden:TRUE, ids), trailing slash in paths, ? as wildcard, missing favorite: alias, a broken mimetype entry
- KQL: quoted strings are marked in the AST, number restrictions (size>100) are parsed as numbers
- Index names carry a generation (opencloud-resource-v3, bleve-v2): a changed mapping starts on a fresh index instead of failing, filling it is manual
- Some new acceptance tests

unfortunately all that also means, we have to create a new index mapping for openSearch but also for bleve, so the PR can be considered breaking and should not to into the next minor version. For details please read [services/search/MIGRATION.md](https://github.com/fschade/opencloud/blob/1c6b40883f914d19ffd18e115a77d6b2cbfe04fc/services/search/MIGRATION.md).

it also acts as a transition point to:

- https://github.com/opencloud-eu/opencloud/pull/3345
- https://github.com/opencloud-eu/opencloud/pull/3197

## Motivation and Context
Searching for a name, a tag or inside a folder returned wrong or no results.

## How Has This Been Tested?
- unit & acceptance tests
- lab installation

## Types of changes
- [X] Bug fix
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [X] Unit tests added
- [X] Acceptance tests added
- [X] Documentation added


services/search/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
+# Migrating the OpenSearch index
+
+Our index definition is versioned. The service does not start on an index that
+was built with an older one. A new installation is not affected.
+
+There are two ways out.
+
+> `$OS` (the OpenSearch address), `opencloud-resource` (default) index name.
```
butonic: I cannot parse this line ... can you make it more explicit? what does `$OS` have to do with this?!?

dschmidt (dismissed):
LGTM

services/search/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+done.
+
+Writing to a missing index creates one without our analyzer and mappings, so
+opencloud has to be down while it is gone.
+
+```shell
+# stop opencloud
+curl -XDELETE "$OS/opencloud-resource"
+# start opencloud, it creates the index from the current definition
+opencloud search index --all-spaces
```
butonic: do I get that right: if between these two lines a new file is added in the web ui ... It will create an index 
> without our analyzer and mappings, ...

and the index needs to be deleted again? How does this relate to line 17-18?

services/search/MIGRATION.md
```diff
@@ -0,0 +1,52 @@
...
+               "Mtime":{"type":"date","ignore_malformed":true}}}'
+
+# start opencloud
+
+curl -XPOST "$OS/opencloud-resource/_update_by_query?conflicts=proceed&wait_for_completion=false"
+```
+
+The last command answers with a task id, watch it with `GET $OS/_tasks/<task-id>`.
+A million documents take a few minutes. Search keeps working while it runs,
+documents get correct one by one, and the run can be repeated at any time.
```
butonic: > [...] Search keeps working while it runs,
documents get correct one by one, and the run can be repeated at any time.

So I can implement a migration for kubernetes that runs while opencloud is running? It just can't index new documents while the opencloud-resource index is closed? That contradicts line 32:
```
# stop opencloud
```
why?

services/search/README.md
```diff
@@ -55,6 +55,14 @@ Additionally, the following optional settings can be set:
 *   `SEARCH_ENGINE_OPEN_SEARCH_CLIENT_ENABLE_DEBUG_LOGGER=val`: Enable debug logging.
 *   `SEARCH_ENGINE_OPEN_SEARCH_CLIENT_INSECURE=val`: Skip TLS certificate verification.
 
+### Bringing an existing index over
+
+A version can bring a new index definition. The service does not start on an index that was built with an
+older one and says so in the error. A new installation is not affected.
```
butonic: This really is what bothers me. It prevents pods from starting up. So I cannot upgrade the image first and then do a migration of the index in the background. Why not? Why do we prevent that? Why does the search process have to DIE and kill the whole process, just because the index is not correct. We should degrade gracefully and queue the search events .... which we do, because we have nats. we could just tune the search queue to keep more than 1000 events. That would allow locking the index and running the service in a degraded state ...

services/search/MIGRATION.md
butonic: Ah ... {$OS} as in `{{host}}`.

services/search/README.md
fschade: migration handling will be part of @dschmidt PR as a followup. its a new index now, so problem should be solved for the moment.

services/search/internal/opensearchtest/suite.go
```diff
@@ -67,7 +67,6 @@ func setupOpenSearchTestContainer(ctx context.Context, cfg *config.Config) (func
 
 	keepContainer := os.Getenv("KEEP_TEST_CONTAINER") == "true"
 	if keepContainer {
-		// the reaper would take the kept container down with the session
```
butonic, line 70: why drop all these comments?

services/search/internal/opensearchtest/suite.go
```diff
@@ -83,14 +82,6 @@ func setupOpenSearchTestContainer(ctx context.Context, cfg *config.Config) (func
 		opensearch.WithPassword(cfg.Engine.OpenSearch.Client.Password),
 		testcontainers.WithName(containerName),
 		testcontainers.WithReuseByName(containerName),
-		// test indexes are tiny; don't let a full host disk trip the flood-stage
-		// create-index / read-only blocks mid-run
-		testcontainers.WithEnv(map[string]string{
-			"cluster.routing.allocation.disk.threshold_enabled": "false",
-		}),
```
butonic, line 90: should this really be dropped?

services/search/pkg/opensearch/index.go
```diff
@@ -16,16 +17,39 @@ import (
 
 var (
 	ErrManualActionRequired                  = errors.New("manual action required")
-	IndexManagerLatest                       = IndexIndexManagerResourceV2
-	IndexIndexManagerResourceV1 IndexManager = "resource_v1.json"
-	IndexIndexManagerResourceV2 IndexManager = "resource_v2.json"
+	IndexManagerLatest                       = IndexIndexManagerResourceV3
+	IndexIndexManagerResourceV3 IndexManager = "resource_v3.json"
```
butonic, line 21: Ok, this is confusing me. was IndexIndexManagerResourceV1  just a leftover?

services/search/internal/opensearchtest/suite.go
dschmidt: I had just introduced it in my pr, it helps with my rather full disk - I would really like to keep it :)

services/search/pkg/opensearch/index.go
dschmidt: Yes, it's going away in my branch as well (together with resource_v2 (and with this merged also v3).

We weren't doing anything useful with those revisions...

fschade: yup leftover, not in use anymore

butonic: yup, just leftover. we always compare the currently online index with latest

butonic (approved):

PR #3420, opened 2026-08-29
fix(search): ignore invalid photo exposure times

### LINKED ISSUE #3267: json: unsupported value: +Inf when indexing files with embedded images that have ExposureTime=0
getPhoto() in tika.go divides by the parsed exif:ExposureTime value without checking for zero: https://github.com/opencloud-eu/opencloud/blob/1c138efac2255518b50f111c0046cb737b75a5c2/services/search/pkg/content/tika.go#L222

When i == 0, this produces +Inf, which causes json.Marshal to fail:

failed to marshal resource: json: error calling MarshalJSON for type *libregraph.Photo: json: unsupported value: +Inf

This happens with files containing embedded images that have default/zeroed EXIF data (e.g. ONLYOFFICE sample PPTX files, where Tika returns exif:ExposureTime: "0.0" for all 8 embedded placeholder images).

The affected file never gets indexed.

fredrikblau:
## Summary

Apache Tika can report `exif:ExposureTime` as `0.0` for embedded placeholder images. Dividing by that value creates `+Inf`, which later makes the Graph photo JSON impossible to marshal and prevents the containing file from being indexed.

Ignore non-positive and non-finite exposure values while preserving valid photo metadata.

## Testing

- `gofmt -w services/search/pkg/content/tika.go services/search/pkg/content/tika_test.go`
- `go test ./services/search/pkg/content`
- `git diff --check`

Fixes #3267

### DIFF
--- a/services/search/pkg/content/tika_photo.go
+++ b/services/search/pkg/content/tika_photo.go
@@ -63,10 +63,12 @@ func (t Tika) getPhoto(meta map[string][]string) *libregraph.Photo {
 	}
 
 	if v, err := getFirstValue(meta, "exif:ExposureTime"); err == nil {
-		if i, err := strconv.ParseFloat(v, 64); err == nil && i > 0 {
-			initPhoto()
-			photo.SetExposureNumerator(1)
-			photo.SetExposureDenominator(math.Round(1 / i))
+		if i, err := strconv.ParseFloat(v, 64); err == nil {
+			if d := math.Round(1 / i); !math.IsNaN(d) && !math.IsInf(d, 0) && d > 0 {
+				initPhoto()
+				photo.SetExposureNumerator(1)
+				photo.SetExposureDenominator(d)
+			}
 		}
 	}
 
--- a/services/search/pkg/content/tika_test.go
+++ b/services/search/pkg/content/tika_test.go
@@ -146,14 +146,16 @@ var _ = Describe("Tika", func() {
 		})
 
 		It("ignores zero exposure time", func() {
-			fullResponse = `[{"exif:ExposureTime": "0.0"}]`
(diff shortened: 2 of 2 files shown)


butonic:
@dschmidt can I get your review on this?

dschmidt:
Thanks, good catch!

One suggestion: I'd guard the computed denominator instead of the input and just skip the block instead of returning early. `ExposureTime` happens to be the last field today, but if someone adds another field below, the early return would silently drop it. Checking the result also covers the subnormal case where `i > 0` but `1/i` still overflows to `+Inf`:

```go
if v, err := getFirstValue(meta, "exif:ExposureTime"); err == nil {
	if i, err := strconv.ParseFloat(v, 64); err == nil {
		if d := math.Round(1 / i); !math.IsNaN(d) && !math.IsInf(d, 0) && d > 0 {
			initPhoto()
			photo.SetExposureNumerator(1)
			photo.SetExposureDenominator(d)
		}
	}
}
```

That's one check for all the broken inputs (0, negative, NaN, subnormal) and no early return. Otherwise looks good to me.

dschmidt:
I'm afraid my merge of https://github.com/opencloud-eu/opencloud/pull/3198 caused a merge conflict here, can you rebase please? :)

fredrikblau:
Rebased onto current  and resolved the conflict. The exposure-denominator guard and regression test are preserved. ok  	github.com/opencloud-eu/opencloud/services/search/pkg/content	(cached) and  pass locally.

fredrikblau:
Rebased onto current main and resolved the conflict. The exposure-denominator guard and regression test are preserved. go test ./services/search/pkg/content and git diff --check pass locally.

dschmidt (approved):
LGTM, thanks! :)

PR #3423, opened 2026-08-31
test(search): fail the parity suite when the committed matrix is stale

dschmidt:
The parity suite regenerated its README.md on every run, but nothing compared the result with the committed file: CI silently rewrote and discarded it, so the table could go stale while everything stayed green.

Now a plain run compares the generated matrix with the committed README.md and fails when they differ, without touching the working tree; `UPDATE_MATRIX=true go test ./services/search/pkg/parity/` regenerates the file, committing it is the fix.

Verified all three paths against a live OpenSearch: clean tree passes, a doctored README fails with `README.md is out of date: regenerate it with UPDATE_MATRIX=true and commit the change` (file untouched), and the update run rewrites it.

### DIFF
--- a/services/search/pkg/parity/matrix_test.go
+++ b/services/search/pkg/parity/matrix_test.go
@@ -171,7 +171,7 @@ func writeMatrix(report types.Report) {
 
 	out := &strings.Builder{}
 	out.WriteString("# Engine parity\n\n")
-	out.WriteString("Written by the parity suite (`UPDATE_MATRIX=true go test ./services/search/pkg/parity/`), do not edit.\n")
+	out.WriteString("Written by the parity suite (`UPDATE_SEARCH_PARITY_MATRIX=true go test ./services/search/pkg/parity/`), do not edit.\n")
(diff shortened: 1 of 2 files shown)


services/search/pkg/parity/matrix_test.go
```diff
@@ -171,7 +171,7 @@ func writeMatrix(report types.Report) {
 
 	out := &strings.Builder{}
 	out.WriteString("# Engine parity\n\n")
-	out.WriteString("Written by the parity suite (`go test ./services/search/pkg/parity/`), do not edit.\n")
+	out.WriteString("Written by the parity suite (`UPDATE_MATRIX=true go test ./services/search/pkg/parity/`), do not edit.\n")
```
fschade: can you please rename `UPDATE_MATRIX` to `UPDATE_SEARCH_PARITY_MATRIX`

dschmidt: done

fschade (approved):
LGTM

PR #3424, opened 2026-08-31
feat(search): scope searches to a drive via the driveId field

dschmidt:
> Replaces #3299, which GitHub auto-closed when its base branch `refactor/search-mapping` was deleted on the #3345 merge. Same two commits, rebased onto `main`, plus parity pins (SCOPE-06..09).

`scope:` takes an opaque resource id, which is hostile to hand-written queries. Accept `driveId:"<storage$space>"` as a regular KQL field instead: it resolves to the indexed `RootID` and a bare drive id is completed to the root resource id. Combined with `path:` this gives a readable location scope with no token stripping, both compose with groups, OR and NOT like any other field. MS KQL has no driveId property (it scopes via `path:` or the SharePoint ListId/SiteId managed properties, and exposes driveId only as response metadata); the field is named after the Graph `parentReference.driveId` our clients actually see.

Second commit prunes the space fan-out when top-level AND conjuncts pin the query to a single root; conservative rules, the restriction stays in the query, mountpoints are kept for path mapping.

Follow-up (deliberately not here): parse the query once in the service and hand the AST to the engines, which currently re-parse per space; that changes the engine interface.

### DIFF
--- a/services/search/pkg/parity/README.md
+++ b/services/search/pkg/parity/README.md
@@ -378,10 +378,10 @@ Fixtures:
 
 Fixtures:
 
-- `parent`, folder
-- `child.pdf`, Path = ./parent/child.pdf
-- `outside.txt`
-- `elsewhere.txt`
+- `parent`, ID = 1$1!parent, folder
+- `child.pdf`, ID = 1$1!child.pdf, Path = ./parent/child.pdf
+- `outside.txt`, ID = 1$1!outside.txt
+- `elsewhere.txt`, ID = 1$1!elsewhere.txt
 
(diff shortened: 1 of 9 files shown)


butonic (approved):
well, ok ... so you are adding an alias ... but KQL queries are not really hand written in the first place.

dschmidt:
true, but even if it's only for developers I want to establish the mental model that "search fields" match "driveItem properties" as closely as possible - and usually we're dealing with driveIds and not rootIds (the only occurence of rootId I found in the libre graph spec was for remoteItems)

thanks for the review :)

PR #3432, opened 2026-08-31
chore(mise): determine go version in mise.toml by parsing go.mod

pbleser-oc:
## Description
Instead of statically defining the version of Go in `mise.toml` as well as in its native file (`go.mod`), use a small shell script to parse go.mod and determine the version dynamically from there.

That version is then available as an env variable in the `mise.toml` file and can be referenced for go, in order to avoid having to keep those versions in sync, especially since `dependabot` will update go in the `go.mod` file, but not in `mise.toml`.

Note that this only impacts developers who use [`mise`](https://mise.jdx.dev/), and has no effect on building, or the CI, nor on the runtime or users.

## Types of changes
- [X] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/.mise-go-version.sh
+++ b/.mise-go-version.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+# Purpose of this script is to parse the go.mod file to retrieve the version of
+# Go that we are using, either from a "toolchain" directive (preferred), or from
+# a "go" directive as a fallback.
+#
+# The script then outputs that version, where it will be picked up as an environment
+# variable in mise.
+
+F="./go.mod"
+
+AWK=awk
+# prefer 'gawk' over 'awk' to make sure we get GNU awk
+command -v gawk &>/dev/null && AWK=gawk
+
+[[ -e $F ]] || { echo "ERROR: failed to find $F" >&2; exit 1;}
+
+GO_VERSION=$("$AWK" '/^toolchain/ {print $2; exit} /^go / {print $2}' "$F")
+GO_VERSION=${GO_VERSION#go} # strip the potential 'go' prefix:
+echo "${GO_VERSION}"
--- a/mise.toml
+++ b/mise.toml
@@ -1,5 +1,8 @@
+[env]
+GO_VERSION = "{{ exec(command='./.mise-go-version.sh') }}"
+
 [tools]
-go = "1.25.9"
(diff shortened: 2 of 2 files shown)


.mise-pnpm-version.sh
```diff
@@ -0,0 +1,18 @@
...
+
+F="./services/idp/package.json"
+
+AWK=awk
+# prefer 'gawk' over 'awk' to make sure we get GNU awk
+command -v gawk &>/dev/null && AWK=gawk
+
+[[ -e $F ]] || { echo "ERROR: failed to find $F" >&2; exit 1;}
+
+exec "$AWK" 'match($0, /^ *"packageManager" *: *"pnpm@(.+)"/, m) {print m[1]}' "$F"
```
fschade:
```suggestion
[[ -e $F ]] || { echo "ERROR: failed to find $F" >&2; exit 1;}

exec sed -n 's/.*"pnpm@\([^"]*\)".*/\1/p' "$F"
```

`match(a, b, c)` (3-arg) fails on macOS awk (BSD), the above worked for me, e.g.

```bash
➜ PATH=/usr/bin:/bin awk 'match($0, /^ *"packageManager" *: *"pnpm@(.+)"/, m){print m[1]}' services/idp/package.json
awk: syntax error at source line 1
 context is
        match($0, /^ *"packageManager" *: >>>  *"pnpm@(.+)"/, <<< 
awk: bailing out at source line 1
```

fschade (changes requested):
clever, thanks! one little nitpick on macos

.mise-pnpm-version.sh
fschade: one other thing, i forgot that we also have the case that web is pulled and build in place (`make node-generate-dev` -> `make -C tmp/assets/core/origin release`) in the web service, and as always, it's complicated, and the pnpm versions differ (web, idp) :(.

the clean solution would be to use corepack... its not a show stopper in my opinion but something we should keep an eye on

pbleser-oc: Thanks for your feedback, I updated the PR to only use a script to determine the Go version, and dropped the part for pnpm since it turns out that it's currently maintained manually, and not updated as often.

In that case, no need to have that second script run every time as well.

fschade (approved):

PR #3437, opened 2026-09-01
graph: expose pendingOperations on driveItems

dschmidt:
Follow-up to #3113, implements opencloud-eu/libre-graph-api#60.

driveItems now carry `pendingOperations.pendingContentUpdate` while post-processing (virus scanning, ...) has not finished, read from the `status` opaque the storage provider already sets. WebDAV PROPFIND reports the same state as a `425 TOO EARLY` propstat; Graph listings use the facet instead.

`queuedDateTime` stays absent, we do not track when the operation was queued, and the spec allows to leave it out.


micbar (approved):
LGTM

PR #3438, opened 2026-09-01
graph: expose @libre.graph.shareTypes on driveItems

dschmidt:
Stacked on #3437 (pendingOperations), implements opencloud-eu/libre-graph-api#58.

driveItems can carry `@libre.graph.shareTypes`, listing the kinds of shares that exist on an item so a listing can render share indicators without expanding permissions. Replaces `oc:share-types` from WebDAV PROPFIND for Graph listings.

`user` and `group` come from the storage grants, which the provider only reports when `share-types` is in the field mask; `link` has to be reassembled from the public share manager, one filter per item. Both are only done when the annotation is requested via `$select`, mirroring WebDAV, where `oc:share-types` is in `requiresExplicitFetching` and never part of an `allprop` response. A failing public share lookup costs the link type, not the response.

`remote` is in the spec enum but never emitted, ocdav does not report federated shares as a share type either.

### DIFF
--- a/services/graph/pkg/service/v0/driveitems.go
+++ b/services/graph/pkg/service/v0/driveitems.go
@@ -15,13 +15,16 @@ import (
 
 	gateway "github.com/cs3org/go-cs3apis/cs3/gateway/v1beta1"
 	cs3rpc "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
+	link "github.com/cs3org/go-cs3apis/cs3/sharing/link/v1beta1"
 	storageprovider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
 	types "github.com/cs3org/go-cs3apis/cs3/types/v1beta1"
 	"github.com/go-chi/render"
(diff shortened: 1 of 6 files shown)


services/graph/pkg/service/v0/driveitems.go
```diff
@@ -500,6 +531,63 @@ func cs3ResourceToDriveItem(logger *log.Logger, publicBaseURL *url.URL, res *sto
...
+func (g Graph) listLinkShares(ctx context.Context, infos []*storageprovider.ResourceInfo) map[string]struct{} {
+	gatewayClient, err := g.gatewaySelector.Next()
+	if err != nil {
+		g.logger.Error().Err(err).Msg("could not select gateway client for public shares")
+		return nil
+	}
+
+	filters := make([]*link.ListPublicSharesRequest_Filter, 0, len(infos))
+	for _, info := range infos {
+		filters = append(filters, publicshare.ResourceIDFilter(info.GetId()))
```
butonic, line 613: yeah same problem we have in webdav ...

butonic (approved):

PR #3457, opened 2026-09-03
fix(thumbnails): bound declared image dimensions before decoding

dschmidt:
## Description

The non-vips (imaging) build decodes the full pixel buffer from the header declared dimensions before the existing `MaxInputWidth`/`MaxInputHeight` guard runs. That guard sits in `SimpleManager.Generate`, at which point `preprocessor.Convert` has already decoded the image. The Go stdlib backend allocates `image.NewGray`/`image.NewYCbCr` for the declared width*height at the SOS marker, before it fails on the mismatched scan data, so a tiny crafted file forces a large allocation and can exhaust the memory of the thumbnails worker.

This reads only the header with `DecodeConfig`, which allocates no pixels, and rejects oversized sources before the decode happens. The limit is threaded through the decoders that recurse into a second attacker controlled image: audio cover art (mp3/flac/ogg) and the geogebra decoders. The vips build gets the same check on the header lazy `vips.NewImageFromReader`, before `ThumbnailWithSize` materializes anything.

The service now passes the configured `MaxInputWidth`/`MaxInputHeight` into the preprocessor options, so a zero limit keeps the previous behaviour and an unconfigured decoder stays unrestricted.

Impact is availability only, and the official container images are not affected because they build with `ENABLE_VIPS=true`. Affected are the non-vips builds produced by `make release-linux` / `release-darwin`, so development builds and the published standalone binaries. Reachable by anyone who can have a preview generated, which means authenticated upload or a public share.

## Related Issue

Reported privately to security@opencloud.eu on 2026-08-18 and cleared for public handling, no public issue exists.

## Motivation and Context

A 375 byte grayscale JPEG whose SOF0 declares large dimensions while the pixel data stays 8x8 drives the allocation in `ImageDecoder.Convert`. Peak live `HeapAlloc`: 20000x20000 gives roughly 381 MB, 65535x65535 gives roughly 4 GB. With the guard in place the same files stay at roughly 17 KB.

## How Has This Been Tested?

- test environment: local build, `go test ./services/thumbnails/...`, plus a running container to check the vips path
- test case 1: crafted JPEG with a 20000x20000 SOF0 header is rejected with `ErrImageTooLarge` before the decode, verified by the new specs and by peak heap measurement
- test case 2: legitimate images (800x600 JPEG, gif, ggs, ggp, audio cover art) decode unchanged, existing preprocessor specs stay green
- test case 3: an oversized cover image embedded in a ggp file is rejected, which covers the recursive decoders
- test case 4: on a vips container a crafted preview was already rejected (HTTP 403/404) with flat memory, so that build keeps its behaviour

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

### DIFF
--- a/services/thumbnails/pkg/preprocessor/dimensionguard_test.go
+++ b/services/thumbnails/pkg/preprocessor/dimensionguard_test.go
@@ -1,0 +1,105 @@
+//go:build !enable_vips
+
+package preprocessor
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"image"
+	"image/jpeg"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	thumbnailerErrors "github.com/opencloud-eu/opencloud/services/thumbnails/pkg/errors"
+)
+
+// craftDimensionBomb encodes a tiny valid grayscale JPEG, then overwrites the
+// SOF0 width/height so the header declares huge dimensions while the payload
+// stays tiny.
+func craftDimensionBomb(width, height uint16) []byte {
+	var buf bytes.Buffer
+	Expect(jpeg.Encode(&buf, image.NewGray(image.Rect(0, 0, 8, 8)), &jpeg.Options{Quality: 10})).To(Succeed())
+	b := buf.Bytes()
+	for i := 0; i+9 < len(b); i++ {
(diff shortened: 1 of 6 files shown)


rhafer (approved):
LGTM. Thanks!

Issue #3082, opened 2026-07-07
chi middleware RealIP is deprecated


butonic:
We use it to determine the client ip for logs and rego checks:

> RealIP is a middleware that sets a http.Request's RemoteAddr to the results of parsing either the True-Client-IP, X-Real-IP or the X-Forwarded-For headers (in that order).
> 
> Deprecated: RealIP is vulnerable to IP spoofing — it mutates r.RemoteAddr to the leftmost X-Forwarded-For value, or to True-Client-IP / X-Real-IP whether or not your infrastructure actually sets them. See GHSA-3fxj-6jh8-hvhx, GHSA-rjr7-jggh-pgcp, GHSA-9g5q-2w5x-hmxf.
> 
> Use [ClientIPFromHeader](vscode-file://vscode-app/c:/Users/jfd/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [ClientIPFromXFF](vscode-file://vscode-app/c:/Users/jfd/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [ClientIPFromXFFTrustedProxies](vscode-file://vscode-app/c:/Users/jfd/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) or [ClientIPFromRemoteAddr](vscode-file://vscode-app/c:/Users/jfd/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and read the IP with [GetClientIP](vscode-file://vscode-app/c:/Users/jfd/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instead. These never mutate r.RemoteAddr.

PR #3464, opened 2026-09-04
fix(chore): upd deprecated chi RealIP middelware

maki5:
replaced deprecated chi RealIP middleware with ClientIPFromRemoteAddr and ClientIPFromXFF as a fallback option for proxy

closes #https://github.com/opencloud-eu/opencloud/issues/3082


services/proxy/pkg/command/server.go
```diff
@@ -341,7 +341,8 @@ func loadMiddlewares(logger log.Logger, cfg *config.Config,
 	}
 
 	return alice.New(
-		chimiddleware.RealIP,
+		chimiddleware.ClientIPFromRemoteAddr,
+		chimiddleware.ClientIPFromXFF(),
```
rhafer: According to the docs (https://github.com/go-chi/chi/blob/master/README.md#choosing-a-clientip-middleware) only one middleware from the `ClientIPFrom....`  should be selected. Which one depends on infrastructure.

As we allow to deploy in many different ways I don't think there is a single choice that works for us in allow possible configurations. 😢

rhafer (changes requested):
Hm, reading through https://github.com/go-chi/chi/blob/master/README.md#choosing-a-clientip-middleware, https://github.com/go-chi/chi/issues/708 and other issues linked from there, this one is a becoming a bit of a can of worms.

Apparently there is no real one-size-fits-all replacement for the `RealIP` middleware, depending on the concrete setup different middlewares from the `ClientIPFrom...` family have to be used and additional configuration would be needed.

For our proxy services:
- When the proxy is directly exposed the `ClientIPFromRemoteAddr` could be used
- It the used reverse proxy unconditionally set and overwrites a specific header with the client IP it would be `ClientIPFromHeader`
- If the IPs of all involved reverseproxys  are know `ClientIPFromXFF`
- If the number of trusted reverseproxy hops between the client and our proxy service is know `ClientIPFromXFFTrustedProxies`

For all other services that used the `RealIP` middleware it would be good if our `proxy` service unconditionally adds and overwrites a header with the value resulting from the above `ClientIPFrom...` middlewares. The "underlying" services could then just use `ClientIPFromHeader`

And even with the above there will be case where we would not be able to determine the correct remote IP address.

maki5:
@rhafer if I got it correctly we need an additional config where the ip "extraction" strategy will be specified, based on the value from config we apply the correct chi middleware and through it extract the real ip address. Did I got it right?

rhafer:
> @rhafer if I got it correctly we need an additional config where the ip "extraction" strategy will be specified, based on the value from config we apply the correct chi middleware and through it extract the real ip address. Did I got it right?

Yes, exactly. I wonder however what a good config would look like that fits the above suggesting without getting to complicated.

Any suggestions?

@micbar @butonic maybe you want to chime in here as well as you have more insights into possible setups, especially when it comes to k8s setups (where multiple reverse proxys a probably not an exception)

PR #3466, opened 2026-09-05
feat(proxy): add optional OIDC access token audience validation

### LINKED ISSUE #3456: Allow configuring valid audiences
When an OpenID Connect identity provider issues access tokens for multiple applications, an access token for e.g. Immich could be used for OpenCloud. The 

From https://www.rfc-editor.org/info/rfc7519/#section-4.1.3

```
4.1.3.  "aud" (Audience) Claim

   The "aud" (audience) claim identifies the recipients that the JWT is
   intended for.  Each principal intended to process the JWT MUST
   identify itself with a value in the audience claim.  If the principal
   processing the claim does not identify itself with a value in the
   "aud" claim when this claim is present, then the JWT MUST be
   rejected.  In the general case, the "aud" value is an array of case-
   sensitive strings, each containing a StringOrURI value.  In the
   special case when the JWT has one audience, the "aud" value MAY be a
   single case-sensitive string containing a StringOrURI value.  The
   interpretation of audience values is generally application specific.
   Use of this claim is OPTIONAL.
```

Currently, OpenCloud does not verify this "aud" claim. We should create an env var `PROXY_OIDC_AUDIENCES` or `OIDC_AUDIENCES` to configure an optional list of valid audiences. Default is empty, since `Use of this claim is OPTIONAL.`

cc @dragotin @rhafer @pbleser-oc

zerox80:
## Description

Add optional OIDC access token audience validation through `PROXY_OIDC_AUDIENCES` or `oidc.audiences` in `proxy.yaml`. The list defaults to empty for compatibility. When configured, a token must contain at least one exact, case-sensitive match in its `aud` claim. Missing, empty, malformed and nonmatching audiences are rejected during token verification.

Token verification, including the audience check, runs on Userinfo cache misses. Cache hits reuse unexpired cached claims without verifying the token again or requesting Userinfo. Existing entries in a shared or persistent cache can therefore remain accepted under the previous audience configuration until they expire. The documentation explains clearing the Userinfo cache after updating all proxy instances when a policy change must take effect immediately.

Audience configuration is validated when creating the OIDC client, so invalid settings fail during startup. The proxy logs an audience warning only when OIDC and JWT verification are enabled and the audience list is empty. The README includes ENV and YAML examples, the built-in IDP's client-ID audience behavior, Keycloak setup guidance and a recommendation to enable the check in production.

The token parsing optimization is in #3530 and can be reviewed independently.

## Related Issue

Fixes #3456

## Motivation and Context

When an IdP serves several applications, OpenCloud can currently accept an otherwise valid access token issued for another application. This gives administrators an explicit way to restrict tokens to OpenCloud without breaking existing IdP configurations by default.

## How Has This Been Tested?

Tested on Ubuntu WSL with Go 1.25.9 and GCC, using the vendored dependencies.

- `go test -mod=vendor -race -count=1 ./pkg/oidc/... ./services/proxy/...` passed after the constructor validation changes and rebase onto main.
- The local changed-lines golangci-lint fast check reached its 60-second timeout while loading packages; no lint result was produced. Full lint validation is left to CI.
- `git diff --check` passed.

The tests use locally generated keys and an HTTP test IdP with discovery, JWKS and Userinfo endpoints. They cover both `skip_user_info` settings, audience matching and rejection, cached Userinfo reuse, expired or corrupt cache entries, logout invalidation, ENV/YAML precedence, rejection of invalid configuration during client setup and the startup warning, including its suppression with verification set to none. A middleware test verifies that cache misses call token verification once and cache hits skip it. Signature, issuer, `nbf`, token expiry and AD FS issuer checks are also covered. No external IdP is needed.

## Types of changes

- [ ] Technical debt

## Checklist

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation added

HTTP integration tests are included in the Go suite. The Behat acceptance suite was not extended.

GPT 6 Astra was used as a supporting tool for the implementation and tests.


zerox80:
Do you think a small performance benchmark for the cache-hit path would be useful here, given that the access token now needs to be verified even on cache hits?

zerox80:
I did a small performance investigation of the audience-enabled cache-hit path.

The additional cost is dominated by the required JWT signature verification. Audience-enabled cache hits were around 99 µs/op compared to around 8 µs/op without per-hit verification.

I looked into avoiding repeated verification through policy-scoped caching, but that would weaken the current behavior around JWKS key removal/replacement and would also require broader cache/logout invalidation changes, so I kept that out of this PR.

I did find one small safe optimization in the existing verification path: avoiding the second full JWT parse after successful verification. In controlled serial benchmarks this reduced audience-enabled cache-hit time by about 5–7% and saves 1,088 B / 22 allocations per hit, while leaving the audience-disabled path unchanged.

Full affected package tests and race tests pass.

rhafer (changes requested):
@zerox80 Thanks a lot for you efforts! 

Would you mind submitting the `fix(config): correct pending version annotations` commit as a separate PR and remove it from this one?

We could fast-track that one, as it is straight-forward. The rest needs a more thorough review.

zerox80:
> @zerox80 Thanks a lot for you efforts!
> 
> Would you mind submitting the `fix(config): correct pending version annotations` commit as a separate PR and remove it from this one?
> 
> We could fast-track that one, as it is straight-forward. The rest needs a more thorough review.

done

changelog/unreleased/feature-oidc-audiences.md
rhafer, line 1: Please remove this. We've switched away from tracking the changelog in `changelog/unrelease` quite a while about. I hope coding agents will understand that as well, after we merged: https://github.com/opencloud-eu/opencloud/pull/3529

zerox80: sure, will do

pkg/oidc/client.go
```diff
@@ -301,14 +311,26 @@ func (c *oidcClient) verifyAccessTokenJWT(token string) (RegClaimsWithSID, jwt.M
...
+	// header and signature again. Keep typed claims above for validation.
+	_, payloadAndSignature, _ := strings.Cut(token, ".")
+	payload, _, _ := strings.Cut(payloadAndSignature, ".")
+	claimBytes, err := new(jwt.Parser).DecodeSegment(payload)
+	if err != nil {
+		return claims, mapClaims, fmt.Errorf("%w: could not base64 decode claim: %w", jwt.ErrTokenMalformed, err)
+	}
+	// Match ParseUnverified's map value semantics, including a null payload.
+	decodedMapClaims := mapClaims
+	err = json.Unmarshal(claimBytes, &decodedMapClaims)
```
rhafer: What is this hunk good for? This is just a performance optimization, right?

If that is true, I'd rather see that submitted in a separate PR and keep this PR limited to the audience validation.

services/proxy/pkg/middleware/oidc_auth.go
```diff
@@ -31,32 +30,45 @@ func NewOIDCAuthenticator(opts ...Option) *OIDCAuthenticator {
...
+	validateAccessTokenOnCacheHit bool
+	skipUserInfo                  bool
+	TimeFunc                      func() time.Time
 }
 
 func (m *OIDCAuthenticator) getClaims(token string, req *http.Request) (map[string]any, bool, error) {
+	var aClaims oidc.RegClaimsWithSID
+	var tokenClaims map[string]any
+	var err error
+	if m.validateAccessTokenOnCacheHit {
```
rhafer: What is this flag actually good for? Why do we need it?

rhafer (changes requested):
Thanks again. Here's round of feedback after a quick look through the code.

pkg/oidc/client.go
zerox80: yes, this is a performance optimization and is independent of audience validation. I will move it, together with its claim-extraction tests, into a separate PR and keep this PR focused on audience validation

services/proxy/pkg/middleware/oidc_auth.go
zerox80: I added this because on a cache hit, `getClaims` returns before `VerifyAccessToken` gets called. The cache key is just the token hash, so we don't know which audience policy was used when the token was first accepted.

With a shared or persistent cache, a token accepted before enabling audience validation could still get through after restarting the proxy with a stricter config. Same thing if the allowed audiences change. `TestOIDCAudiencePolicyChangesWithCachedUserinfo` covers this for both `skip_user_info` settings.

We also cant just check `aud` from the cached Userinfo, since that can be different from the audience in the access token.

The flag is set internally when `cfg.OIDC.Audiences` is non-empty. So the extra JWT verification only happens when audience validation is enabled. Userinfo is still cached, and on a cache miss we reuse the verification result instead of verifying twice.

i can add a clearer comment explaining why it's there.

rhafer: > I added this because on a cache hit, `getClaims` returns before `VerifyAccessToken` gets called. The cache key is just the token hash, so we don't know which audience policy was used when the token was first accepted.
> 
> With a shared or persistent cache, a token accepted before enabling audience validation could still get through after restarting the proxy with a stricter config.

Can we please remove this? I think the benefit this adds is pretty limited, given the short live time of access tokens.

zerox80: yeah I get the concern about checking the signature on every cache hit. My worry is that we don't enforce a max token lifetime, so depending on the IdP config this could be longer than a few minutes.

Maybe we could store which audience config was used when caching the token? If it's missing or doesn't match anymore, we verify again. That should avoid the extra work on normal cache hits and still handle config changes.

Would that work for you?

rhafer: I'd really prefer to have this removed. In a single instance setup this is a non-issue as the cache is not persisted and is basically empty after a restart.

For multi instance setup, the solution would be to invalidate the cache manually after a changed config as been rolled out to all instances. 

Oh, I just saw you already removed it, thanks.

services/proxy/pkg/command/oidc.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/opencloud-eu/opencloud/pkg/log"
+	"github.com/opencloud-eu/opencloud/pkg/oidc"
+	"github.com/opencloud-eu/opencloud/services/proxy/pkg/config"
+	"github.com/opencloud-eu/opencloud/services/proxy/pkg/middleware"
+	"go-micro.dev/v4/store"
+)
+
+func newOIDCAuthenticator(logger log.Logger, cfg *config.Config, userInfoCache store.Store, httpClient *http.Client) *middleware.OIDCAuthenticator {
+	if cfg.OIDC.Issuer != "" && len(cfg.OIDC.Audiences) == 0 {
+		logger.Warn().Msg("OIDC access token audience validation is disabled. Configure PROXY_OIDC_AUDIENCES to enable it; this is recommended for production.")
```
rhafer, line 15: We should only log this if  `AccessTokenVerifyMethod != "none'`. Otherwise this is misleading.

pkg/oidc/client.go
```diff
@@ -270,6 +272,14 @@ func (c *oidcClient) UserInfo(ctx context.Context, tokenSource oauth2.TokenSourc
...
 
 func (c *oidcClient) VerifyAccessToken(ctx context.Context, token string) (RegClaimsWithSID, jwt.MapClaims, error) {
+	if len(c.accessTokenAudiences) > 0 && c.accessTokenVerifyMethod != config.AccessTokenVerificationJWT {
+		return RegClaimsWithSID{}, jwt.MapClaims{}, errors.New("access token audience validation requires the jwt verification method")
+	}
+	for _, audience := range c.accessTokenAudiences {
+		if strings.TrimSpace(audience) == "" {
+			return RegClaimsWithSID{}, jwt.MapClaims{}, errors.New("access token audiences must not contain empty or whitespace-only entries")
+		}
+	}
```
rhafer, line 282: This config validation should IMO happen in `NewOIDCClient`, so that we have a chance to fail early (`NewOIDCClient` would obviously need a separate error return value for that), instead of failing every single request.

services/proxy/README.md
```diff
@@ -13,6 +13,62 @@ The following request authentication schemes are implemented:
...
+claims without verifying the token again or requesting Userinfo. Existing entries
+in a shared or persistent cache can remain valid under the previous audience
+configuration until they expire. Clear the Userinfo cache after updating all
+proxy instances if the new policy must take effect immediately.
+
+The disabled default is a compatibility decision. It does not relax the
+[audience validation requirement in RFC 9068, Section 4](https://www.rfc-editor.org/rfc/rfc9068.html#name-validating-jwt-access-token):
+a resource server following that JWT access token profile must reject tokens
+whose audience does not identify the resource server.
+
```
rhafer, line 71: I think we should add a few words/links about:

- how the builtin IDP does (not) handle the audience claim. (Basically it just hardcodes the `aud` claim to the client id of the authenticated client)
- how properly setup audiences in keycloak (a pointer to: https://www.keycloak.org/docs/latest/server_admin/#audience-support)

services/proxy/pkg/command/oidc.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/http"
+
+	"github.com/opencloud-eu/opencloud/pkg/log"
+	"github.com/opencloud-eu/opencloud/pkg/oidc"
+	"github.com/opencloud-eu/opencloud/services/proxy/pkg/config"
+	"github.com/opencloud-eu/opencloud/services/proxy/pkg/middleware"
+	"go-micro.dev/v4/store"
+)
+
+func newOIDCAuthenticator(logger log.Logger, cfg *config.Config, userInfoCache store.Store, httpClient *http.Client) *middleware.OIDCAuthenticator {
```
rhafer, line 13: Nitpick: please move this function to server.go. I think it not needed to have it in a separate file.

rhafer (changes requested):
Just a few smaller comments after some testing.

Also: Please rebase this once more on the latest main branch to get the latest CI fixes.

services/proxy/README.md
```diff
@@ -13,6 +13,79 @@ The following request authentication schemes are implemented:
...
+this proxy option does not change the tokens issued by the IDP.
+
+For Keycloak, add an **Audience** protocol mapper to a client scope. Set
+**Included Client Audience** to the OpenCloud resource client, or use
+**Included Custom Audience** for a value such as `opencloud-api`, and enable
+**Add to access token**. Assign the scope as a default scope to each client
+accessing OpenCloud so the audience is included without an extra `scope`
+parameter. Use that same audience in `PROXY_OIDC_AUDIENCES`. See
+[Keycloak's audience support documentation](https://www.keycloak.org/docs/latest/server_admin/#audience-support)
+for details and the alternative based on client roles.
```
rhafer:
```suggestion
For other IDPs, please refer to their documentation for proper support for the `aud` claim. For Keycloak see e.g.: [Keycloak's audience support documentation](https://www.keycloak.org/docs/latest/server_admin/#audience-support)
```

rhafer (changes requested):

rhafer (approved):

Issue #3469, opened 2026-09-06
OpenCloud process grows to ~6 GB anonymous RSS, hits container memory limit and becomes unresponsive


Surfer2010:
## Description

I am seeing recurring severe memory growth of the OpenCloud server process.

Over time, `/usr/bin/opencloud server` grows from a normal memory footprint of roughly 100–300 MB to approximately 6 GB RSS.

Almost the entire RSS consists of anonymous memory (`RssAnon`).

Once this happens, the container reaches its configured 6 GiB memory limit, memory and CPU pressure become extremely high, OpenCloud stops responding properly on port 9200, and eventually the instance has to be restarted.

This has happened repeatedly and is reproducible across multiple OpenCloud versions.

I have collected detailed 5-second monitoring data from outside the affected container.

---

## Environment

OpenCloud is running in Docker inside a Proxmox LXC container.

Host:

- Proxmox
- Kernel: `6.8.12-2-pve`
- LXC container memory limit: 6 GiB
- OpenCloud exposed internally on TCP port `9200`

OpenCloud versions where the problem has been observed:

- 7.2.4
- newer versions were tested as well
- the installation has since been upgraded further while investigating the issue

The issue is not limited to the public reverse proxy path. We have also observed direct internal failures when connecting to the OpenCloud backend at `192.168.16.13:9200`.

---

## Normal state

After restarting OpenCloud, typical memory usage is approximately:

```text
VmRSS:    ~100–250 MB
RssAnon:  ~80–120 MB
```

There are normally no CLOSE-WAIT connections on port 9200.

---

## Failure state

During one captured failure:

```text
memory.current = 6395109376
memory.max     = 6442450944
```

OpenCloud process:

```text
RSS: ~6004348 kB
```

The OpenCloud process alone therefore consumes approximately 5.7 GiB RSS.

Almost all of this memory is anonymous memory.

Previous captures showed values such as:

```text
VmRSS:   ~6000 MB
RssAnon: ~6000 MB
```

This is not page cache growth.

It appears to be anonymous userspace memory retained by the OpenCloud Go process.

---

## cgroup pressure during the failure

Example:

```text
memory.pressure:

some avg10=99.13 avg60=99.11 avg300=99.09
full avg10=80.84 avg60=80.87 avg300=80.66
```

CPU pressure:

```text
some avg10=99.05 avg60=99.08 avg300=99.05
full avg10=5.80 avg60=5.60 avg300=5.70
```

I/O pressure:

```text
[...]

Issue #1269, opened 2025-07-24
Out of memory: Killed process 1680404 (opencloud) total-vm:17239740kB...


cscholz:
Out of memory: Killed process 1680404 (opencloud) total-vm:17239740kB, anon-rss:11584020kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:24268kB oom_score_adj:0 2025-07-24T02:01:16.665086+02:00 pi kernel: [65667.196863] Out of memory: Killed process 1682955 (opencloud) total-vm:17025884kB, anon-rss:12785864kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:26440kB oom_score_adj:0 2025-07-24T02:06:49.696520+02:00 pi kernel: [66000.187638] Out of memory: Killed process 1686575 (opencloud) total-vm:17176784kB, anon-rss:12798440kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:26580kB oom_score_adj:0 2025-07-24T02:12:10.145965+02:00 pi kernel: [66320.669511] Out of memory: Killed process 1690485 (opencloud) total-vm:17882900kB, anon-rss:12787128kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:27624kB oom_score_adj:0

## Describe the bug
I removed about 100 GB of data and emptied the trash bin using the web interface.
After restarting the opencloud container, it started consuming all available RAM until it crashed and starts doing the same. Now since 2 days

## Steps to reproduce
1. delete 100 GB of files
2. restart container
3. check ram usage, syslog

## Expected behavior
no crash :-)

## Actual behavior
crash

## Setup
OpenCloud 3.2.0
OpenCloud Web UI 3.2.0

docker-compose setup running in podman

<details>
<p>

```console output
**console output**
2025-07-24T01:56:42.596890+02:00 pi kernel: [65393.075584] Out of memory: Killed process 1680404 (opencloud) total-vm:17239740kB, anon-rss:11584020kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:24268kB oom_score_adj:0
2025-07-24T02:01:16.665086+02:00 pi kernel: [65667.196863] Out of memory: Killed process 1682955 (opencloud) total-vm:17025884kB, anon-rss:12785864kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:26440kB oom_score_adj:0
[...]

PR #3485, opened 2026-09-08
fix(search): bound memory of the bleve descendant lookup

n-at-han-k:
# fix(search): bound memory of the bleve descendant lookup

## Description

Deleting, moving, restoring or purging a **folder** makes the search service
look up every descendant of that folder in the bleve index.
`searchResourcesByPath` does this with a wildcard query:

```go
bleve.NewQueryStringQuery("Path:"+escapeQuery(lookupPath+"/*")),
```

`Path` is a keyword field — one term per document, the full path — so bleve
expands this wildcard into **one term searcher per descendant, all alive at
once**, each holding zapx segment-dictionary structures and vellum FST readers.
Peak live memory scales with `descendants × segments` and is superlinear in
index size. On a production instance a routine folder delete held **2.24 GB of
a 2.29 GB live heap** in this one stack (`go tool pprof`, opencloud 7.5.0):

```
1287MB  github.com/blevesearch/vellum.(*FST).Reader
 502MB  github.com/blevesearch/zapx/v17.(*SegmentBase).dictionary
 416MB  github.com/blevesearch/bleve/v2/index/scorch.(*IndexSnapshot).TermFieldReader
        ... under WildcardQuery.Searcher ← Batch.Delete
```

The process was OOM-killed by the kernel 112 times in one day on that host.
This is very likely the mechanism behind #1269 (comments describe OOM on
folder delete / trash emptying) and plausibly contributes to #3469.

## The fix

Keep the same function, signature and result set, but enumerate the matching
path terms from the field dictionary (`FieldDictPrefix` — an iterator, no
searchers) and fetch the documents in bounded batches of 500 exact
`TermQuery`s. Live searcher memory becomes O(500) instead of O(descendants).
The `RootID` filter moves from `QueryStringQuery` to an exact `TermQuery`, so
IDs containing `$`/`!` no longer pass through the query parser. `escapeQuery`
loses its last caller and is removed.

The opensearch backend already avoids this class of problem via its
`path_hierarchy` term query; this brings the bleve backend in line.

## Benchmark

`BenchmarkSearchResourcesByPath` (added in this PR) builds an index of N files
in one folder and runs the descendant lookup. Cumulative allocations are
similar for any implementation that visits every descendant — what OOMs
servers is the **peak live heap** while the lookup runs, reported as `peak-MB`
via a background `HeapInuse` sampler:

```
go test -run XXX -bench BenchmarkSearchResourcesByPath -benchtime 3x ./services/search/pkg/bleve/
```

| docs in folder | main | this PR | |
|---|---|---|---|
| 20,000 | 196 peak-MB · 814 ms/op | **18 peak-MB** · 815 ms/op | 11× lower peak |
| 100,000 | 1194 peak-MB · 4.64 s/op | **102 peak-MB** · 4.44 s/op | 11.7× lower peak, 4% faster |

Note the superlinear growth on main (5× docs → 6× peak) versus linear with the
fix — the gap keeps widening with index size.

Both variants return the identical result set (asserted inside the benchmark:
`len(res) == N`), and the existing package tests pass unchanged.

## Regression tests

`descendants_test.go` (added in this PR):

- `TestSearchResourcesByPath` — correctness edges the lookup must keep:
  1001 descendants (crosses the 500-term batch boundary twice), the folder
  itself excluded, a sibling whose name shares the prefix (`./big2` vs
  `./big`) excluded, the same path under another space's RootID excluded,
  and paths containing `* [ ] ? :` and spaces.
- `TestSearchResourcesByPathMemoryBounded` — fails any implementation whose
  live memory scales with folder size: 20k-doc folder, 64 MB peak-heap limit.
  The former wildcard implementation holds ~168 MB here and fails; this
  implementation peaks under ~20 MB. Skipped with `-short`.

## End-to-end reproduction

Branch [`memory-bug-test`](https://github.com/n-at-han-k/opencloud/tree/memory-bug-test/memory-bug-test)
contains a self-contained nix flake with **two microVMs** that run the full
scenario against real servers built from source — one unpatched, one carrying
only this patch, that being the only difference. Each VM uploads N files over
WebDAV, waits for search to index them, deletes the folder, and records the
opencloud cgroup at 1–10 Hz plus search-service pprof heap profiles to
`repro-out/`:

```
cd memory-bug-test
nix run .#vm-unpatched   # then: nix run .#vm-patched
```

## Related Issue

- Fixes the mechanism reported in #1269 (OOM on folder delete / empty trash) —
  closed as unreproducible; the benchmark above is the missing reproducer
- Related to #3469 (multi-GB anonymous RSS growth until unresponsive)

## How Has This Been Tested?

- `go test ./services/search/pkg/bleve/ ./services/search/pkg/search/` — pass
- New regression tests verified in both directions: memory-bound test fails on
  the wildcard implementation (168 MB > 64 MB limit) and passes on this one
- New benchmark above, before/after on this branch
- Two 3 GB NixOS microVMs built from source (one with this change, one
  without): 12k-file upload → index → folder delete over WebDAV; identical
  end-to-end behaviour and search results, no regression
- Production heap profile identifying the stack (7.5.0); the same code is on
  `main`

## Types of changes

## Checklist:

- [x] Code changes
- [x] Unit tests added (correctness + memory-bound regression tests, benchmark)
- [ ] Acceptance tests added
- [ ] Documentation ticket raised

https://claude.ai/code/session_01UozPhD3wkKcnfisgoENx6G

### DIFF
--- a/changelog/unreleased/fix-search-descendant-lookup-memory.md
+++ b/changelog/unreleased/fix-search-descendant-lookup-memory.md
@@ -1,0 +1,17 @@
+Bugfix: Bound memory of the search descendant lookup
+
+Deleting, moving, restoring or purging a folder made the search service look
+up every descendant of that folder with a Path wildcard query. Path is a
+keyword field, so bleve expanded the wildcard into one term searcher per
+descendant, all alive at once. Peak live memory scaled with the number of
+descendants and the kernel OOM-killed the whole server on folder deletes; on
+one production instance a routine delete held 2.24 GB of a 2.29 GB live heap
+in this single query.
+
+The lookup now enumerates the matching path terms from the field dictionary
+and fetches the documents in bounded batches of exact term queries, returning
+the same result set with O(1) live searcher memory: a 100k-file folder went
+from 1194 MB peak to 102 MB, slightly faster than before.
+
+https://github.com/opencloud-eu/opencloud/issues/1269
+https://github.com/opencloud-eu/opencloud/issues/3469
--- a/services/search/pkg/bleve/bleve.go
+++ b/services/search/pkg/bleve/bleve.go
@@ -1,7 +1,6 @@
 package bleve
 
 import (
-	"regexp"
 
 	bleveSearch "github.com/blevesearch/bleve/v2/search"
 	storageProvider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
@@ -11,7 +10,6 @@ import (
 	"github.com/opencloud-eu/opencloud/services/search/pkg/search"
 )
 
-var queryEscape = regexp.MustCompile(`([` + regexp.QuoteMeta(`+=&|><!(){}[]^\"~*?:\/`) + `\-\s])`)
 
 func getFieldValue[T any](m map[string]any, key string) (out T) {
 	val, ok := m[key]
@@ -85,6 +83,3 @@ func matchToResource(match *bleveSearch.DocumentMatch) *search.Resource {
 	return mapping.Deserialize[search.Resource](match.Fields)
 }
 
-func escapeQuery(s string) string {
-	return queryEscape.ReplaceAllString(s, "\\$1")
-}
--- a/services/search/pkg/bleve/descendants_bench_test.go
+++ b/services/search/pkg/bleve/descendants_bench_test.go
@@ -1,0 +1,89 @@
+package bleve
+
+import (
+	"fmt"
+	"runtime"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/opencloud-eu/opencloud/pkg/log"
+	"github.com/opencloud-eu/opencloud/services/search/pkg/search"
+)
+
+// BenchmarkSearchResourcesByPath measures the descendant lookup that backs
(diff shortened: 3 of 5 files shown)


dschmidt:
This is a really good catch, I'm just not sure the approach is the best we can do.
We just broke the search index schema, which requires a rescan on OpenCloud upgrade anyhow ... we might as well just fix this at the root instead of finding a slightly better cure for the symptom.

Let me cook something up - would be much appreciated, if you could give it a spin?

dschmidt:
Could you check https://github.com/opencloud-eu/opencloud/pull/3495?
What do you think?

n-at-han-k:
I honestly couldn't comment. I'm not much of GO developer.
I'm running 7.5.0 at the moment, pinned from the tag, which works with my patch.
I ran your tests on `main` and it runs much quicker and level than mine.

dschmidt:
@n-at-han-k that's honestly very valuable feedback. Thanks a lot. 

@butonic @fschade 👆

butonic:
The fix was implemented with https://github.com/opencloud-eu/opencloud/pull/3510

@n-at-han-k Thx for testing!

PR #3486, opened 2026-09-08
test(search): wait for expected properties and documents

dschmidt:
`apiSearchContent/extractedProps.feature` got the driveItem via the Graph API and waited only until *any* extraction facet key (`image`/`photo`/`location`/`audio`/`video`) appeared, then a separate step validated the full JSON schema.

Property extraction reaches the driveItem incrementally (async metadata write on the storage node), so the wait could stop on a partial payload, e.g. an `audio` facet that still lacks `album`/`genre`/`year`, and the schema validation then failed intermittently.

This combines the get and the schema match into one step that retries the GET until the response satisfies the expected schema, using the same `WaitHelper` timeout. A partial payload never matches the schema, so it keeps polling until the complete data is present (or it times out with a clear schema-mismatch failure). No product change; acceptance-only, complements the storage-side write fix.

### DIFF
--- a/tests/acceptance/bootstrap/SpacesContext.php
+++ b/tests/acceptance/bootstrap/SpacesContext.php
@@ -55,6 +55,9 @@ class SpacesContext implements Context {
 	 * key is space name and value is the username that created the space
 	 */
 	private array $createdSpaces = [];
+	// the request of the last driveItem GET, so a later "should eventually
+	// match" step can re-run it (the counterpart of the stored response)
+	private array $lastDriveItemRequest = [];
 	private string $ocsApiUrl = '/ocs/v2.php/apps/files_sharing/api/v1/shares';
 
 	/**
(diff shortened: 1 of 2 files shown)


fschade:
@v-scharf FYI

tests/acceptance/bootstrap/SpacesContext.php
```diff
@@ -4557,25 +4563,36 @@ public function userGetsTheDriveItemInSpace(string $user, string $file, string $
...
 				}
-				$body = $this->featureContext->getJsonDecodedResponseBodyContent($response);
-				return \is_object($body)
-					&& !empty(\array_intersect($extractionFacets, \array_keys((array) $body)));
 			}
 		);
 
 		$this->featureContext->setResponse($response);
+		// Final assert so a timeout surfaces a clear schema-mismatch failure.
+		$this->featureContext->assertJsonDocumentMatchesSchema(
```
v-scharf, line 4627: Since `Then the HTTP status code should be "200"` was removed from the scenarios, a 4xx/5xx now fails as a confusing schema mismatch or as a bare JsonException: Syntax error when the body isn't JSON.

Could you assert the status before the final schema check?

```
$this->featureContext->theHTTPStatusCodeShouldBe(
    200,
    "Expected 200 while waiting for extracted properties of '$file' in space '$space', got "
        . $response->getStatusCode() . "\nBody: " . (string)$response->getBody(),
    $response
);
```

dschmidt: Thanks, good catch. I restructured this so the get and the assertion are separate When/Then steps again: the new `Then the JSON data of the response should eventually match` re-fetches the item until the response satisfies the schema, and it asserts the 200 status first (with a descriptive message including the file/space and the body), as you suggested, so a genuine 4xx/5xx fails clearly instead of as a confusing schema mismatch.

v-scharf (approved):

dschmidt:
Got another approval for the latest state in chat :)

PR #3495, opened 2026-09-08
refactor(search): stream bleve descendants

dschmidt:
## Description

Move, delete, restore and purge collected every descendant before writing. They now walk the folder in pages of 20k, so live memory is bounded by the page, not the folder: 133 MB → 19 MB at 100k files. Follow-up to #3510.

## Related Issue

- #1269, #3469

## Types of changes


dschmidt:
Marking it as draft again because the important core change was extracted to #3510.

PR #3500, opened 2026-09-09
api-tests: add search retry to all search tests

v-scharf:
- reduse sleep(10) for search indexing
- retry search request using WaitHelper::waitUntil
- retry getting tags (graph api) using WaitHelper::waitUntil

now all search tests use same mechanism. I think that with this approach, we should not have failed tests in expected failures file - it would be a huge waste of time and resources. So we need fix and delete from expected failures file these tests: 

```
- [apiSearch1/search.feature:437](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L437)
- [apiSearch1/search.feature:438](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L438)
- [apiSearch1/search.feature:439](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L439)
- [apiSearch1/search.feature:465](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L465)
- [apiSearch1/search.feature:466](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L466)
- [apiSearch1/search.feature:467](https://github.com/opencloud-eu/opencloud/blob/main/tests/acceptance/features/apiSearch1/search.feature#L467)
```

Issue: no `<d:getetag></d:getetag>` in report response
CC @fschade

### DIFF
--- a/tests/acceptance/TestHelpers/WaitHelper.php
+++ b/tests/acceptance/TestHelpers/WaitHelper.php
@@ -39,16 +39,25 @@ class WaitHelper {
 	/**
 	 * Repeat $makeAttempt until $shouldStop returns true or the timeout elapses.
 	 *
-	 * @param callable $makeAttempt  makes one attempt (e.g. sends a request) and returns its result
-	 * @param callable $shouldStop   receives that result, returns true to stop polling
+	 * @param callable $makeAttempt      makes one attempt (e.g. sends a request) and returns its result
+	 * @param callable $shouldStop       receives that result, returns true to stop polling
+	 * @param int|null $intervalMs       pause between attempts in ms; defaults to self::INTERVAL_MS
+	 * @param int|null $timeoutSeconds   overall time to keep polling; defaults to self::TIMEOUT_SECONDS
 	 *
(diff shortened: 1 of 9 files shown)


v-scharf:
Fixing the failing search tests would take some time, so we decided to skip them for now.

here is issue https://github.com/opencloud-eu/opencloud/issues/3501

cc @rhafer

v-scharf:
image


looks good and faster: 

sleep(2) for index instead of sleep(10). 
how it works: 
Search request -> if empty -> wait 2 sec -> send search request again -> if empty -> wait 2 sec -> .... 

after 20 seconds -> stop polling -> test failed. If the search cannot re-index within 22 seconds, we consider this to be a bug (not flaky test)

micbar (approved):
LGTM!

We can use any improvement on that topic right away.

Let us merge, observe and iterate.

Thanks for the quick response.

PR #3505, opened 2026-09-10
add insecure to search reindex

v-scharf:
during the testing https://github.com/opencloud-eu/opencloud/pull/3197 I noticed that
```
$ opencloud search index --all-spaces --force-rescan
rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
```

after added `--insecure` works fine:

```
~ $ opencloud search index --all-spaces --force-rescan --insecure
[1/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$dc67b41d-fbb3-4c52-b4d7-59a0774d0ce5!dc67b41d-fbb3-4c52-b4d7-59a0774d0ce5 in 10.579733921s
[2/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$c3919acb-aa79-4d1e-aa56-2d48cb36c891!c3919acb-aa79-4d1e-aa56-2d48cb36c891 in 10.668783838s
[3/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$b5a2d63c-5223-4263-bb2f-b1efbe500f13!b5a2d63c-5223-4263-bb2f-b1efbe500f13 in 9.702030129s
[4/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$e8ca4dda-cf6c-437f-b6a8-1e31033b0a0a!e8ca4dda-cf6c-437f-b6a8-1e31033b0a0a in 9.751792088s
[5/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$9ef27369-04f4-4b34-97fc-ccff31a9985d!9ef27369-04f4-4b34-97fc-ccff31a9985d in 8.726422712s
[6/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$4702248d-7570-4b77-bedc-0364af8a45b4!4702248d-7570-4b77-bedc-0364af8a45b4 in 9.808745254s
[7/7] indexed space 7230b71a-5362-486c-8f35-414d443b638b$cae518d9-869b-4cd1-8847-2591ab7a4a70!cae518d9-869b-4cd1-8847-2591ab7a4a70 in 1m41.059535047s
```

also added to log: 
`{"level":"info","service":"search","index":"/var/lib/opencloud/search/bleve-v4","time":"2026-09-10T09:08:19Z","line":"github.com/opencloud-eu/opencloud/services/search/pkg/mapping/reconcile.go:43","message":"created a new empty search index; if this OpenCloud instance already held files, they are not in it yet, index them by running: opencloud search index --all-spaces --force-rescan --insecure"}`

### DIFF
--- a/services/search/MIGRATION.md
+++ b/services/search/MIGRATION.md
@@ -13,7 +13,7 @@ Fill the new index by indexing all spaces again:
 
 ```shell
 # the service keeps running while it happens
-opencloud search index --all-spaces
+opencloud search index --all-spaces --insecure
 ```
 
 Once the new index is filled, every index but the one with the highest
@@ -31,7 +31,7 @@ The new index is a directory next to the old `bleve` one, both in
 bleve index cannot be copied, index all spaces again:
 
 ```shell
-opencloud search index --all-spaces
+opencloud search index --all-spaces --insecure
 ```
 
 Once the new index is filled, every directory but the one with the highest
--- a/services/search/README.md
+++ b/services/search/README.md
@@ -124,14 +124,14 @@ opencloud search index --space $SPACE_ID
 It can also be used to re-index all spaces:
 
 ```shell
-opencloud search index --all-spaces
+opencloud search index --all-spaces --insecure
 ```
 
 Please note that a reindex only picks up new or changed files. Files that have already been indexed are not scanned again, even if the configuration or the whole extractor has been changed. To force a full rescan (re-running the extractor on every file) you need to use the `force-rescan` flag:
 
 
 ```shell
-opencloud search index --all-spaces --force-rescan
+opencloud search index --all-spaces --force-rescan --insecure
 ```
 
 ## Metrics
--- a/services/search/pkg/mapping/reconcile.go
+++ b/services/search/pkg/mapping/reconcile.go
@@ -27,7 +27,7 @@ func Reconcile(index string, r SchemaReconciler, logger log.Logger) (Classificat
 	case VerdictAdditive:
 		persisted, err := r.ApplyAdditive()
 		if persisted {
-			logger.Warn().Strs("fields", classification.NewFields).Str("index", index).Msg("extended the search index mapping with new fields; documents indexed before the upgrade do not contain them and queries on these fields will miss those documents until they are re-indexed; to re-index everything run: opencloud search index --all-spaces --force-rescan")
+			logger.Warn().Strs("fields", classification.NewFields).Str("index", index).Msg("extended the search index mapping with new fields; documents indexed before the upgrade do not contain them and queries on these fields will miss those documents until they are re-indexed; to re-index everything run: opencloud search index --all-spaces --force-rescan --insecure")
 		}
 		if err != nil {
 			return classification, err
@@ -40,5 +40,5 @@ func Reconcile(index string, r SchemaReconciler, logger log.Logger) (Classificat
 // LogNewIndexCreated logs that a fresh, empty index was created and how to
 // backfill it. The create path does not run through Reconcile.
 func LogNewIndexCreated(logger log.Logger, index string) {
-	logger.Info().Str("index", index).Msg("created a new empty search index; if this OpenCloud instance already held files, they are not in it yet, index them by running: opencloud search index --all-spaces --force-rescan")
+	logger.Info().Str("index", index).Msg("created a new empty search index; if this OpenCloud instance already held files, they are not in it yet, index them by running: opencloud search index --all-spaces --force-rescan --insecure")
 }


dschmidt:
I'm not sure about this. 

Actually I wouldn't want to encourage usage of the --insecure flag, especially not as a default. 

Could we instead improve the error message (or even deduct the --insecure flag automatically from the config file?)?

butonic:
I have a history with this. The search index command uses the configured search endpoint. which is an internal grps endpoint. which is not tls encrypted (unless you enable tls for grpc ... try finding the docs). we default to grpc over h2 ... not http2 ... which is why we need to use --insecure. Sucks ... I know ... and annoys me ... but also ... is not really a security issue because we just want to trigger a rescan. which happens in the opencloud process, not the `opencloud search index` command ... 🤪 

IMO the error message should just use the `--insecure` flag. or we hardcode insecure ... that actually seems less more wrong to me.

butonic:
documenting it like in https://github.com/opencloud-eu/docs/pull/1153#pullrequestreview-5166336402 is the least hassle.

dschmidt:
That's hella confusing 😂

I think requiring --insecure always and having it all over the docs is the worst of all options. So maybe just enable it always.. or..  why not the following? 

Here's what my claude suggests: 

The CLI already has the server's gRPC TLS config, it just evaluates it the wrong way round. `OC_GRPC_CLIENT_TLS_MODE` is never set automatically (`opencloud init` doesn't write it, nothing derives it), so the default is `""`. The go-micro client treats `""`/`off` as "no TLS" ([client.go#L75-L100](https://github.com/opencloud-eu/opencloud/blob/994086ceb2b06ee7ae2c0de518b384b7a05ac1db/pkg/service/grpc/client.go#L75-L100)), and the search server is plaintext unless `OC_GRPC_TLS_ENABLED=true` ([server.go#L16-L20](https://github.com/opencloud-eu/opencloud/blob/994086ceb2b06ee7ae2c0de518b384b7a05ac1db/services/search/pkg/server/grpc/server.go#L16-L20)). The CLI however only dials plaintext for `insecure` and uses TLS for everything else, including the default ([index.go#L49-L56](https://github.com/opencloud-eu/opencloud/blob/994086ceb2b06ee7ae2c0de518b384b7a05ac1db/services/search/pkg/command/index.go#L49-L56)). Plaintext server + TLS client is exactly the "first record does not look like a TLS handshake" error.

| Mode | go-micro client (all services) | `index.go` on main |
|---|---|---|
| `""` / `off` (default) | no TLS | TLS |
| `insecure` | TLS, no verification | no TLS |
| `on` | TLS, verified | TLS, verified |

The config is available in the CLI via `PersistentPreRunE` ([services.go#L264-L273](https://github.com/opencloud-eu/opencloud/blob/994086ceb2b06ee7ae2c0de518b384b7a05ac1db/opencloud/pkg/command/services.go#L264-L273)) and `EnsureDefaults` ([defaultconfig.go#L89-L94](https://github.com/opencloud-eu/opencloud/blob/994086ceb2b06ee7ae2c0de518b384b7a05ac1db/services/search/pkg/config/defaults/defaultconfig.go#L89-L94)). Note `OC_INSECURE` is unrelated here, it only maps to events TLS and the CS3 extractor in the search config.

Proposed fix, mirroring the go-micro client (the endpoint hunk is optional, today it is a hardcoded flag default):

```diff
--- a/services/search/pkg/command/index.go
+++ b/services/search/pkg/command/index.go
@@ -46,10 +46,20 @@ func Index(cfg *config.Config) *cobra.Command {
 				return fmt.Errorf("concurrency %d exceeds max allowed %d", concurrencyFlag, cfg.ReindexMaxConcurrency)
 			}
 
+			if !cmd.Flags().Changed("endpoint") {
+				endpointFlag = cfg.GRPC.Addr
+			}
+
 			var dialOpts []grpc.DialOption
-			if cfg.GRPCClientTLS.Mode == "insecure" || insecureFlag {
+			// mirror pkg/service/grpc.NewClient: ""/"off" -> no TLS, "insecure" -> TLS without verification, "on" -> TLS
+			switch {
+			case insecureFlag, cfg.GRPCClientTLS.Mode == "", cfg.GRPCClientTLS.Mode == "off":
 				dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
-			} else {
+			case cfg.GRPCClientTLS.Mode == "insecure":
+				dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{
+					InsecureSkipVerify: true, //nolint:gosec
+				})))
+			default:
 				dialOpts = append(dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{
 					MinVersion: tls.VersionTLS12,
 				})))
```

Verified against a running server built from main with an `opencloud init` default config: main fails without `--insecure`, the patched CLI connects without it. With `OC_GRPC_TLS_ENABLED=true` + `OC_GRPC_CLIENT_TLS_MODE=insecure`, main fails (`error reading server preface: EOF`) while the patched CLI accepts the self-signed cert, and `MODE=on` correctly rejects it. With this, the README/MIGRATION/log changes in this PR are not needed. `on` still ignores `OC_GRPC_CLIENT_TLS_CACERT`, that could be a follow-up.

fschade (approved):

PR #3510, opened 2026-09-10
fix(search): hierarchy tokenizer for bleve path fields

dschmidt:
## Description

Path in bleve was a keyword, so the descendant lookup (delete/move/restore/purge), the scoped search and the KQL `path:` predicate expanded into one term searcher per descendant and OOM-killed the server on large folders.

Path is now analyzed into its ancestor prefixes (`./a/b.txt` → `.`, `./a`, `./a/b.txt`), like `path_hierarchy` in OpenSearch. All three call sites are a single term query. Schema 4 → 5 (v4 never shipped).

The same tokenizer with `tag_depth` is registered as the `geohash` analyzer so #3272 can add its field additively. Minimal version of #3495: no descendant streaming, no geohash sibling.

Peak live heap, descendant lookup at 100k files: 1217 MB → 133 MB.

## Related Issue

- #1269, #3469

## How Has This Been Tested

- Tokenizer unit tests
- Parity: MOVE-03 (sibling folder sharing the prefix), MOVE-04 (special characters in the path)

## Types of changes

- [x] performance improvement

# This is an (imho) important performance fix. We've broken the index scheme for the next release already so now is the PERFECT time to get it in. 
 I've separated the breaking parts from my original PR into this. This already gets us a 10x less memory usage effect. So I would strongly suggest to merge it for the 8.0.0 release. The other improvements are non breaking and can happen at another time in the 8.x cycle. 

BTW: I don't have any other breaking changes planned (alas this wasn't planned but it was brought up and debugged just in time by another user/contributor). 
I put everything breaking into my refactor PR, which I separated out of a feature branch with all my stuff merged together and working. So I'm pretty confident that everything I've been working on can be added gradually during the 8.x series without a single breaking change.

### DIFF
--- a/services/search/pkg/mapping/opts.go
+++ b/services/search/pkg/mapping/opts.go
@@ -26,6 +26,11 @@ const WordsSuffix = "_words"
 // WordsAnalyzer names the analyzer both engines register for the words sibling.
 const WordsAnalyzer = "words"
 
+// PathAnalyzer names the analyzer both engines register for TypePath fields:
+// every ancestor prefix of a path is a term, so one term query matches a
+// folder and its descendants.
+const PathAnalyzer = "path_hierarchy"
+
 // FieldOpts overrides the default type inference for a struct field. Keys in
 // the override map are json-tag names (e.g. "Name", "location", "audio.artist"),
 // not Go field names.
--- a/services/search/pkg/bleve/backend.go
+++ b/services/search/pkg/bleve/backend.go
@@ -75,14 +75,10 @@ func (b *Backend) Search(_ context.Context, sir *searchService.SearchIndexReques
 			},
 		)
 		// Scope below the space root: restrict at query level so totals and
-		// paging respect the path too. Path is a case-preserving keyword
-		// (paths act as references, /Foo and /foo are distinct), so the exact
-		// folder or the folder prefix matches all of, and only, the scope.
+		// paging respect the path too. The folder term matches the folder and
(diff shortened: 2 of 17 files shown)


services/search/pkg/mapping/opts.go
```diff
@@ -26,6 +26,15 @@ const WordsSuffix = "_words"
...
 const WordsAnalyzer = "words"
 
+// PathAnalyzer names the bleve analyzer for TypePath fields: every ancestor
+// prefix is a term, like path_hierarchy in OpenSearch.
+const PathAnalyzer = "path"
+
+// GeohashAnalyzer names the bleve analyzer for a geohash: every prefix is a
+// depth-tagged term (1/u, 2/u4, ...), so a terms facet with TermPrefix
+// "<precision>/" is a geohash grid at that precision.
+const GeohashAnalyzer = "geohash"
```
fschade: Leftover?

dschmidt: Yes and no. 

This is the very minimal part that's needed to make use of this for the geohash analyzer later on without a breaking change or with new mechanics for handling new analyzers.

All the computation etc is removed.

dschmidt: That was a good catch, it actually didn't belong there at all. 

I would still like to keep the definition, but it's an implementation detail of the geopoint implementation in bleve, so absolutely no need to define this in the mapping package

fschade (approved):

Issue #3513, opened 2026-09-10
Share notifications are sent to disabled users


michaelstingl:
## Describe the bug

OpenCloud sends space-share notification emails to disabled users. This occurs both when a space is shared directly with a disabled user and when it is shared with a group containing a disabled member.

## Steps to reproduce

The following was reproduced locally with the existing acceptance-test stack and Inbucket as the SMTP sink.

1. Start `opencloudeu/opencloud-rolling:7.5.0` with the notifications service and SMTP delivery to Inbucket.
2. Create a disposable space owned by Alice.
3. For the direct-share case, disable Brian through the Graph API and share the space directly with Brian.
4. For the group-share case, create a group containing Brian and active Carol, disable Brian through the Graph API, and share the space with the group.
5. Check the Inbucket mailboxes.

## Expected behavior

Disabled users should not receive space-share notification emails. Active members of a shared group should continue to receive them.

## Actual behavior

Both share operations returned HTTP 200, but Brian's Inbucket mailbox contained one email although no email was expected. In the group-share case, active Carol also received her notification. The behavior was reproduced with both direct and group shares on `7.5.0`.

## Setup

- Server image: `opencloudeu/opencloud-rolling:7.5.0`
- Acceptance runner: local `woodpecker-cli exec` on Rancher Desktop Moby
- Mail sink: `inbucket/inbucket:latest`, SMTP on `email:2500`, API on `email:9000`
- Storage driver: posix

## Source trace

The observed direct-share path is:

```text
Graph API direct share
  -> SpaceShared event
    -> notifications: handleSpaceShared()
      -> ensureGranteeList() / getGranteeList()
        -> direct grantee: disableEmails() check
          -> GetUser() returns the CS3 user object
            -> missing-email check
              -> render and send the email to Brian's address
```

The observed group-share path is:

```text
Graph API group share
  -> SpaceShared event
    -> notifications: handleSpaceShared()
      -> ensureGranteeList() / getGranteeList()
        -> GetGroup() and iterate all group members
          -> filter executant and notification opt-out
            -> GetUser() returns the CS3 user object
              -> missing-email check
                -> render and send the email to Brian's address
```

In the `v7.5.0` source, `getGranteeList()` has no account-state filter for expanded group members (`services/notifications/pkg/service/service.go:200-249`). The CS3 user object used by `GetUser()` contains identity and mail fields but no `AccountEnabled` field. The Graph/LDAP model maps account state separately.

[...]

PR #3518, opened 2026-09-11
fix(notifications): skip disabled email recipients

michaelstingl:
## Description

Skip notification emails to users disabled through OpenCloud's user management. This covers share notifications to direct recipients and group members, mentions, and daily or weekly grouped emails, including events stored before disablement.

## Related Issue

Related to https://github.com/opencloud-eu/opencloud/issues/3513

## Motivation and Context

The check uses `GetUserByClaim` before sending because `GetUser` does not exclude disabled LDAP users.

An existing LDAP lookup cache entry can delay suppression until it expires (10 seconds by default). Disabling a user only in an external identity provider is outside this change.

ScienceMesh invitations use an email address without requiring an OpenCloud user, so they skip the recipient lookup.

Lookup errors skip the recipient. Grouped-email events are still removed before delivery and are not retried after a later lookup or SMTP failure.

### Unit tests and static checks

- Notification package tests and service race tests pass, including disablement after events are stored and recipient lookup failures.
- Static checks pass for the changed notification code.

### Acceptance tests

Local runs with Inbucket compare the same source base without and with the fix:

- Seven regression scenarios fail without the fix and pass with it.
- Active recipients still receive notifications.
- The new acceptance scenarios pass with POSIX and decomposed storage.
- The full `emailNotification.feature` passes with decomposed storage.
- The removed expected-failure entries cover space-membership expiry and share-removal emails; both tests already pass without this fix.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation added

<sub>🤖 drafted with OpenAI Codex, reviewed before submitting.</sub>

### DIFF
--- a/services/notifications/pkg/service/service.go
+++ b/services/notifications/pkg/service/service.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/opencloud-eu/reva/v2/pkg/events"
 	"github.com/opencloud-eu/reva/v2/pkg/rgrpc/todo/pool"
+	"github.com/rs/zerolog"
 
 	"github.com/opencloud-eu/opencloud/pkg/l10n"
 	"github.com/opencloud-eu/opencloud/pkg/log"
@@ -160,10 +161,17 @@ func (s eventsNotifier) Close() {
 	}
 }
 
+// recipientMessage keeps the recipient's user ID with the rendered message.
+// send resolves the email address before passing the message to the channel.
+type recipientMessage struct {
+	message   *channels.Message
+	recipient *user.UserId
+}
+
 func (s eventsNotifier) render(ctx context.Context, template email.MessageTemplate,
-	granteeFieldName string, fields map[string]string, granteeList []*user.User, sender string) ([]*channels.Message, error) {
+	granteeFieldName string, fields map[string]string, granteeList []*user.User, sender string) ([]recipientMessage, error) {
 	// Render the Email Template for each user
-	messageList := make([]*channels.Message, len(granteeList))
+	messageList := make([]recipientMessage, len(granteeList))
 	for i, usr := range granteeList {
 		locale := l10n.MustGetUserLocale(ctx, usr.GetId().GetOpaqueId(), "", s.valueService)
 		fields[granteeFieldName] = usr.GetDisplayName()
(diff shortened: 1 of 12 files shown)


services/notifications/pkg/service/service.go
```diff
@@ -173,18 +181,70 @@ func (s eventsNotifier) render(ctx context.Context, template email.MessageTempla
...
+	if r.GetStatus().GetCode() != rpc.Code_CODE_OK {
+		return nil, fmt.Errorf("recipient lookup failed: %d: %s", r.GetStatus().GetCode(), r.GetStatus().GetMessage())
+	}
+	u := r.GetUser()
+	if u.GetId().GetOpaqueId() != id.GetOpaqueId() ||
+		(id.GetIdp() != "" && u.GetId().GetIdp() != id.GetIdp()) ||
+		(id.GetTenantId() != "" && u.GetId().GetTenantId() != id.GetTenantId()) {
+		return nil, errors.New("recipient lookup returned a missing or mismatched identity")
+	}
+	if u.GetStatus() == user.UserStatus_USER_STATUS_BLOCKED || strings.TrimSpace(u.GetMail()) == "" || s.disableEmails(ctx, id) {
```
rhafer, line 237: Please note that we're currently not populating the `Status` field in reva anywhere (. So unless that changes this whole PR is pretty much useless.

Currently in OpenCloud there is no way to figure out on the CS3 API if a user is enabled or disabled.

When adding that enabled/disable feature on the Graph API  the only purpose of it was to be able to prevent users from being able to login to OpenCloud. We deliberately accepted that fact that it is still possible to create shares with users that are disabled.

michaelstingl: The tested LDAP behavior does not depend on `User.Status`. The PR calls `GetUserByClaim("userid", ...)` before delivery and skips the email on `NOT_FOUND`. In the tested LDAP configurations, the disablement filters exclude disabled users from this lookup. This is not a general CS3 enabled/disabled status query. Cached lookup results can delay suppression until expiry.

Local acceptance tests show delivery to disabled users without the fix and suppression with it.

The proposal leaves sharing with disabled users possible, but suppresses their notification emails. Whether disablement should have that effect is the product decision.

<sub>🤖 drafted with OpenAI Codex, reviewed before submitting.</sub>

michaelstingl: Removed the `User.Status == BLOCKED` check and its dedicated test case in [d2a0d7dddb](https://github.com/opencloud-eu/opencloud/commit/d2a0d7dddba531b4f4849caded9d53de60a83266). The notification tests and local LDAP lookup probes still pass.

<sub>🤖 drafted with OpenAI Codex, reviewed before submitting.</sub>

PR #3521, opened 2026-09-11
feat(proxy): add per-service metrics to the proxy service

pbleser-oc:
## Description
Improve the metrics for the proxy service.

* extend the `instrumenter` middleware to add a label with the name of the service the request is being dispatched to
* add a native Prometheus histogram that tracks the durations and also includes a label for the name of the service the request is being dispatched to, as well as a `result` label (`success`, `client-error`, `server-error`) based on the HTTP status code of the response
* add per-service gauge functions to count the number of in-flight requests
* add a counter for routing failures, for when an inbound request cannot be mapped to a route
* extend the route.`RoutingInfo` struct with a `service` attribute, and a `Service()` getter
* remove the `routing_failure_count` metric, as it is redundant

Added metrics:
* `opencloud_proxy_request_duration_seconds{service=..., method=..., result=...}` (native histogram)
* `opencloud_proxy_service_request_total{service=..., method=..., result=...}` (counter)
* `opencloud_proxy_service_duration_seconds{service=..., method=..., result=...}`  (legacy histogram)

The labels:
* `method`: is the HTTP method (`GET`, ...)
* `result`: maps the HTTP response status code to one of three values to reduce the cardinality, either `success` (0-399), `client-error` (400-499), `server-error` (500-)
* `success`: is the service identifier, e.g. `eu.opencloud.web.webdav`, `eu.opencloud.web.graph`, ...

> [!NOTE]
> Previously existing metrics have been kept as-is, backwards compatibility is retained.

## Motivation and Context
More detailed metrics will allow for better introspection of a running production installment, in order to determine if something is slow, and what, by being able to drill down by service and not just with global metrics that do not differentiate.

## How Has This Been Tested?
Manually by checking the resulting metrics when querying the `/metrics` endpoint of the debug address, as well as with a Prometheus scraping container.

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [X] Documentation added


services/proxy/README.md
```diff
@@ -300,12 +300,14 @@ In this mode, the proxy service only exposes its own metrics. The metrics of the
...
-|----------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------|
-| `opencloud_proxy_requests_total`      | [Counter](https://prometheus.io/docs/tutorials/understanding_metric_types/#counter) metric which reports the total number of HTTP requests.                                                                                   | `method`: HTTP method of the request  |
-| `opencloud_proxy_errors_total`        | [Counter](https://prometheus.io/docs/tutorials/understanding_metric_types/#counter) metric which reports the total number of HTTP requests which have failed. That counts all response codes >= 500                           | `method`: HTTP method of the request  |
```
micbar, line 306: I have a lot of dashboards which need the error rate built by 
`opencloud_proxy_requests_total / opencloud_proxy_errors_total`

Why does `opencloud_proxy_errors_total` get removed? `opencloud_proxy_routing_failure_count` is, according to its naming a different scope.

micbar (commented):
See my question.

services/proxy/README.md
pbleser-oc: It's redundant because it contains the same as `sum(opencloud_proxy_request_total{result=server-error})`, but I can also reintroduce them if it helps with metric "API" compatibility.

pbleser-oc: I will re-add the "old" ones to keep the metrics compatible.

pbleser-oc: Resolved with an additional commit, f588ce94a01ed3fe778070ac25a64dafb65b85c3, which re-adds the metrics that existed previously, with the same labels and name, for backwards compatibility.

For more useful dashboards, one can switch to the new metrics instead, preferably to `opencloud_proxy_duration_seconds`: a native histogram, that includes:
* `opencloud_proxy_duration_seconds_count`,
* `opencloud_proxy_duration_seconds_sum`,
* and duration buckets with `opencloud_proxy_duration_seconds_bucket`.

That one contains labels for
* `method`,
* `result`: to lower cardinality, that label maps HTTP status codes to `success` (-399), `client-error` (400-499), `server-error` (500-599),
* and `service`.

micbar (approved):
Nice!

PR #3526, opened 2026-09-14
fix(graph): fix PatchMe method to prevent password change

maki5:
Prevent updating user password using `PATCH /graph/v1.0/me` method


services/graph/pkg/service/v0/users.go
```diff
@@ -852,6 +852,13 @@ func (g Graph) PatchMe(w http.ResponseWriter, r *http.Request) {
 		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "user is not allowed to change own mail")
 		return
 	}
+
+	if changes.HasPasswordProfile() {
+		logger.Info().Interface("user", changes).Msg("could not update user: user is not allowed to change own password profile")
+		errorcode.InvalidRequest.Render(w, r, http.StatusBadRequest, "user is not allowed to change own password profile")
```
rhafer: I think this should rather be an `errorcode.AccessDenied` with `http.StatusForbidden`, right?

rhafer (changes requested):
While this fixes the `/me` issue

services/graph/pkg/service/v0/users.go
maki5: Forbidden make sense. Should we then change the statuses and error codes for other checks as well? because they follow the same logic of non permitting some operation

rhafer: Makes sense.

rhafer:
@maki5 I think this needs a rebase for the changed minio image.

rhafer (approved):

PR #3542, opened 2026-09-15
fix(search): refresh service credentials during long index walks

### LINKED ISSUE #3541: Search: long-running IndexSpace traversal reuses an expired service token
# Search: long-running IndexSpace traversal reuses an expired service token

## Describe the bug

A full-space reindex can outlive the service token acquired at the beginning of `IndexSpace`. The same authentication context is then reused for later gateway operations, causing the traversal to fail instead of refreshing its credentials.

In our original deployment, a force-rescan failed after **86,424 seconds** (24 hours and 24 seconds) with `PermissionDenied` reporting an invalid core access token. OpenCloud and Tika remained healthy, with no container restarts or OOM events. Index writes and extraction activity stopped at that failure.

## Setup and verification scope

- Originally observed on **OpenCloud 7.4.0**, using Bleve and Tika for a large document archive. That build included a separate Bleve/zapx fix, but not this authentication refresh fix.
- The original deployment ran in rootless Podman containers managed by systemd/Quadlet. The current deployment uses rootful Podman/Quadlet.
- The original incident and diagnosis were recorded on **2026-08-19**.
- The same one-time authentication pattern is still present in the **v8.0.0** source, checked on **2026-09-16**.
- Our current deployment uses OpenCloud 8.0.0, OpenSearch and Tika OCR with a local workaround. We have **not** repeated the full 24-hour failure on an unpatched v8.0.0 deployment.
- This concerns the internal service/core token used by indexing, not the user's browser session or OIDC refresh token.

## Steps to reproduce

1. Use an unpatched search service and a space whose full traversal/content extraction takes longer than the internal service token lifetime.
2. Start a forced reindex of that space with concurrency 1.
3. Keep the indexing request alive beyond the token lifetime. A separate client-side timeout must not terminate the request first (see below).
4. Allow a subsequent authenticated gateway operation to occur after the original token expires.

In the observed configuration, the token lifetime was 86,400 seconds. The token manager's vendored default is also [86,400 seconds](https://github.com/opencloud-eu/opencloud/blob/v8.0.0/vendor/github.com/opencloud-eu/reva/v2/pkg/token/manager/jwt/jwt.go#L36).

For a faster regression test, use an injected clock and an authentication stub with distinguishable tokens. Advance the clock between the walker's initial `Stat` and a later `ListContainer`, and assert that the later request uses refreshed credentials.

## Expected behavior

Indexing should refresh its internal authentication before expiry and continue traversing the space without requiring a restart or a global increase of token lifetime.

## Actual behavior

The long-running traversal reused its initial service context and failed with `PermissionDenied` once the core token expired. This left the full reindex incomplete.

The elapsed time and error above come from our recorded incident report; they are not presented as a verbatim raw log excerpt.

## Source analysis

In [v8.0.0 `services/search/pkg/search/service.go`](https://github.com/opencloud-eu/opencloud/blob/v8.0.0/services/search/pkg/search/service.go#L463), `IndexSpace`:

1. Calls `getAuthContext(...)` once before walking the tree (line 464).
2. Passes that context into `w.Walk(...)` (line 505).
3. Reuses it for `s.engine.Search(...)` in the callback (line 532).

There is no refresh in this traversal path. This explains why a healthy service can fail partway through a sufficiently long reindex.

## Local workaround and regression test

We have a small local patch that caches an indexing-specific authentication context and refreshes it every **12 hours**. It supplies the current context to the walker's `Stat` and `ListContainer` operations and the per-resource index lookup. It leaves the global token lifetime unchanged.

The accompanying test, `TestIndexSpaceWalkerRefreshesServiceTokenDuringLongWalk`, uses the real walker, mocked gateway calls and a controlled clock. It verifies that traversal crosses the refresh boundary and subsequent gateway calls receive the second token. The test passes on our patched v8.0.0 build.

The 12-hour interval is a workaround for our 24-hour token lifetime, not a proposed universal default. An upstream fix should account for the actual expiry/configuration, preserve cancellation and deadlines when refreshing authentication metadata, and propagate authentication-refresh failures.

I can provide the patch and regression test for review.

## Separate CLI timeout issue

The v8.0.0 CLI also has a [hard-coded 10-minute context timeout](https://github.com/opencloud-eu/opencloud/blob/v8.0.0/services/search/pkg/command/index.go#L71). That can prevent a long-running CLI reproduction from reaching token expiry. It is a separate issue: `DeadlineExceeded` after 10 minutes is not evidence of an expired token.

For our current background job, we added a local configurable `--timeout` option and disabled that deadline. This report is specifically about refreshing authentication during `IndexSpace`.

pntone:
## Description

Refresh the internal service credentials during `IndexSpace` instead of reusing the context created at the beginning of the entire traversal. A space taking longer than the token lifetime can otherwise fail with `PermissionDenied` partway through reindexing.

The walker obtains current credentials for `Stat` and `ListContainer`; the per-resource index lookup uses the same cache. JWT credentials are renewed at half their remaining lifetime, capped at 12 hours. Tokens without a readable JWT expiry use the 12-hour fallback. Reading the expiry only controls refresh scheduling; token verification and authorization remain with the gateway.

Only authentication metadata is replaced. The calling context's deadline, cancellation, values and unrelated outgoing metadata are preserved. Authentication failures propagate instead of falling back to stale credentials. The cache is local to one space traversal and serializes concurrent refreshes.

This does not change the global token lifetime, index mappings, Web search behavior or the separate indexing CLI timeout.

## Related Issue

Fixes #3541

## Motivation and Context

A recorded full-space reindex failed after 86,424 seconds against a 24-hour core token lifetime while OpenCloud and Tika remained healthy. The one-time authentication pattern is also present in the current source. Renewing credentials within the traversal allows long jobs to continue without increasing the lifetime of all service tokens.

## How Has This Been Tested?

- Test environment: Linux amd64, Go 1.26.8; based on the current `main` branch.
- `go test ./services/search/pkg/search -race -count=1 -timeout=120s` — passed.
- Real walker with mocked gateway and controlled clock: later `ListContainer` uses renewed credentials after the refresh boundary.
- Short-lived JWT: refreshed halfway through its one-hour lifetime, before the 12-hour fallback interval.
- Caller deadline, cancellation, values and outgoing metadata survive refresh; parent metadata is unchanged and stale token values are replaced.
- Refresh errors propagate; missing and expired authentication results are rejected.
- Concurrent callers share one authentication; race detector passes.
- `gofmt` and `git diff --check` — passed.

A new 24-hour end-to-end run of this exact upstream candidate has not been performed. The original fixed-interval workaround is deployed locally; this PR additionally handles shorter JWT lifetimes and preserves caller context.

## Types of changes

- [ ] Technical debt

## Checklist

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added

The behavior and fallback are documented in source comments; there are no new configuration options.


aduffeck:
Thank you for your contribution, we ended up fixing the issue in a slightly different fashion in #3556

PR #3556, opened 2026-09-17
Use an auth.Session for indexing spaces

### LINKED ISSUE #3541: Search: long-running IndexSpace traversal reuses an expired service token

aduffeck:
This makes sure that the token does not expire if the indexing process takes longer than the token lifetime.

This commit also makes sure that the authenticated context is reused instead of generating a new context for each resource which is much more efficient.

**Note:** Depends on https://github.com/opencloud-eu/reva/pull/804

Fixes #3541
Supersedes #3542


services/search/pkg/search/service.go
```diff
@@ -16,7 +16,9 @@ import (
 	rpcv1beta1 "github.com/cs3org/go-cs3apis/cs3/rpc/v1beta1"
 	collaborationv1beta1 "github.com/cs3org/go-cs3apis/cs3/sharing/collaboration/v1beta1"
 	provider "github.com/cs3org/go-cs3apis/cs3/storage/provider/v1beta1"
+	"github.com/golang-jwt/jwt/v4"
```
rhafer: Can we use jwt/v5 here? I know we're also still bundling v4 but just because of transient dependencies.

services/search/pkg/search/service.go
```diff
@@ -794,21 +817,57 @@ func (s *Service) MoveItem(ref *provider.Reference) {
...
+}
 
-	return ownerCtx, statRes, r.GetPath()
+func tokenExpiry(ctx context.Context) (time.Time, error) {
+	tkn := tokenFromContext(ctx)
+	if tkn == "" {
+		return time.Time{}, errors.New("no auth token in context")
+	}
+	var claims jwt.RegisteredClaims
+	if _, _, err := jwt.NewParser().ParseUnverified(tkn, &claims); err != nil {
```
rhafer, line 866: The unit tests use `authtoken` as the token value. Which will cause an error here. I guess the test need to generate a valid JWT now?

rhafer (changes requested):

rhafer (approved):

Issue #3559, opened 2026-09-18
search re-indexing fails on disabled spaces


schweigisito:
## Describe the bug
opencloud 8.x.x introduces a new search index version in both bleve and also opensearch. This enforces a re-indexing of all spaces. Unfortunately the search re-indexing fails on disabled spaces.

## Steps to reproduce
- Create a new space, e.g. `testspace` and disable it
- Trigger the re-indexing via 

```
$ docker compose exec opencloud opencloud search index --all-spaces --force-rescan --insecure
```

=> It fails with:

```
failed to index space: rpc error: code = Unknown desc = error: not found: stat: error: not found: d9dc2372-2684-48f8-aa99-3fc672eddcf3
Error: rpc error: code = Unknown desc = error: not found: stat: error: not found: d9dc2372-2684-48f8-aa99-3fc672eddcf3
Usage:
  opencloud search index [flags]

Aliases:
  index, i

Flags:
      --all-spaces        index all spaces instead. This or --space is required.
      --endpoint string   the address of the search service gRPC endpoint. (default "127.0.0.1:9220")
      --force-rescan      force a rescan of all files, even if they are already indexed. This will make the indexing process much slower, but ensures that the index is up-to-date using the current search service configuration.
  -h, --help              help for index
      --insecure          disable TLS for the gRPC connection.
  -s, --space string      the id of the space to travers and index the files of. This or --all-spaces is required.

rpc error: code = Unknown desc = error: not found: stat: error: not found: d9dc2372-2684-48f8-aa99-3fc672eddcf3
```

- Enable space
- Re-run the re-indexing

=> Space gets re-indexed successfully.

## Expected behavior
There are different scenarios, what the expected behaviour might be:

- either skipping disabled spaces, but at least log them to tell the admin it needs to be re-indexed once it has been enabled again
- or also re-indexing it although it is disabled, with the assumption that the search will work, once the space has been enabled again
- or implement an automation which triggers the re-indexing once a disabled space gets enabled again

## Actual behavior
See abouve.

## Setup
Happens with `7.2.4` in k8s deployment using `opensearch` backend and also with docker compose using`bleve`.

8.0.1 seems to skip disabled spaces but not really index them.

v-scharf:
sorry, I couldn't reproduce: 
[...]

PR #3565, opened 2026-09-18
[stable-7.2] fix(search): skip disabled spaces when reindexing all spaces

schweigisito:
## Description

`opencloud search index --all-spaces` aborted on the first disabled space it ran into.

`ListStorageSpaces` returns disabled spaces, but the storage provider answers with `not found` for their root (`node.ReadNode` is called with `canListDisabledSpace=false`). reva's walker stats the space root before it descends, so the walk fails immediately and `IndexSpace` returns that error. The handler propagated it out of the loop, which ended the whole run and left every space that had not been processed yet unindexed:

```
failed to index space: rpc error: code = Unknown desc = error: not found: stat: error: not found: d9dc2372-2684-48f8-aa99-3fc672eddcf3
```

Disabled spaces are now skipped and logged, using the same `trashed` opaque marker the search service already uses to keep disabled spaces out of query results (`services/search/pkg/search/service.go`). A summary line at the end tells the admin how many spaces were skipped so they can be indexed again after being enabled.

Indexing a disabled space is not possible through the CS3 gateway at all — both `Stat` and `ListContainer` resolve the space root through `ReadNode` and get `not found` — so skipping is the only option available here without a reva change.

This is not a backport. `main` does not abort the run, because `c8805265e2` swallows `not found` inside the walk callback, but as a result it reports disabled spaces as successfully indexed without touching them. That needs a separate fix on `main`, see the issue.

Scope is deliberately kept to the disabled space failure. Two adjacent things are **not** in this PR:
- a space failing for any other reason still aborts the run (on `main` this was changed by #3207 / `243514ec17`)
- the CLI's 10 minute client timeout, which is a second way the charts post upgrade job can fail on large instances — that is #3544

## Related Issue

- Fixes https://github.com/opencloud-eu/opencloud/issues/3559

## Motivation and Context

Indexing all spaces is what the post upgrade job of the helm charts runs after the search index version bump in 8.x. A single disabled space made that job fail and left the instance with a partially rebuilt index.

## How Has This Been Tested?

- test environment: local build of this branch, `go build ./services/search/...`, `go vet`, `go test ./services/search/...` (the `pkg/opensearch` suite needs a Docker testcontainer and is unaffected by this change)
- test case 1: create a space, disable it, run `opencloud search index --all-spaces --force-rescan --insecure` — the command now completes, the disabled space is logged as skipped and the remaining spaces are indexed
- test case 2: enable the space again and re-run the command — the space is indexed as before

An acceptance test for this scenario exists for `main` in #3560 and should be backported once it has been merged.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


schweigisito:
Closing it for now.

PR #3568, opened 2026-09-20
Feature: file-based secrets from environment

### LINKED ISSUE #3567: Support file-based secrets from environment
## More secure file-based provision of sensitive values

Sensitive values such as passwords, api keys, tokens, JWT secrets, etc. can easily be leaked when provided directly as environment variables.

For reference and more details why environment variables are a poor choice for providing secrets read for example https://blog.diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/

## Add support for `_FILE`-suffixed environment variables for sensitive values

To still allow configuration via the environment but securely load secrets from files add support for additional environment variables that ("magically") load the secrets from file. The common pattern seems to be the add a `_FILE` suffixed environment variable for each such environment variable.

An example how this looks elsewhere: https://www.authelia.com/configuration/methods/secrets/

This allows to provide file paths via the environment and use container secrets, systemd credentials or similar mechanisms to provide sensitive values more securely.

parasite-lost:
## Description

Add support for all environment variables providing sensitive values to provide a path to a file via `_FILE` suffixed environment variable instead where the file contains the sensitive value.

This can be used for example in conjunction with container secrets or systemd credentials to securely provide sensitive values.

## Related Issue

- Fixes https://github.com/opencloud-eu/opencloud/issues/3567

This is also related to https://github.com/opencloud-eu/opencloud/pull/3034 (PR https://github.com/opencloud-eu/opencloud/pull/3034).

## Motivation and Context
Securely provide sensitive values by files (referenced via environment) instead of leaking sensitive values in the environment.

## How Has This Been Tested?

- added several unittests for envdecode

## Types of changes

While maintaining full backwards compatibility (all prior options for configuring parameters haven't changed):
- [x] Minor refactoring
- [x] Support `_FILE` suffixed environment variables via envdecode struct tags (by adding `,file` parameter to the environment variable definition)
- [x] Add support for all secrets, api keys, tokens, passwords that can be configured via the environment that I found

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation for envdecode
- [ ] Documentation added


pkg/config/envdecode/envdecode.go
```diff
@@ -214,6 +240,20 @@ func decode(target any, strict bool) (int, error) {
 	return setFieldCount, nil
 }
 
+func readFileEnv(envVariable string) (string, bool, error) {
+	if file, set := os.LookupEnv(envVariable); set {
+		if file == "" {
+			return "", true, fmt.Errorf("no file provided: %s=", envVariable)
+		}
+		content, err := os.ReadFile(file)
```
parasite-lost, line 248: Codacy Static Code Analysis complains that this line is not up to standards as it would allow reading user-defined files.

But this is exactly the intent - let the user (admin in this case) provide paths to arbitrary files that should be read to provide sensitive values.

*question*: how can I suppress this finding?

dschmidt:
To my eyes the comma syntax is a bit weird, it looks like a list (especially when there's only one env var)

Two thoughts on this: 
1. You can already avoid having secrets as env vars by using config files
2. If adding an additional mechanism for handling _FILE in a generic way, why limit it to specific env vars? Why not handle it for all env vars? Then you don't need to invent a new declaration syntax for that

parasite-lost:
@dschmidt 

> To my eyes the comma syntax is a bit weird, it looks like a list (especially when there's only one env var)

That's the syntax of envdecode struct tags. You can already define default values by appending the definition with `,default=my-default-value` or appending `,required` to denote that the variable must be provided. I just extended this with `,file`.

> Two thoughts on this:
> 
>     1. You can already avoid having secrets as env vars by using config files

Using config files feels a bit unwieldy and verbose as (if I understand things correctly) you have to provide the full configuration including default settings. Plus the documentation states that the aim is that opencloud can easily be configured by setting just the necessary parameters via environment variables.


>     2. If adding an additional mechanism for handling _FILE in a generic way, why limit it to specific env vars? Why not handle it for all env vars? Then you don't need to invent a new declaration syntax for that

I'm fine with adding the `_FILE` suffix for all environment variables - if that is preferable (it would make things much easier, code and documentation).

PR #3569, opened 2026-09-20
fix(collaboration): match WOPI app extensions without case

### LINKED ISSUE #3449: OpenInApp file-extension lookup is case-sensitive; UI error hides the cause
## Describe the bug

Opening a Word file in Euro-Office / OnlyOffice (WOPI) fails for `*.DOCX` and succeeds for the same content saved as `*.docx`.

The web UI shows:

> Error contacting the requested application, please use a different one or try again later

That reads like the office app is down. Collaboration logs:

```
OpenInApp: neither edit nor view app URL found
```

WOPI discovery was fine. `Registering mimetypes` included `application/vnd.openxmlformats-officedocument.wordprocessingml.document`.

## Steps to reproduce

1. Compose with `weboffice/euro-office.yml` (or OnlyOffice). Confirm `/hosting/discovery` is 200 and collaboration registers mimetypes.
2. Upload `test.DOCX` and open it in the web UI.
3. Upload `test.docx` and open it.

## Expected behavior

Office extensions are matched case-insensitively (Windows/macOS uploads are often `.DOCX`). If there is still no app URL, the UI and log should include the extension that was looked up.

## Actual behavior

- `test.DOCX` → UI error above; log `neither edit nor view app URL found` (no `fileExt` in the log).
- `test.docx` → editor opens.

`OpenInApp` uses `path.Ext(resourceInfo.GetPath())` as a map key. Discovery stores `"." + ext` from XML (`ext="docx"`). Go `path.Ext` does not fold case, so `.DOCX` ≠ `.docx`.

https://github.com/opencloud-eu/opencloud/blob/v7.5.0/services/collaboration/pkg/service/grpc/v0/service.go#L111-L118

## Suggested fix

- Look up `strings.ToLower(path.Ext(...))` (and lowercase keys when parsing discovery).
- Log `fileExt` (and maybe available keys) on this error.
- Map the frontend `SERVER_ERROR` to something like “No editor for extension .DOCX”.

## Setup

- `opencloudeu/opencloud-rolling:7.5.0`
- `COLLABORATION_APP_NAME=Euro-Office`, `COLLABORATION_APP_PRODUCT=OnlyOffice`

NickWalters:
## Description

The WOPI discovery response lists its extensions in lower case, and `parseWopiDiscovery` stores them as the app published them. `OpenInApp` then looks up `path.Ext()` of the real file name against those keys, which is an exact match. So `report.DOCX` finds no app url while `report.docx` opens, and the web UI says it cannot contact the application, which reads like the editor is down.

Both ends now fold the case: the parser lowercases the `ext` before it becomes a map key, and `GetAppURLFor` lowercases the extension it is given. Either one alone would leave the other side broken if an app ever published an upper case `ext`.

The log line and the returned error also name the extension. Neither said which file type had no app, which is the second half of the issue.

Behaviour worth knowing: if a discovery response declared the same extension twice in different cases, the last one in document order now wins instead of both being kept. No app does that today, and both entries would point at the same action.

## Related Issue

- Fixes #3449

## Motivation and Context

Files arrive with whatever case the client gave them. A `.DOCX` from a Windows share or an old export is a normal thing to find in a space, and there is no reason for it to open differently from a `.docx`. The generic error made it look like an outage of the office app, so the cause was hard to find.

## How Has This Been Tested?

- test environment: `go test ./services/collaboration/...`, all packages pass
- test case 1: `GetAppURLFor` with `.DOCX` and `.DocX` against stored lower case keys, for both the view and the edit action
- test case 2: a discovery response with `ext="ODT"`, which has to come back as `.odt`
- both specs fail without the change (2 failed, 16 passed) and pass with it
- `go vet ./services/collaboration/...` is clean

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation added


aduffeck (approved):
lgtm, thank you :)

NickWalters:
@aduffeck could you add `Type:Bug` to my PRs? The release notes are built from the `Type:*` labels, so unlabeled PRs get left out. This one is already missing from the draft 8.0.2 notes in #3534.

- opencloud-eu/opencloud#3569
- opencloud-eu/reva#796
- opencloud-eu/reva#797
- opencloud-eu/reva#807
- opencloud-eu/reva#808
- opencloud-eu/reva#809

Thanks!

PR #3572, opened 2026-09-21
fix(collaboration): restricted endpoints for collaboration service

### LINKED ISSUE #3131: collaboration fonts management review
## Doesn't work out of the box, missing documentation

- The office-settings app is disabled by default (makes sense, since it's only useful in combination with Collabora)
- opencloud-compose, does not enable the app when collabora is enabled, neither does it configure the `remote_font_config` url required to collabora to download the fonts
- There is no documentation for how to set things up.

## New unprotected endpoints at `/collaboration/fonts` (enabled by default)

The change introduces a couple of new unauthenticated endpoints that are supposed to enable collabora to list and download additional fonts. We're exposing them on the proxy service by default. Which allows anybody to list, download and generate previews of those additional fonts (potential DoS vector)
Technically it shouldn't be need to expose them on the proxy. In our compose deployment collabora and opencloud are running in the same docker network, so collabora would be able to talk the the `collaboration` service directly for the font listing and download (bypassing the proxy). 

The `preview` endpoint is needed by the `office-settings` app and should not be unprotected at all, but should require authentication.

The default proxy configuration adds an unprotected endpoint that forwards all requests on `/collaboration` to the `collaboration` service, unauthenticated. We should NOT do that.

Other than that, most of what was already written on #3128 is true.

maki5:
closes https://github.com/opencloud-eu/opencloud/issues/3131

dependent on https://github.com/opencloud-eu/opencloud-compose/pull/388


services/settings/pkg/store/defaults/permissions.go
```diff
@@ -89,7 +89,7 @@ func ChangeLogoPermission(c settingsmsg.Permission_Constraint) *settingsmsg.Sett
 // CollaborationManageFontsPermission is the permission to manage fonts
 func CollaborationManageFontsPermission(c settingsmsg.Permission_Constraint) *settingsmsg.Setting {
 	return &settingsmsg.Setting{
-		Id:          "ed83fc10-1f54-4a9e-b5a7-fb517f5f3e01",
+		Id:          "0cbb6276-fbcd-4594-a2b7-ea96812f2968",
```
rhafer: Why is the permissions's id changing?

rhafer, line 92: Why is the permissions's id changing?

maki5: @rhafer this id was already used for ChangeLogoPermission, so it basically was duplicated, should I revert it?

rhafer: Ah, ok. Would have been cool to have that in a separate commit with a proper message to make it more obvious..

rhafer (commented):
The changes look good to me.

It will however require changes to the `office-settings` app.  The `/collaboration/fonts/preview` endpoint requires authentication now. Currently the app is accessing them unauthenticated. 

@fschade @JammingBen Could you looking into that?

JammingBen:
> The changes look good to me.
> 
> It will however require changes to the `office-settings` app. The `/collaboration/fonts/preview` endpoint requires authentication now. Currently the app is accessing them unauthenticated.
> 
> @fschade @JammingBen Could you looking into that?

https://github.com/opencloud-eu/web/pull/3422 takes care of that.

rhafer (approved):

